PR #103, opened 2021-11-29
Added sbt flags to fix Docker build

vietjtnguyen:
On CADRE our cross compilation process occurs inside Yocto based Docker containers. In order to get `fpp` building in that I needed to resolve https://github.com/sbt/sbt/issues/6101. Specifically I used the solution in https://github.com/sbt/sbt/issues/6101#issuecomment-924554619. Admittedly this is a copy-paste-hey-it-worked solution so I'm not sure if it's something long-term to mainline.

### DIFF
--- a/compiler/install
+++ b/compiler/install
@@ -42,7 +42,10 @@ sed -i.bak -e "s/val commit = .*/val commit = \"$commit\"/" \
   $util/Version.scala
 
 echo "Building jar files"
-sbt assembly
+# The --batch and -Dsbt.server.forcestart=true flags were added to get the
+# build to work without user interaction in a Docker container. See
+# https://github.com/sbt/sbt/issues/6101#issuecomment-924554619.
+sbt --batch -Dsbt.server.forcestart=true assembly
 
 echo "Restoring Version.scala"
 cp $util/Version.scala.bak $util/Version.scala


bocchino:
That change doesn't work for me. My sbt doesn't like the `--batch` option. Also, I can't see any documentation suggesting that sbt should accept this option.

Have you tried running `sbt -Dsbt.io.virtual=false -Dsbt.server.autostart=false` as suggested here?

https://github.com/sbt/sbt/issues/6101#issuecomment-980819088

That works for me, but I'm not sure if it resolves your issue.

bocchino:
I am on sbt 1.4.6 (via MacPorts). Perhaps `--batch` was added in a later version.

vietjtnguyen:
I used SBT 1.5.5 which was the latest release version on the website (https://www.scala-sbt.org/download.html). I didn't have it available in the continuous integration environment (Ubuntu 20.04 based Docker image, doesn't seem to have Scala packages at all) or our cross compilation environment (a Yocto based Docker image). As such I just grabbed the latest. Is there a particular version that should be used for FPP?

vietjtnguyen:
In retrospect I should have just made this an issue with the patch included rather than a merge request. Feel free to close.

bocchino:
There's no particular sbt version required, but I'd prefer not to break compatibility with MacPorts/1.4.6.

I'm OK to leave the PR open. If you try `sbt -Dsbt.io.virtual=false -Dsbt.server.autostart=false` and it works, then I think we should make that change.

bocchino:
I updated my MacPorts, and now I'm at sbt 1.5.5. The `--batch` option works, but I notice that it suppresses the colors in the console display.

vietjtnguyen:
Unfortunately just changing the flag didn't work for me.

```
+ docker run --rm --net=host --privileged -w /home/vnguyen --volume=/dev/bus/usb:/dev/bus/usb -e LOCAL_USER_ID=0 -e LOCAL_USER_NAME=root -e LOCAL_GID=0 -v /media/psf/Home/projects/cadre/repos/fpp-ipk:/home/root:rw -w /home/root cadregitlab.jpl.nasa.gov:5050/cadre/cadre-fsw/cross-compile:latest /bin/bash in_container_build.bash
+ set -e
+ readonly BUILDROOT=/home/root/ipk/data
+ BUILDROOT=/home/root/ipk/data
+ opkg install openjdk_11.0.13+8_x86_64.ipk sbt_1.5.5_x86_64.ipk
Installing openjdk (11.0.13+8) on root.
Installing sbt (1.5.5) on root.
Configuring openjdk.
Configuring sbt.
+ rm -rf /home/root/.sbt
+ pushd /home/root/fpp
/home/root/fpp /home/root
+ git clean -fdx .
Removing compiler/bin/
Removing compiler/lib/src/main/scala/util/Version.scala.bak
Removing compiler/lib/target/
Removing compiler/project/build.properties
Removing compiler/project/project/
Removing compiler/project/target/
Removing compiler/target/
Removing compiler/tools/fpp-check/target/
Removing compiler/tools/fpp-depend/target/
Removing compiler/tools/fpp-filenames/target/
Removing compiler/tools/fpp-format/target/
Removing compiler/tools/fpp-from-xml/target/
Removing compiler/tools/fpp-locate-defs/target/
Removing compiler/tools/fpp-locate-uses/target/
Removing compiler/tools/fpp-syntax/target/
Removing compiler/tools/fpp-to-cpp/target/
Removing compiler/tools/fpp-to-xml/target/
+ git reset --hard HEAD
HEAD is now at 364c09d Update Docker build flag to support sbt 1.4.6
/home/root
+ popd
+ pushd /home/root/fpp/compiler
+ ./install
/home/root/fpp/compiler /home/root
Updating commit to 364c09d31bceb27ac1ff9d79229ce57bcb986d62
Building jar files
copying runtime jar...
[info] [launcher] getting org.scala-sbt sbt 1.5.5  (this may take some time)...
[info] [launcher] getting Scala 2.12.14 (for sbt)...
```

It strangely just suddenly quits with that above being all it outputs. Note I didn't downgrade my SBT version. Here's the `fpp` patch I just tried it against:

```diff
commit 364c09d31bceb27ac1ff9d79229ce57bcb986d62 (HEAD -> cadre)
Author: Viet T. Nguyen <vnguyen@jpl.nasa.gov>
Date:   Mon Nov 29 19:15:55 2021 -0800

    Update Docker build flag to support sbt 1.4.6

diff --git a/compiler/install b/compiler/install
index 3b180c13..70879f24 100755
--- a/compiler/install
+++ b/compiler/install
@@ -42,10 +42,7 @@ sed -i.bak -e "s/val commit = .*/val commit = \"$commit\"/" \
   $util/Version.scala

 echo "Building jar files"
-# The --batch and -Dsbt.server.forcestart=true flags were added to get the
-# build to work without user interaction in a Docker container. See
-# https://github.com/sbt/sbt/issues/6101#issuecomment-924554619.
-sbt --batch -Dsbt.server.forcestart=true assembly
+sbt -Dsbt.io.virtual=false -Dsbt.server.autostart=false assembly

 echo "Restoring Version.scala"
 cp $util/Version.scala.bak $util/Version.scala
```

Do you think downgrading SBT will make a difference?

bocchino:
Thanks for checking that out.

Would it work to add `$FPP_SBT_FLAGS` as the flags? Then by default the flags would be empty (should work for most users), and users who need special flags can set the environment variable. If we do this, we should document it in the README.

vietjtnguyen:
Yea I think that'd be perfect, just a way to forward flags to the `sbt` call.

bocchino:
OK, I'll add that.

bocchino:
This is done in commit 4b338768dc9318f377c8e9bca1fecc1c687d8eb6. Closing this PR. Let me know if there are any remaining issues.

Issue #104, opened 2021-12-22
Bug In JAR Wrapper Shell 


LeStarch:
The JAR shell wrappers that wrap the FPP tools have an error in them:

```
java -jar /Users/mstarch/code/fprime-infra/fpp/fpp-tools/fpp-locate-defs.jar $@
```

Should read as the following (with quotes around `"$@"`:

```
java -jar /Users/mstarch/code/fprime-infra/fpp/fpp-tools/fpp-locate-defs.jar "$@"
```

This informs bash to pass all arguments maintaining the escaping provided by the above call.  Without this, any escaping done in the caller is ignored. This is specifically painful to users with paths that require escaping (e.g.  `/home/user/my code/fprime`) will be broken into two partial paths by the existing code, but should be handled properly by the suggested code.

LeStarch:
Note: there is a symmetric error in the CMake integration that was masked by this error.

bocchino:
On the FPP side, this is not a bug/oversight. It is a conscious decision not to support path names with spaces in them.

If we want to support path names with spaces, then I think we will need a fair amount of specification, design, implementation, and testing that has not been attempted. For example:

. On the input side
   . What input should fpp-depend see if a user writes `fpp-depend a\ b.fpp c\ d.fpp`?
   . Same question for `fpp-depend "a b.fpp" "c d.fpp"?
   . Same question for `fpp-depend 'a b.fpp' 'c d.fpp'?

. On the output side
   . What should the tools do if a location file has a space in it? e.g., `a b.fpp`? For example, should fpp-depend write out the file as `a b.fpp` or `a\ b.fpp`? If the former, is every tool that uses the output of fpp-depend required to escape spaces before passing them to other tools? Are we doing that? Are we testing the use cases?

As far as I know, we are not using file names with spaces in F Prime. This suggests: (a) we don't need to handle it and (b) we are not testing the relevant use cases. I am not in favor of adding quotation marks to one or two places and hoping/guessing how the rest of it all works.

bocchino:
Note that if we put in the quotation marks as suggested, then case (1)(i) should work as one might expect, though I haven't tested it. I don't know what would happen in cases (1)(ii) and (1)(iii). Maybe case (1)(iii) would work reasonably. Case (1)(ii) almost certainly would not.

LeStarch:
[...]

PR #105, opened 2021-12-22
lestarch: adding release script, versioning, and fixes

LeStarch:
This adds 3 fixes:

1. It changes the version information to be use `git describe --tags --always`
2. It escapes shell arguments in the shell wrappers (#104)
3. It adds a `release` script that can build Graal JVM based native-images that increase fpp speed substantially.

Note: the above version change has the above desirable properties over the raw commit hash:

1. Collapses to <tag> when versioning an exact tag
2. Human readable (<last tag>-<commits past tag>-g<commit short has>)
2. Used elsewhere in F´ to capture version
3. Comparable to other versions with string greater and lesser operations

It retains the following properties:
1. Git hash is present
2. `git checkout <version>` works as expected

### DIFF
--- a/compiler/install
+++ b/compiler/install
@@ -42,7 +42,7 @@ fpp-to-cpp
 fpp-to-xml
 "
 
-commit=`git log | awk 'NR == 1 { print $2 }'`
+commit=`git describe --tags --always`
 util=lib/src/main/scala/util
 echo "Updating commit to $commit"
 sed -i.bak -e "s/val commit = .*/val commit = \"$commit\"/" \
@@ -65,6 +65,6 @@ do
   cp $jar $dest/$tool.jar
   echo '#!/bin/sh
 
-  java -jar '$dest'/'$tool'.jar $@' > $dest/$tool
+  java -jar '$dest'/'$tool'.jar "$@"' > $dest/$tool
   chmod +x $dest/$tool
 done
--- a/compiler/install
+++ b/compiler/install
@@ -42,7 +42,7 @@ fpp-to-cpp
 fpp-to-xml
 "
 
-commit=`git describe --tags --always`
+commit=`git log | awk 'NR == 1 { print $2 }'`
 util=lib/src/main/scala/util
 echo "Updating commit to $commit"
 sed -i.bak -e "s/val commit = .*/val commit = \"$commit\"/" \
--- a/compiler/release
+++ b/compiler/release
@@ -1,0 +1,55 @@
+#!/bin/sh -e
+
+# ----------------------------------------------------------------------
+# FPP release builder
+# ----------------------------------------------------------------------
+# To pass flags to sbt, set the environment variable
+# FPP_SBT_FLAGS
+# ----------------------------------------------------------------------
+
+DIRECTORY=`dirname $0`
+FAST_BIN="${DIRECTORY}/fast-fpp-`uname`-`uname -m`"
+NATIVE_IMAGE=`which native-image || exit 0`
+JAVA_PATH=`which java || exit 0`
+
+# Looking for native-image
+if [ ! -x "${NATIVE_IMAGE}" ]
+then
+    echo "[ERROR] native-image not found, install Graal JVM, native-image, and ensure it runs on your system" 1>&2
+    exit 1
+fi
+JAVA_DIR=`dirname "${JAVA_PATH}"`
+NATIVE_DIR=`dirname "${NATIVE_IMAGE}"`
+
+if [ "${JAVA_DIR}" != "${NATIVE_DIR}" ]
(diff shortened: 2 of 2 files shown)


compiler/install
```diff
@@ -42,7 +42,7 @@ fpp-to-cpp
 fpp-to-xml
 "
 
-commit=`git log | awk 'NR == 1 { print $2 }'`
+commit=`git describe --tags --always`
```
bocchino, line 45: I think we need to use the full commit hash here, for development purposes. On the devel branch, we will need to check in dependencies to specific commits ahead of the last release tag. (We did this frequently while developing release 3.0.) Also, I am reluctant to introduce alternate/ambiguous ways to refer to the same commit. If we use the full has, there's no ambiguity (no danger of hash conflict, and no danger that a string comparison against the expected commit will fail in a surprising way).

compiler/install
```diff
@@ -65,6 +65,6 @@ do
   cp $jar $dest/$tool.jar
   echo '#!/bin/sh
 
-  java -jar '$dest'/'$tool'.jar $@' > $dest/$tool
+  java -jar '$dest'/'$tool'.jar "$@"' > $dest/$tool
```
bocchino, line 68: See my comments on #104. I'll accept this change, but I'm skeptical that it fixes anything. It seems to invite more problems/issues than it solves.

compiler/release
```diff
@@ -0,0 +1,55 @@
...
+
+# ----------------------------------------------------------------------
+# FPP release builder
+# ----------------------------------------------------------------------
+# To pass flags to sbt, set the environment variable
+# FPP_SBT_FLAGS
+# ----------------------------------------------------------------------
+
+DIRECTORY=`dirname $0`
+FAST_BIN="${DIRECTORY}/fast-fpp-`uname`-`uname -m`"
```
bocchino, line 11: Can we use the prefixes NATIVE_ and native- instead of fast? It's more precise. The native build should be faster, but all we can guarantee is that it's a native build.

bocchino (changes requested):
Looks good! Just a few comments.

compiler/install
LeStarch: An alternative approach would be to replace the hardcoded version (e.g. v1.0.0) in the string with this version, and leave the commit hash so we have both.  Using this alternate, the released versions would look identical to what we have now...but the interim versions would include some extra information.

**Tag version Example**
```
fpp-depend v1.0.0 commit aabbccddeeff....
```

**Intermin Example**
```
fpp-depend v1.0.0-gaabbccdd commit aabbccddeeff....
```

This is the best of both worlds.  The exact hash is there but a human readable version+ is also available.  It also means we wouldn't have to remember to manually bump that version....

LeStarch: ...and it would give us a version in the same format that F´ uses to report versions.

bocchino: On further investigation, I think the full commit hash is not necessary. It looks like (1) `git describe` provides enough info to reconstruct the exact commit and (2) it does so in a form that other commands (e.g., `git checkout`) will accept.

Maybe we should replace the version stamp and the commit hash with the output of `git describe`. For example, at HEAD^ from v1.0.0, `git describe` says this:
```
[I] bocchino@MT-306179 ~/J/T/fpp ((23165a51…))> git describe --tags --always
v0.3.0-616-g23165a51
```
In this case maybe `fpp-depend --help` should say
```
fpp-depend v0.3.0-616-g23165a51
```
instead of what it says now
```
fpp-depend v1.0.0 commit 23165a5173848397743edffc675c94b0c57d58c2
```
We will also need to make the corresponding change to the version stamp in the docs. That is stored here:

https://github.com/fprime-community/fpp/blob/ee355fc99eb8040157c62e69f58ac6a8435cd981/defs.sh#L5

Since this is a nontrivial change to the version stamping mechanism of FPP, I'd like to move it out of this PR and into a separate issue or PR. Let's revert the change on this PR.

bocchino: I guess the 616 means "616 commits ahead of v0.3.0." I'm not sure what the `g` is doing in there ahead of the hash prefix. But ultimately it doesn't matter, if git knows unambiguously what it means.

LeStarch: Sure, I can revert it out of the PR.  FYI the "g" means "git" to signify that those last 8 digits are a git commit hash in shortened form.  This was done such that the same tag format could be generated by some other source control using a unique identifier in their native format.

LeStarch:
@bocchino reverted out versioning.  I can do another PR for the version change if you'd like, or I can leave you to it.

bocchino (approved):
This PR looks good. I verified that the Graal VM script works.

PR #106, opened 2022-01-04
updating documentation. nested namespaces are now working

kevin-f-ortega:
Updating documentation. As I was using fpp, I realized I could use nested modules/namespaces.

### DIFF
--- a/docs/users-guide/Defining-Modules.adoc
+++ b/docs/users-guide/Defining-Modules.adoc
@@ -103,11 +103,7 @@ module M {
 }
 ----
 
-*XML limitation:* The F Prime XML format supports
-only one level of {cpp} namespaces.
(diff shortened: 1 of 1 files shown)


bocchino:
I'm pretty sure there has been no progress on nested namespaces since we wrote those docs. For example, if the namespace is A::B, this will still generate broken code:

https://github.com/nasa/fprime/blob/0b669e108b43bc61c699acf4925ac783fbdc7f12/Autocoders/Python/src/fprime_ac/generators/templates/enums/enum_hpp.tmpl#L24-L25

The current plan is fix this when we replace the Python autocoder with the FPP code generator.

@LeStarch FYI

kevin-f-ortega:
Ah okay. Yes, I do see those broken include guards.

bocchino:
Closing for now, until fixed.

PR #111, opened 2022-02-01
Update version stamps

### LINKED ISSUE #108: Revise version stamp
Instead of manually updating the version number, use `git-describe` as described in https://github.com/fprime-community/fpp/pull/105.

bocchino:
Closes #108 

@LeStarch please take a look. It uses `git describe` as discussed. The regex match in the F Prime build will need to change a bit.

### DIFF
--- a/compiler/install
+++ b/compiler/install
@@ -42,10 +42,10 @@ fpp-to-cpp
 fpp-to-xml
 "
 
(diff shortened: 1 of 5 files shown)


LeStarch (approved):
Does what I expect it to. LGTM.

PR #112, opened 2022-03-03
Fix indent in vim for fpp files

kevin-f-ortega:
these lines are needed to avoid vim from indenting 8 spaces.
Instead, vim will indent by 2 spaces. Also, tabs will be treated as 2 spaces

### DIFF
--- a/editors/vim/fpp.vim
+++ b/editors/vim/fpp.vim
@@ -6,6 +6,8 @@
 "
 setl cinwords=component,enum,module,struct,topology
 setl cindent
+setl shiftwidth=2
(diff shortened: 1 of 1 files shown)


bocchino:
Closing this one for now. I think vim has bugs. Indentation is supposed to be handled according to cinoptions:

http://vimdoc.sourceforge.net/htmldoc/indent.html

In particular, the default indentation is supposed to follow the existing shift width. Instead it seems to go to 6 characters. Setting the shift width to 2 (a) should not be necessary and (b) does not have the desired effect on my system.

PR #115, opened 2022-04-11
removes copyright notice in generated files.

### LINKED ISSUE #114: Remove Copyright Notices in fpp-to-cpp
In F´ we have decided to strip copyright headers from generated files.  This frees projects using F´ from the burden of a caltech copyright on files they have generated. `fpp-to-cpp` is still adding this information to files it generates.

LeStarch:
Fixes #114


bocchino (changes requested):
Looks good, but this is going to break unit tests. We need to make sure all the tests pass before merging.

bocchino:
It should be sufficient to run `./update-ref` in `fpp/compiler/tools/fpp-to-cpp/test`.

bocchino:
Also if we are no longer generating copyright info, we should remove this function

https://github.com/fprime-community/fpp/blob/a8705b3ec490f95a80379e88bc21bae4b7401e81/compiler/scripts/test-utils.sh#L18-L22

and all of its uses.

LeStarch:
Ok!  I can run those tests!

LeStarch:
@bocchino I think this is now corrected per your comments.  I removed the unnecessary function, updated the references, and tested with:

```
cd compiler
./install
./test
...
...
...
All tests PASSED
```

bocchino (approved):
Looks good!

PR #125, opened 2022-05-20
making fpp install location movable

LeStarch:
Allows the install FPP Scala tools directory to be portable.  This is needed to allow installing to one location and then moving the directory afterwards, which is part of the installation process when installing non-native tools.

### DIFF
--- a/compiler/install
+++ b/compiler/install
@@ -67,7 +67,6 @@ do
   echo "  $jar"
   cp $jar $dest/$tool.jar
   echo '#!/bin/sh
-  LOCAL_DIR="`dirname $0`"
-  java -jar "${LOCAL_DIR}/'$tool'.jar" "$@"' > $dest/$tool
+  java -jar "dirname $0`/'$tool'.jar" "$@"' > $dest/$tool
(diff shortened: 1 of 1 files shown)


LeStarch:
Example of new script:

```sh
#!/bin/sh
  LOCAL_DIR="`dirname $0`"
  java -jar "${LOCAL_DIR}/fpp-depend.jar" "$@"
```

bocchino (commented):
Looks good, but can we delete the definition of LOCAL_DIR and write 
```
`dirname $0`
```
instead of `${LOCAL_DIR}`?

In general we should adhere to the shell programming convention that CAPITAL_CASE is for environment variables and shell variables, and script-local variables are in snake_case.

LeStarch:
Alright fixed.  In general, the F´ convention is to not ship shell scripts at all, as they aren't truly cross-platform.  We've been systemically replacing them with true cross-platform equivalents for years.

bocchino:
Thanks! OK, let's discuss in a separate context how we can replace these scripts.

bocchino (approved):
Looks good.

PR #130, opened 2022-06-28
Updates cpp banner comment to include the tool it was generated from

legounicycler:
Tested in Fpp with the ./test script


bocchino (approved):

PR #132, opened 2022-06-29
CppWriter function qualifiers

### LINKED ISSUE #131: Fix qualifiers for function names and return types for C++ codegen
In function CppDocCppWriter.visitFunction()

tiffany1618:
Resolves #131

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/CppDocCppWriter.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/CppDocCppWriter.scala
@@ -95,11 +95,10 @@ object CppDocCppWriter extends CppDocWriter {
         val retType = function.retType.getCppType match {
           case "" => ""
           case t => s"$t "
-
         }
         in.classNameList match {
-          case head :: _ => {
-            val line1 = line(s"$retType$head ::")
+          case _ :: _ => {
+            val line1 = line(s"$retType${in.getEnclosingClassQualified} ::")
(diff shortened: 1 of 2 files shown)


compiler/lib/src/main/scala/codegen/CppWriter/CppDocCppWriter.scala
```diff
@@ -95,11 +95,10 @@ object CppDocCppWriter extends CppDocWriter {
         val retType = function.retType.getCppType match {
           case "" => ""
           case t => s"$t "
-
         }
         in.classNameList match {
-          case head :: _ => {
-            val line1 = line(s"$retType$head ::")
+          case _ :: _ => {
```
bocchino, line 100: I think we can change line 100 to `case _ => {`.

bocchino (commented):
Looks great! Just one comment.

compiler/lib/src/main/scala/codegen/CppWriter/CppDocCppWriter.scala
bocchino: You will also need to put the `Nil` case first to make this work. I think it's a little clearer than writing `_ :: _` to match against a list of size >= 1.

bocchino: However, I see that the `_ :: _` pattern is used elsewhere in the same file. So let's keep it!

bocchino (approved):
Good work!

PR #139, opened 2022-07-08
Use U8 as the serialized size of bool

bocchino:
`sizeof(bool)` is not defined in C++.  See https://stackoverflow.com/questions/4897844/is-sizeofbool-defined-in-the-c-language-standard.

F Prime serializes `bool` as U8. See https://github.com/nasa/fprime/blob/776af784414b8df7c148845dfca8a1956e5ac4cb/Fw/Types/Serializable.cpp#L210-L225.


tiffany1618 (approved):
Looks good!

PR #140, opened 2022-07-08
Revise generated C++ headers

bocchino:
* Don't generate headers for constant symbols. The names of constant symbols are resolved away before code gen.
* Write out all user (non-system) headers as a single sorted list.


tiffany1618 (approved):
Looks good!

PR #141, opened 2022-07-12
Fix C++ array toString() member function

tiffany1618:
- Directly call toChar() instead of toString() on string type members in generated toString() member function
- Fix C++ comments


bocchino:
This PR is an optimization. It saves allocating memory for a new string and copying over the string value.

bocchino (approved):

PR #143, opened 2022-07-14
C++ struct writer

### LINKED ISSUE #137: Implement C++ structs
Implement C++ code gen for structs

### LINKED ISSUE #138: Test C++ structs
Test C++ code gen for structs

tiffany1618:
Closes #137
Closes #138 

Added
- `StructCppWriter`: Writes C++ for FPP struct definitions
- Unit tests for C++ struct writer

Changes from Python XML struct autocoder:
- Add const and non-const getters for arrays and serializable type members, excluding enum types, which are passed by value (resolves issue of getting and then setting a struct member; e.g. a pattern like `s.get_a().set_b()`, where `s` is a struct containing a struct type member `a`, which in turn contains another member `b`)
- Pass array type members by C++ reference instead of by pointer and size (simpler, enforces checking of array length)
- Remove pointer copy constructor (reference copy constructor is sufficient)


tiffany1618:
fpp-to-xml gives an error for empty structs, but fpp-to-cpp currently doesn't. Do we want fpp-to-cpp to also give an error?

bocchino:
> fpp-to-xml gives an error for empty structs, but fpp-to-cpp currently doesn't. Do we want fpp-to-cpp to also give an error?

No. Empty structs are allowed in FPP and in C++. The error in fpp-to-xml exists because the XML schema does not support them. So this is a case where we are implementing part of the FPP spec that is not implemented when going through XML.

compiler/lib/src/main/scala/codegen/CppWriter/StructCppWriter.scala
```diff
@@ -0,0 +1,775 @@
...
+          CppDoc.Lines.Both
+        )
+      ),
+      CppDoc.Class.Member.Constructor(
+        CppDoc.Class.Constructor(
+          Some("Constructor (default value)"),
+          Nil,
+          "Serializable()" :: nonArrayMemberNames.map(n => {
+            val defVal = defaultValues(n)
+            // Parse non-literal default values, e.g. T() for abstract type T
```
bocchino, line 236: Why do we need this special-case handling? If we need it, let's not parse string output. Let's match against FPP semantic concepts, e.g., types.

bocchino (commented):
Looks great! I have one comment so far. I'll schedule a review to go over the details.

bocchino:
Also, same comment as on #144 -- please re-run all unit tests and update them as necessary.

compiler/lib/src/main/scala/codegen/CppWriter/StructCppWriter.scala
bocchino: Looks good!

bocchino (approved):
Looks good! I confirmed that the unit tests pass.

PR #144, opened 2022-07-15
Add array code gen to ComputeCppFiles

### LINKED ISSUE #142: Add array code gen to ComputeCppFiles
- [ ] Record that we are generating arrays. Similarly to what we do for enums here:

https://github.com/fprime-community/fpp/blob/a2f2ac062a2da1ef849e8a001307a137857a0de1/compiler/lib/src/main/scala/codegen/CppWriter/ComputeCppFiles.scala#L55-L63

- [ ] Add a unit test. It can be similar to the duplicate test for XML structs.
- [ ] For completeness add a unit test for duplciate enums too.

tiffany1618:
Resolves #142 

- Record generation of C++ array files in ComputeCppFiles
- Add unit tests for duplicate arrays and enums

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ComputeCppFiles.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ComputeCppFiles.scala
@@ -52,6 +52,16 @@ object ComputeCppFiles extends AstStateVisitor {
     aNode: Ast.Annotated[AstNode[Ast.DefConstant]]
   ) = addMappings(s, FileNames.getConstants, None)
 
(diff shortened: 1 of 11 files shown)


bocchino (changes requested):
Looks good!

Please rerun all the unit tests and run `update-ref` as necessary. In particular, the unit tests for `fpp-filenames` and `fpp-depend` have to be updated. Then we will have regression tests for the changes to the generated file names.

bocchino (approved):
Looks great! I confirmed that all tests pass.

PR #146, opened 2022-07-21
Generate C++ code for types inside components

### LINKED ISSUE #145: Generate C++ code for types inside components
fpp-to-cpp isn't generating code for types declared inside components, e.g.,
```
passive component C { enum E { X } }
```
In this example, fpp-to-cpp should generate a C++ enum `C_E`, not enclosed in any namespace.

- [x] Add a Component visitor to CppWriter to handle this case
- [x] Add unit tests for enums, arrays, and structs

tiffany1618:
Resolves #145 

- Add `defComponentAnnotatedNode()` to `CppWriter`
- Remove namespace generated by enclosing component (mirrors [XML writer namespace generation](https://github.com/fprime-community/fpp/blob/98820fe7676752d69bf81775b5666efb4bfd0764/compiler/lib/src/main/scala/codegen/XmlWriter/XmlWriterState.scala#L84-L100))
- Add unit tests for enums, structs, and arrays inside a component


bocchino:
Looks great! A couple of suggestions:

1. I think we need to make a corresponding revision to `getNamespace` in CppWriterState: https://github.com/tiffany1618/fpp/blob/7db2adf53cb08a1b85671815814ec4f5c6b34621/compiler/lib/src/main/scala/codegen/CppWriter/CppWriterState.scala#L76-L80
2. I think we can refactor XmlWriter to call CppWriterState.getNamespace and remove XmlWriterState.getNamespace.

bocchino:
We could also implement getNamespace using getNamespaceIdentList: just call getNamespaceIdentList and join the elements with '::'. Then all the special-case behavior will be in getNamespaceIdentList.

tiffany1618:
Ok, those sound good. The `writeSymbol` function also defines [`removeComponentQualifers`](https://github.com/tiffany1618/fpp/blob/98820fe7676752d69bf81775b5666efb4bfd0764/compiler/lib/src/main/scala/codegen/CppWriter/CppWriterState.scala#L104-L118) that does the same thing as [`getQualifierList`](https://github.com/tiffany1618/fpp/blob/7db2adf53cb08a1b85671815814ec4f5c6b34621/compiler/lib/src/main/scala/codegen/CppWriter/CppWriterState.scala#L85-L96), so I could also make that a separate private function in CppWriterState.

bocchino:
Sounds good. The higher-level principles here are:

1. In FPP analysis, we'll keep everything clean and uniform regarding parent symbols and qualified names.
2. In the C++ codegen there's some weirdness because we have to generate `component C { enum E }` as C_E instead of C::E. Let's confine that weirdness to the CppWriter, by replacing functions in Analysis with specialized functions in the CppWriter as necessary.
3. Eventually (after this round of code gen) I'd like to revise the C++ code gen so that we generate `component C { enum E }` as a C++ class C with an inner definition E. That's for later though. The current XML-based code gen does not support this at all. The first step is to get the FPP C++ code gen to do more or less what the XML code gen is doing.

tiffany1618:
I see, that makes sense.

bocchino (approved):
Looks great!

PR #147, opened 2022-07-21
Add line utilities for C++ writer

tiffany1618:
- Add `CppWriterLineUtils.wrapInIfDirective()`: generic utility for wrapping members in an if directive
- Add `CppWriterLineUtils.writeOstreamOperator()`: utility for writing operator<< function
- Refactor enum, array, and struct writers to use those utilities
- Add operator member section with banner comment to struct classes


tiffany1618:
@bocchino Ok, I refactored the cpp writers and added the extra parameter to the if directive utilities.

bocchino (approved):
Looks great! I confirmed that the tests pass.

PR #150, opened 2022-07-22
Fix ambiguous use of overloaded operator== error for C++ enums

### LINKED ISSUE #148: Fix ambiguous use of overloaded operator== error for C++ enums
Comparing an enum class type with its raw enum type using the == or != operator gives an ambiguous use error.

For example, using the C++ header generated from this FPP model:

```
enum E { X }
```

to compile this program:

```
#include "EEnumAc.hpp"
 
bool f(E e) {
  return e == E::X;
}
```

results in this error:

```
error: use of overloaded operator '==' is ambiguous (with operand types 'E' and 'E::T')
  return e == E::X;
         ~ ^  ~~~~
```

The ambiguity is a result of the conversion operator that converts an enum class type to a raw enum type; to fix this and keep the conversion operator, the operator== and != functions can be changed to take in a raw enum type instead of an enum class type.

tiffany1618:
Resolves #148 

- Change operator== and operator!= to take in raw enum type T instead of enum class type
- Change conversion operator type from t to T
- Update unit tests

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/EnumCppWriter.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/EnumCppWriter.scala
@@ -280,7 +280,7 @@ case class EnumCppWriter(
       CppDoc.Class.Member.Function(
         CppDoc.Function(
           Some(s"Conversion operator"),
-          s"operator t",
+          s"operator T",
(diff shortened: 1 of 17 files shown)


bocchino (approved):

PR #151, opened 2022-07-22
Compute C++ files after checking semantics

### LINKED ISSUE #149: Compute C++ files after checking semantics
C++ files are computed using the Analysis data structure, but they are being computed before the current Analysis instance is populated by `CheckSemantics.tuList()`: https://github.com/fprime-community/fpp/blob/98820fe7676752d69bf81775b5666efb4bfd0764/compiler/tools/fpp-to-cpp/src/main/scala/fpp-to-cpp.scala#L38-L47

This causes errors like the following:

 An FPP model containing

```
passive component C {
  enum E { X }
}

enum E { X }
```

causes fpp-to-cpp to generate a duplicate C++ file error even though there should be no name duplication here (the generated enum classes would be C_E and E).

tiffany1618:
Resolves #149 

- Move call to `ComputeCppFiles.visitList()` to after call to `CheckSemantics.tuList()`
- Add unit tests


bocchino (approved):
Looks good!

PR #154, opened 2022-08-04
Remove typedef enum from generated C++ code

### LINKED ISSUE #153: Remove typedef enum from generated C++ code
It has been pointed out that in C++, instead of writing
```
typedef enum { ... } E;
```
we should write
```
enum E { ... };
```

tiffany1618:
Closes #153


bocchino (approved):
Looks good!

PR #156, opened 2022-08-15
Update array unit tests for C++ writer 

tiffany1618:
- Replace generated struct files from XML autocoder with generated struct files from fpp-to-cpp


compiler/tools/fpp-to-cpp/test/array/S1SerializableAc.ref.hpp
```diff
@@ -0,0 +1,302 @@
...
+      // ----------------------------------------------------------------------
+
+      class StringSize80 :
+        public Fw::StringBase
+      {
+
+        public:
+
+          enum {
+            SERIALIZED_SIZE = 80 + sizeof(FwBuffSizeType) //!< Size of buffer + storage of two size words
```
bocchino, line 33: I just notice this: Why does it say "Size of buffer + two size words"? Shouldn't it be "Size of string buffer plus size of string length"?

bocchino (commented):
Looks good, just one comment.

compiler/tools/fpp-to-cpp/test/array/S1SerializableAc.ref.hpp
tiffany1618: Yes that would be clearer -- I think the "two words" was because the type of FwBuffSizeType is U16.

bocchino: OK, let's make a separate issue on this.

bocchino (approved):
Looks good!

PR #158, opened 2022-08-16
Revise comment in generated C++

### LINKED ISSUE #157: Revise generated comment in Cpp Writer
This comment:

https://github.com/fprime-community/fpp/blob/4ab4edaee3bae19c4881f2cf3b6c3f06f213330f/compiler/lib/src/main/scala/codegen/CppWriter/StringCppWriter.scala#L80

could be made clearer. It should say "size of string buffer plus size of string length."

bocchino:
Closes #157.

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/StringCppWriter.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/StringCppWriter.scala
@@ -76,9 +76,14 @@ case class StringCppWriter(
         CppDoc.Lines(
           List(
             CppDocHppWriter.writeAccessTag("public"),
-            addBlankPrefix(wrapInEnum(lines(
-              s"SERIALIZED_SIZE = $size + sizeof(FwBuffSizeType) //!< Size of buffer + storage of two size words"
-            ))),
+            addBlankPrefix(
+              wrapInEnum(
+                List(
+                  line("//!< The size of the string length plus the size of the string buffer"),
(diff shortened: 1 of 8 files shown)


compiler/lib/src/main/scala/codegen/CppWriter/StringCppWriter.scala
```diff
@@ -76,9 +76,14 @@ case class StringCppWriter(
         CppDoc.Lines(
           List(
             CppDocHppWriter.writeAccessTag("public"),
-            addBlankPrefix(wrapInEnum(lines(
-              s"SERIALIZED_SIZE = $size + sizeof(FwBuffSizeType) //!< Size of buffer + storage of two size words"
-            ))),
+            addBlankPrefix(
+              wrapInEnum(
+                List(
+                  line("//!< The size of the string length plus the size of the string buffer"),
```
tiffany1618, line 82: I think the `<` shouldn't be there since the comment refers to the line below it.

tiffany1618 (commented):
Looks good, just one comment.

compiler/lib/src/main/scala/codegen/CppWriter/StringCppWriter.scala
bocchino: Good catch, fixed.

tiffany1618 (approved):
Looks good!

PR #161, opened 2022-08-17
Make struct member variables protected instead of private in C++ writer

### LINKED ISSUE #159: Make struct member variables protected in generated C++
The access specifier for FPP struct class member variables should be changed to `protected` instead of `private`.

tiffany1618:
Closes #159


bocchino:
This issue responds to a review of the generated code held on 8/16. Rationale: `protected` allows direct access to member variables by classes that inherit from the generated class. This access is convenient.

bocchino (approved):
Looks good!

PR #162, opened 2022-08-17
Inline small functions in generated C++

### LINKED ISSUE #160: Inline small functions in generated C++
Consider inlining enum functions, such as:

Enums:
- Constructors
- `operator T()`
- `operator==()`
- `operator!=()`

Structs:
- Getter functions

tiffany1618:
Closes #160


bocchino:
Per the review held on 8/16. Rationale: inlining these function should improve the performance of these functions, at a very small cost in code size.

compiler/lib/src/main/scala/codegen/CppWriter/EnumCppWriter.scala
```diff
@@ -189,43 +189,46 @@ case class EnumCppWriter(
...
-          List(
-            CppDoc.Function.Param(
-              CppDoc.Type("const T"),
-              "e",
-              Some("The raw enum value")
-            )
-          ),
-          Nil,
-          lines("this->e = e;")
+      CppDoc.Class.Member.Lines(
```
bocchino, line 206: Can we merge the adjacent Lines members into a single member? I think this would make the code cleaner.

compiler/lib/src/main/scala/codegen/CppWriter/EnumCppWriter.scala
```diff
@@ -277,57 +280,42 @@ case class EnumCppWriter(
...
-              Some("The other enum value"),
-            ),
-          ),
-          CppDoc.Type("bool"),
-          List(
-            line("return this->e == e;"),
-          ),
-          CppDoc.Function.NonSV,
-          CppDoc.Function.Const
+      CppDoc.Class.Member.Lines(
```
bocchino, line 295: Same comment as above.

bocchino (commented):
Great job! Please see comments. Also, please confirm that everything passes on the F Prime branch with the FPP unit tests.

compiler/lib/src/main/scala/codegen/CppWriter/EnumCppWriter.scala
tiffany1618: Ok, fixed! Both F Prime and FPP functional unit tests still pass.

bocchino (approved):
Thanks, looks great!

PR #167, opened 2022-10-13
Replace Fw/Types/BasicTypes.hpp includes with FpConfig.hpp

### LINKED ISSUE #152: Include FpConfig.hpp directly in generated C++ code
F Prime will soon replace includes of `BasicTypes.hpp` with includes of `FpConfig.hpp`. See https://github.com/nasa/fprime/pull/1607. We need to revise the generated C++ code to match.

tiffany1618:
Resolves #152 

- Update `fprime-gcc` script to include `cmake/platform/types` instead of `Fw/Types/Linux`
- Remove old files from `compiler/tools/fpp-to-cpp/test/constants/output_dir`, update `clean` script to remove all files ending in `*Ac.hpp` or `*Ac.cpp`

Passes F Prime unit tests.


bocchino (approved):
Looks great!

PR #179, opened 2022-10-26
Fix bug in XML code gen

### LINKED ISSUE #178: fpp-to-xml inserts extra quotation marks in string attributes
See https://github.com/nasa/fprime/discussions/1733. When generating a string value inside an XML attribute, we need to strip off the quotation marks.

bocchino:
## Overview

This PR fixes a bug in the generation of XML attributes that represent C++ literal string values. We need to delete the enclosing quotation marks from the string value, so the F Prime autocoder can re-insert them. Also, we need to use XML escapes for embedded quotation marks in this case.

## Changes

* Fix XML code gen
* Update unit tests to cover the fix

## Associated Issues

Closes #178.


tiffany1618 (approved):
Looks good!

PR #180, opened 2022-10-28
lestarch: fixing graalvm build

LeStarch:
@bocchino @tiffany1618 this should fix the graal VM build.

### DIFF
--- a/compiler/release
+++ b/compiler/release
@@ -57,7 +57,7 @@ for jar_file in $directory/bin/*.jar
 do
     out_file=$native_bin/`basename $jar_file .jar`
     echo "Building $out_file"
-    "$native_image" --no-fallback --install-exit-handlers -jar "$jar_file" "$out_file"
+    "$native_image" --configure-reflection-metadata --no-fallback --install-exit-handlers -jar "$jar_file" "$out_file"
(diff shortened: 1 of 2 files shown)


LeStarch:
Also, those warnings are from the fact that all classes are not used in all native images.  Thus they really do not exist, and the warning are innocuous.

LeStarch:
Once merged, we should build an `a3` release and bump the F´  version to test CI.

compiler/release
```diff
@@ -76,6 +76,15 @@ sync; sync; sync;
 rm "$native_bin"/*.txt
 sync; sync; sync;
 
+# Testing of the native binaries
+mv "$directory/bin" "$directory/bin.bak"
```
bocchino, line 80: Let's add `rm -rf bin.bak` here. Otherwise the build will fail if bin.bak exists already.

bocchino (commented):
Looks good! Just one comment.

LeStarch:
@bocchino I just changed the move to a `rm`.  Since release runs primarily in CI, there isn't a lot driving us to keep these files around.

bocchino (approved):
Looks good!

PR #181, opened 2022-10-29
Revise release script

bocchino:
* Improve diagnostic messages.
* Improve validation of release archive. Use diff instead of size estimate to check correctness.
* Revise the directory structure in the release archive. On my Mac, I noticed that we were giving absolute paths starting with / to tar. But these paths are meaningless in a release tar ball. My Mac was warning about it.
* Bump the GraalVM version to 22.3.0. This also fixes the problem with the missing trace info.
* In GraalVM 22.3.0, the trace info that we added by hand caused a warning, so I deleted it.
* Delete the code that runs all the JVM unit tests and captures the trace. Since we are checking the trace into the repo, and it seems brittle to changes in FPP and/or GraalVM, the procedure will be to run the trace manually, verify that everything is working, and commit any changes to the trace before submitting a release to GitHub.
* Revert the GraalVM README to what we had in the previous release. Hopefully, with the latest GraalVM, we won't have to hand-edit trace files.
* Use gtar instead of tar on Mac OS -- tar does not work properly.
* Switch from ubuntu-latest to ubuntu-18.04 for compatibility.
* Add and document an FPP_NATIVE_IMAGE_FLAGS configuration variable.

### DIFF
--- a/.github/workflows/build-native.yml
+++ b/.github/workflows/build-native.yml
@@ -6,7 +6,7 @@ jobs:
   Build-Native-Images:
     strategy:
       matrix:
-        os: ["ubuntu-latest", "macos-11.0"]
+        os: ["ubuntu-18.04", "macos-10.15"]
     runs-on: ${{ matrix.os }}
     steps:
     - name: "Checkout FPP Repository"
--- a/.github/workflows/build-native.yml
+++ b/.github/workflows/build-native.yml
@@ -6,7 +6,7 @@ jobs:
   Build-Native-Images:
     strategy:
       matrix:
-        os: ["ubuntu-latest", "macos-10.15"]
+        os: ["ubuntu-latest", "macos-11.0"]
     runs-on: ${{ matrix.os }}
     steps:
     - name: "Checkout FPP Repository"
--- a/.github/workflows/env-setup
+++ b/.github/workflows/env-setup
@@ -1,5 +1,5 @@
 #!/bin/bash
-graal_ver="22.1.0"
+graal_ver="22.2.0"
 graal_os="$( uname -s | tr "[:upper:]" "[:lower:]")"
 graal_ar="graalvm-ce-java11-${graal_os}-amd64-${graal_ver}.tar.gz"
 graal_dir="$(pwd)/graalvm-ce-java11-${graal_ver}"
--- a/.github/workflows/env-setup
+++ b/.github/workflows/env-setup
@@ -1,5 +1,5 @@
 #!/bin/bash
-graal_ver="22.2.0"
+graal_ver="22.3.0"
 graal_os="$( uname -s | tr "[:upper:]" "[:lower:]")"
 graal_ar="graalvm-ce-java11-${graal_os}-amd64-${graal_ver}.tar.gz"
 graal_dir="$(pwd)/graalvm-ce-java11-${graal_ver}"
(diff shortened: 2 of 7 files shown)


bocchino:
The archive creation failure appears to be a known issue with the Mac OS virtual environment. See https://github.com/actions/runner-images/issues/2619.

bocchino:
> The archive creation failure appears to be a known issue with the Mac OS virtual environment.

Attempting to fix by using gtar if available.

bocchino:
I successfully got release v1.1.0a6 through the GitHub build process. Whew!

https://github.com/fprime-community/fpp/releases/tag/v1.1.0a6

@tiffany1618 This release should be good to test with. I used it to build F Prime on your branch.

LeStarch:
@bocchino we should switch the runner to `ubuntu-18.04` for similar reasons to using the latest macOS. In the long-term I do not think this method of distributing FPP is sustainable as the runners to build these tools are deprecating faster than the platforms we need to support.

LeStarch:
Note: building on 18.04 has kernel version 4.15 which is old enough for our Redhat customers.

LeStarch:
@bocchino also, why not bump to 22.3.0?

compiler/release
```diff
@@ -12,87 +12,138 @@
...
-$directory/install-trace
-# Test a run before running test
-$directory/bin/fpp-locate-defs tools/fpp-locate-defs/test/defs/defs-1.fpp tools/fpp-locate-defs/test/defs/defs-2.fpp 1>/dev/null
-if [ $? -ne 0 ]
-then
-    echo "[ERROR] Failed to run JAR with tracing enabled"
-    exit 1
-fi
+# Install jar files in bin
+./install
```
LeStarch, line 79: Don't we need to run the trace of the tools?

.github/workflows/env-setup
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-graal_ver="22.1.0"
+graal_ver="22.2.0"
```
LeStarch, line 2: There is a newer version that may be more effective.

.github/workflows/build-native.yml
```diff
@@ -6,7 +6,7 @@ jobs:
   Build-Native-Images:
     strategy:
       matrix:
-        os: ["ubuntu-latest", "macos-10.15"]
+        os: ["ubuntu-latest", "macos-11.0"]
```
LeStarch, line 9: As it turns out `ubuntu-latest` is too new for many linux users (Redhat).  These need to build on `18.04`, which will soon expire.

LeStarch (commented):
Just some comments.  Nothing blocking.

LeStarch:
I might also add an argument for extra `native-image` flags.  For example, the issue I was seeing on the redhat was `/tmp` was mounted as `noexec`.  Thus I needed to supply an alternative `/tmp` using `-H:TempDirectory=/path/to/something/executable`.

compiler/release
bocchino: The current plan is to run the trace manually and correct any issues before submitting the release to GitHub. The trace results are checked in, so running the trace on GitHub should not be necessary. We run the unit tests in on GitHub on the native binaries, so if anything goes wrong the release process will catch it.

bocchino:
> I might also add an argument for extra native-image flags.

Done.

.github/workflows/build-native.yml
bocchino: Switched to ubuntu-18.04 and switched back to macos-10.15.

.github/workflows/env-setup
bocchino: I re-ran the trace with this version, committed the changes, and updated the version in the env setup.

LeStarch:
I ran on 22.3.0 last night and everything went fine.

PR #187, opened 2022-11-04
Revise include headers in generated C++

### LINKED ISSUE #184: Revise include headers in generated C++
This code needs a small change to the way that it gets the include path:

https://github.com/fprime-community/fpp/blob/a540390083cce051eacb06cbfd51bed1a3483e79/compiler/lib/src/main/scala/codegen/CppWriter/CppWriterState.scala#L136-L139

We need to get the include path from the location of the symbol being included. See how we do it for XML imports:

https://github.com/fprime-community/fpp/blob/a540390083cce051eacb06cbfd51bed1a3483e79/compiler/lib/src/main/scala/codegen/XmlWriter/XmlWriterState.scala#L65-L68

- [ ] Revise code
- [ ] Add unit test covering this case

tiffany1618:
Resolves #184


bocchino (approved):
Looks great!

PR #189, opened 2022-11-08
fixing topology includes

### LINKED ISSUE #188: Includes Inside Topology Node Does Not Work
The user guide, and unit-tests imply that one can use an include statement inside a topology file. For example: 

**instance-list.fppi:**
```fppi
instance mySuperInstance
```

**topology.fpp**
```fpp
module MyDeployment
{
    topology MyDeployment
    {
        include "instance-list.fppi"
    }
}
```

The expectation here is that the included `fppi` file will transform the topology such that it included `mySuperInstance`.  However, when running with this in-practice, the Ai.xml file generated is empty.

This is needed to support faux-subtopologies.

LeStarch:
This fixes #188.  The work performed was to:

1. Add missing `defTopologyAnnotatedNode` that uses the pre-existing and unused function `topologyMember`
2. Change the unit-test from including a constant (invalid syntax for including constants within typologies) to including an instance
3. Fix unit-test output.

### DIFF
--- a/compiler/lib/src/main/scala/transform/ResolveSpecInclude.scala
+++ b/compiler/lib/src/main/scala/transform/ResolveSpecInclude.scala
@@ -27,6 +27,21 @@ object ResolveSpecInclude extends AstStateTransformer {
     }
   }
 
+  override def defTopologyAnnotatedNode(
+    a: Analysis,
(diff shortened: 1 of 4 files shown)


bocchino (approved):
Looks great, thanks for this fix!

PR #190, opened 2022-11-08
Fix duplicate array constructors in generated C++

### LINKED ISSUE #186: Fix duplicate array constructors in generated C++
For an array of length 1, the single element and multiple element array constructors will have the same arguments. A check should be added in one of the constructor generators to avoid duplication in this case:

https://github.com/fprime-community/fpp/blob/c7c6e68f7205b3e6172bb42c4a36d0fffe88e545/compiler/lib/src/main/scala/codegen/CppWriter/ArrayCppWriter.scala#L216-L243

- [ ] Revise code
- [ ] Add unit test covering this case

tiffany1618:
Resolves #186


LeStarch (approved):
I decided to review this PR as I am trying to improve my understanding of Scala and the FPP code base.  To that end, I see:

1. The single element -> multiple set elements constructor is defined only in the case that the array has multiple elements.
2. That ctor (or None) is appended/concatenated to the list of constructors to generate. Meaning the list will only contain that constructor iff the array has more elements
3. Unit tests are updated
4. A single element array test was added.

This looks good to me!

compiler/lib/src/main/scala/codegen/CppWriter/ArrayCppWriter.scala
```diff
@@ -172,6 +172,26 @@ case class ArrayCppWriter (
 
   private def getConstructorMembers: List[CppDoc.Class.Member] = {
     val defaultValues = getDefaultValues
+    // Only write this constructor if the array has more than one element
+    val singleElementConstructor =
+      if arraySize == 1 then None
```
bocchino, line 177: Should this be Nil and List(...) instead of None and Some(...). If this works, it looks like Scala may be implicitly converting None to Nil and Some(...) to List(...).

compiler/lib/src/main/scala/codegen/CppWriter/ArrayCppWriter.scala
```diff
@@ -255,7 +261,8 @@ case class ArrayCppWriter (
           indexIterator(lines("this->elements[index] = obj.elements[index];")),
         )
       ),
-    )
+    ) ++
+      singleElementConstructor
```
bocchino, line 265: Can we put the single element constructor back where it was, before the multiple element constructor? It will cause less disruption in the unit test files. I think you can write List(...) ++ singleElementConstructor ++ List(...) instead of List(...) ++ singleElementConstructor.

bocchino (commented):
Looks good! Just a couple of minor comments.

compiler/lib/src/main/scala/codegen/CppWriter/ArrayCppWriter.scala
tiffany1618: I think that is what's happening, but Nil and List() does seem a little clearer

compiler/lib/src/main/scala/codegen/CppWriter/ArrayCppWriter.scala
tiffany1618: Ok, fixed!

bocchino (approved):
Looks great!

Issue #193, opened 2022-11-10
Integrate Port C++ with F Prime


bocchino:
Integrate the C++ code gen for ports with F Prime

- [x] Make sure it works ("smoke test")
- [x] Write functional unit tests in the F Prime repo
- [x] Merge into F Prime mainline

bocchino:
PR to F Prime mainline is https://github.com/nasa/fprime/pull/1852.

Issue #192, opened 2022-11-10
Test C++ ports


bocchino:
Test C++ code gen for ports.

bocchino:
Closed in #194.

PR #194, opened 2022-11-12
C++ port writer

### LINKED ISSUE #191: Implement C++ ports
Implement C++ code gen for ports.

tiffany1618:
Closes #191, #192

FPP changes:
- Add `PortCppWriter` for writing C++ for port definitions
- Add unit tests for C++ port writer
- Update unit tests for `fpp-depend` and `fpp-filenames`
- Add support for writing un-nested string classes in `StringCppWriter` 

Changes from Python port autocoder:
- Port buffer class only generated for ports without return types, since it is only used for port serialization
- Port string classes are wrapped the namespace `[port type name]PortStrings` to avoid polluting the global namespace

Temporary fixes for compatibility with Python component autocoder:
- Typedefs are generated to support string type names derived from the string parameter names, e.g. `typedef StringSize[size] [param name]String;`


bocchino:
Looks good! For this:

>Temporary fixes for compatibility with Python component autocoder:

>Generated string class names are derived from the name of the corresponding string parameter instead of the string size (<paramName>String instead of StringSize<size>)
Typedefs are generated for strings of the same size to avoid duplicating string classes

I had envisioned that we would keep generating string class names according to the string size (since that's what we eventually want to do) and add typedefs with the names of the string parameters to keep the component autocoder happy. Then later we can delete the typedefs. What do you think?

bocchino:
A related point is that I see changes to the struct and array code gen. I don't think this is necessary. I think we can just add the necessary typedefs to the port code gen.

tiffany1618:
> I had envisioned that we would keep generating string class names according to the string size (since that's what we eventually want to do) and add typedefs with the names of the string parameters to keep the component autocoder happy. Then later we can delete the typedefs. What do you think?

I agree, this makes more sense.

> A related point is that I see changes to the struct and array code gen. I don't think this is necessary. I think we can just add the necessary typedefs to the port code gen.

This was actually unrelated to the string class renaming. Arrays and structs both generate nested string classes, but ports do not, so the string class writer previously only generated nested classes. When I added generation for the un-nested string classes, I renamed the `write` function

bocchino:
> Arrays and structs both generate nested string classes, but ports do not

OK, thanks. I guess the string classes aren't nested in the port classes because there are two port classes. However, we need some kind of name qualification for the string classes. We can't put generic names like 'StringSize80' in the global namespace if we are generating one of these classes for each port type. Can we put the definitions in a namespace like `[port type name]Strings`?

tiffany1618:
>Can we put the definitions in a namespace like `[port type name]Strings`?

Yes, that sounds good.

bocchino:
OK, great! Maybe the namespace should be `[port type name]PortStrings`. This is consistent with the other code gen that adds `Port` to the port type name in the generated symbols.

tiffany1618:
Ok, done! The string typedefs are still in the global namespace, since the Python port autocoder was generating its string classes there.

compiler/lib/src/main/scala/codegen/CppWriter/PortCppWriter.scala
```diff
@@ -0,0 +1,624 @@
...
+  }
+
+  private def getHppIncludes: CppDoc.Member = {
+    val serializableHeader = data.returnType match {
+      case Some(_) => Nil
+      case None => List("Fw/Types/Serializable.hpp")
+    }
+    val standardHeaders = (
+      List(
+        "cstdio",
```
bocchino, line 166: cstdio and cstring should be system headers <...> not standard headers "...".

compiler/tools/fpp-to-cpp/test/port/ReturnTypePortAc.ref.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+    return this->m_func(this->m_comp, this->m_portNum, u);
+  }
+
+#if FW_PORT_SERIALIZATION == 1
+
+  Fw::SerializeStatus InputReturnTypePort ::
+    invokeSerial(Fw::SerializeBufferBase& _buffer)
+  {
+    // For ports with a return type, invokeSerial is not used
```
bocchino, line 63: clang is giving me an unused parameter warning when compiling this code. F Prime currently turns off these warnings, but I think we should make our generated code comply with them going forward. Let's add the following line to the function body to suppress the warning:
```
(void) _buffer;
```

bocchino (commented):
Looks great! Just a couple of comments.

Also, I think this PR does not close #193. So let's remove "closes #193" from the initial comment.

compiler/lib/src/main/scala/codegen/CppWriter/PortCppWriter.scala
bocchino: Compare how we handle system headers in the other Cpp writers.

tiffany1618: Fixed!

compiler/tools/fpp-to-cpp/test/port/ReturnTypePortAc.ref.cpp
tiffany1618: Fixed!

bocchino (approved):
Looks good!

PR #195, opened 2022-11-14
Minor changes to port cpp

bocchino:
I thought it would be easiest to do these minor changes in a separate PR.

* Consolidate duplicate code in string type generation
* ~Remove system headers -- I don't think they are used~
* Separate system headers from user headers

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/PortCppWriter.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/PortCppWriter.scala
@@ -160,7 +160,7 @@ case class PortCppWriter (
     val systemHeaders = List(
       "cstdio",
       "cstring",
-    ).map(CppWriter.systemHeaderString)
+    ).map(CppWriter.systemHeaderString).map(line)
     val serializableHeader = data.returnType match {
       case Some(_) => Nil
       case None => List("Fw/Types/Serializable.hpp")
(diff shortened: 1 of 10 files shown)


compiler/lib/src/main/scala/codegen/CppWriter/PortCppWriter.scala
```diff
@@ -157,10 +157,6 @@ case class PortCppWriter (
   }
 
   private def getHppIncludes: CppDoc.Member = {
-    val systemHeaders = List(
-      "cstdio",
-      "cstring",
```
tiffany1618, line 162: The component implementation classes in F Prime seem to be expecting both of these headers, which aren't included in other files. Some of them use `memset()` from `cstring` or `snprintf()` from `cstdio`, so I think these might be needed here.

tiffany1618 (commented):
Looks good! Just one comment

compiler/lib/src/main/scala/codegen/CppWriter/PortCppWriter.scala
bocchino: OK, good catch. I think those headers are not needed here, but you're right, the component code may depend on them. Let's fix this later when we revise the component code.

bocchino:
I put the system headers back.

tiffany1618 (approved):
Looks good!

PR #202, opened 2023-01-04
Add Emacs major mode

dennisog:
This adds a GNU Emacs major mode for editing `.fpp` source files. The major mode supports basic syntax highlighting and auto-indentation.


bocchino (approved):
Thanks, this looks good to me! I have not tested it.

PR #208, opened 2023-01-11
Revise semantic checking of serial port connections

### LINKED ISSUE #183: Enforce rules concerning return types with serial ports
When a serial port is connected to a typed input port `p`, the port type of `p` should not have a return type.

- [x] Update spec
- [x] Add check to semantics
- [x] Unit test
- [x] Update User's Guide

Add the corresponding rule for a typed output port `p` and a serial input port, when the port type of `p` has a return type.

- [x] Update spec
- [x] Add check to semantics
- [x] Unit test
- [x] Update User's Guide

bocchino:
* Exclude the cases where a serial port is connected to a typed port, and the typed port has a return type.
* Update the spec, implementation, unit tests, and user's guide.
* Make some incidental revisions to the user's guide.

Closes #183.

### DIFF
--- a/compiler/lib/src/main/scala/analysis/Semantics/Connection.scala
+++ b/compiler/lib/src/main/scala/analysis/Semantics/Connection.scala
@@ -31,6 +31,40 @@ case class Connection(
     }
   }
 
+  /** Checks the rule for a serial port connected to a typed
+   *  input port */
+  def checkSerialToTypedInput: Result.Result[Unit] = {
+    val fromInstance = from.port.portInstance
+    val fromType = fromInstance.getType
+    val toInstance = to.port.portInstance
+    val toType = toInstance.getType
+    (fromType, toType) match {
+      case (
+        Some(PortInstance.Type.Serial),
+        Some(PortInstance.Type.DefPort(Symbol.Port(aNode)))
+      ) => {
+        aNode._2.data.returnType match {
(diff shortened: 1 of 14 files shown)


tiffany1618:
Should we also be excluding the case where a typed output port with a return type is connected to a serial input port?

bocchino:
Yes, that sounds right. What happens in this case in the current implementation? Does the code even compile?

tiffany1618:
I think the expected behavior is to have an assertion fail on a nullptr, but currently it seg faults when `FW_PORT_SERIALIZATION` is defined. Here is the code that's generated:

https://github.com/fprime-community/fpp/blob/d55f67d48c8fe01994bd1abddd47974e98ab6667/compiler/tools/fpp-to-cpp/test/port/ReturnTypePortAc.ref.cpp#L110-L115

Maybe we should remove lines 110-112 for ports with return types.

bocchino:
> Maybe we should remove lines 110-112 for ports with return types.

Yes, that sounds right. Also, we should systematically replace `FW_ASSERT(ptr)` with `FW_ASSERT(ptr != nullptr)`. C++ does not guarantee that `nullptr == 0`.

bocchino:
OK, I revised the PR to exclude both serial -> typed and typed -> serial.

tiffany1618 (approved):
Looks good!

PR #212, opened 2023-01-19
Fix bugs in generated C++ for ports

### LINKED ISSUE #209: Fix bugs in generated C++ for ports
We need to fix the following bugs in the Python port autocoder:

1. The Python autocoder generates incorrect code in the `invoke` function of the `OutputPort` class. If port serialization is turned on, then there is an access to `m_port` guarded by a check that either `m_port` or `m_serPort` is non-null. In this case, if `m_serPort` is non-null and `m_port` is null, a seg fault occurs.

2. The Python autocoder assumes that `nullptr == 0`, which C++ does not guarantee.

See https://github.com/fprime-community/fpp/pull/208#issuecomment-1385933219.

tiffany1618:
Closes #209

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/PortCppWriter.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/PortCppWriter.scala
@@ -351,8 +351,8 @@ case class PortCppWriter (
                |this->trace();
                |#endif
                |
-               |FW_ASSERT(this->m_comp);
-               |FW_ASSERT(this->m_func);
+               |FW_ASSERT(this->m_comp != nullptr);
+               |FW_ASSERT(this->m_func != nullptr);
                |"""
           ) ++
           paramList.flatMap((n, tn, _) => {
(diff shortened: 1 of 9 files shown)


compiler/lib/src/main/scala/codegen/CppWriter/PortCppWriter.scala
```diff
@@ -501,38 +501,47 @@ case class PortCppWriter (
...
+      )
+      case None => List(
+        lines(
+          s"""|
+              |#if FW_PORT_SERIALIZATION
+              |FW_ASSERT((this->m_port != nullptr) || (this->m_serPort != nullptr));
+              |
+              |if (this->m_port != nullptr) {
+              |  $invokeCall
+              |} else {
```
bocchino, line 520: Can we move the else statement to the next line?

bocchino (commented):
Looks great! Just one comment.

compiler/lib/src/main/scala/codegen/CppWriter/PortCppWriter.scala
tiffany1618: Ok, fixed!

bocchino (approved):
Looks great!

PR #223, opened 2023-02-07
[CppDocCppWriter] Refine handling of pure virtual functions

bocchino:
* When the body of the function is Nil, don't write an implementation.
* Otherwise write out the implementation. This becomes a default implementation for the pure virtual function, which is allowed in C++.


tiffany1618 (commented):
Looks good! Should we add a test case for this too?

bocchino:
OK, good idea. I added some more code to the CppDocWriter test.

tiffany1618 (approved):
Looks good!

PR #225, opened 2023-02-15
Reorganize C++ writer tests

tiffany1618:
- Move fprime files out of `compiler/tools/fpp-to-cpp/test/top` and into `compiler/tools/fpp-to-cpp/test/fprime`
- Add `scala.` prefix to import statement for `scala.util.parsing.input.Position` in `ParserState.scala`


bocchino (commented):
Looks good, but let's remove the duplicate `Fw` directory inside `check-cpp-dir`.

bocchino (approved):
Looks good!

PR #227, opened 2023-02-16
Fix conditional compilation in generated port code

### LINKED ISSUE #226: Incorrect conditional compilation in generated port code
In this code:

https://github.com/fprime-community/fpp/blob/e951b43ee1e401b0f5828d63b5aff8f2274668d8/compiler/lib/src/main/scala/codegen/CppWriter/PortCppWriter.scala#L338-L340

the guard should say `#if FW_PORT_TRACING` not `#if FW_PORT_SERIALIZATION`. This issue causes compilation to fail if port serialization is on and port tracing is off.

@LeStarch FYI. If this issue is causing problems we can issue a point release to fix it. The current workaround is to turn on port tracing whenever port serialization is on.

tiffany1618:
Closes #226


bocchino (approved):
Looks good, thanks!

PR #230, opened 2023-02-22
Add CI workflow

tiffany1618:
- Add a Github actions workflow that runs all the sbt tests and FPP unit tests on PRs and pushes to main

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -22,12 +22,18 @@ jobs:
         java-version: '11'
         distribution: 'temurin'
         cache: 'sbt'
+    - name: Install Scala
+      run: |
+        curl -fL https://github.com/coursier/launchers/raw/master/cs-x86_64-pc-linux.gz | gzip -d > cs
+        chmod +x cs
+        echo Y | ./cs setup
     - name: Install fpp tools
       run: compiler/install
     - name: Run sbt tests
       run: sbt test
     - name: Run fpp tests
       run: |
+        export PATH=~/.local/share/coursier/bin:$PATH
         cd compiler/
         ./test
     - name: Clean up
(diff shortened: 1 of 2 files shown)


bocchino (approved):
Looks great! This is a nice process improvement.

bocchino:
@tiffany1618 The comment says "Test PR." Is this ready to merge? It looks good to me.

tiffany1618:
@bocchino I was trying to figure out why the tests in `lib/test/codegen/CppWriter/` weren't running. The `build` script was failing because it couldn't find `scalac`. Unfortunately the latest Ubuntu only has Scala 2.11, so I installed Scala through their Coursier tool to get a more recent version. I think it should work now.

bocchino:
This looks good to me! @LeStarch can you take a look?

LeStarch:
Looks good to me.  What is coursier?

tiffany1618:
@LeStarch Coursier is Scala's equivalent of a package manager. I used it since we're using a more recent version of Scala than the Ubuntu repositories provide.

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,42 @@
...
+        cache: 'sbt'
+    - name: Install Scala
+      run: |
+        curl -fL https://github.com/coursier/launchers/raw/master/cs-x86_64-pc-linux.gz | gzip -d > cs
+        chmod +x cs
+        echo Y | ./cs setup
+    - name: Install fpp tools
+      run: compiler/install
+    - name: Run sbt tests
+      run: sbt test
```
bocchino, line 33: Should we run `cd compiler` before `sbt test`? It looks like the run is warning that it can't find `project/sbt-properties`. If we run `sbt test` in `compiler`, that warning should disappear.

tiffany1618: Ok, I made that change.

bocchino:
Looks great!

PR #231, opened 2023-02-23
C++ writer refactor

tiffany1618:
- Write string types in C++ using names from `StringCppWriter` by default
- Add `stringTypeName` parameter to `TypeCppWriter` to optionally specify a string type name
- Add `FormalParamsCppWriter` for writing formal parameters as `CppDoc.Function.Param`s
- Add C++ writer utilities


bocchino (approved):
Looks great!

PR #237, opened 2023-03-30
Improve support for serializable values in text logging

### LINKED ISSUE #233: Improve support for serializable values in text logging
See issue https://github.com/nasa/fprime/issues/903. FPP allows arrays, structs, and enums as event arguments, but if FW_TEXT_LOG is on and FW_SERIALIZABLE_TO_STRING is off, the code that generates text logging for these values won't compile. Let's support this case by generating a lightweight `toString` function. For enums, this function can return a short string with just the numerical value of the enum represented as digits. For arrays and structs, we can use the placeholder value `NOSPEC`. We just need to revise this framework code 

https://github.com/nasa/fprime/blob/35625ebb72efaa292cfda3c06b28fd6a13163f16/Fw/Types/Serializable.cpp#L22-L28

to handle the case FW_TEXT_LOG as well.

For enums, we should also replace the `%d` with the correct format specifier here:

https://github.com/fprime-community/fpp/blob/10cbd1610ddfab23bef9ba64938222f74288c031/compiler/tools/fpp-to-cpp/test/enum/ExplicitEnumAc.ref.cpp#L94

- [x] Revise the %d format specifier
- [x] Add the lightweight enum string in the case that TEXT_LOG is on but serializable strings are off.
- [ ] Make a PR to F Prime to have the Serializable base class return "NOSPEC" when TEXT_LOG is on and serializable strings are off.

tiffany1618:
Closes #233.

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/EnumCppWriter.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/EnumCppWriter.scala
@@ -351,47 +351,74 @@ case class EnumCppWriter(
         )
       )
     ) ++
-      wrapClassMembersInIfDirective(
-        "\n#if FW_SERIALIZABLE_TO_STRING || BUILD_UT",
+      List.concat(
         List(
-          functionClassMember(
-            Some(s"Convert enum to string"),
-            "toString",
-            List(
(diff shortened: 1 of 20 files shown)


bocchino (approved):
Looks great!

Issue #216, opened 2023-01-25
Test C++ component base classes


bocchino:
Test C++ code gen for component base classes.

tiffany1618:
- [ ] Hpp files compile with cpp files generated by Python autocoder
- [ ] All files compile with F Prime
- [ ] All files pass F Prime unit tests
- [ ] Write comprehensive functional tests for components in F Prime
- [ ] All files pass functional tests for components

bocchino:
Closed as completed in #240.

PR #240, opened 2023-03-31
C++ component writer

### LINKED ISSUE #215: Implement C++ component base classes
Implement C++ code gen for component base classes.

tiffany1618:
Closes #215, #216.

FPP changes:
- Add `ComponentCppWriter` and accompanying helper writers for writing C++ component definitions
- Add support for writing event format strings in `FormatCppWriter`
- Refactor `PortCppWriter` to use `FormalParamsCppWriter` and to provide functions to compute names found within port classes
- Add unit tests for C++ component writer
    - Includes support for compilation tests with different combinations of values of F Prime guards (see `compiler/tools/fpp-to-cpp/test/component/check-cpp`)
- Update unit tests for `fpp-depend` and `fpp-filenames`

Changes from Python Autocoder:
- Fix issue where the implementations for callback functions for command ports are not generated when a component contains command ports but no commands
- Fix return type for handler, handler base, and callback functions for ports that return a string type to be the string type defined in the corresponding generated port file instead of `string`
- Add static assertion for components containing serial ports to make sure `FW_PORT_SERIALIZATION` is on
- Reorganization of class members
    - All constants appear before function declarations in the header file
    - All member variables appear after function declarations in the header file
    - Member functions are first organized by access specifier in the following order:
        - `public`
        - `PROTECTED`
        - `PRIVATE`
    - Within each access specifier, member functions are organized by category in the following order:
        - Component functions (constructor, destructor, initialization, etc.)
        - Input/output port functions
        - Internal port functions
        - Command functions
        - Time get function
        - Event functions
        - Telemetry functions
        - Parameter functions
- Add an assertion to check the port number in handler base functions for serial ports


compiler/lib/src/main/scala/codegen/CppWriter/ComputeCppFiles.scala
```diff
@@ -44,7 +44,13 @@ object ComputeCppFiles extends AstStateVisitor {
   ) = {
     val node = aNode._2
     val data = node.data
-    visitList(s, data.members, matchComponentMember)
+    val name = s.getName(Symbol.Component(aNode))
+    val loc = Locations.get(node.id)
+    for {
+      s1 <- addMappings(s, FileNames.getComponent(name), Some(loc))
```
bocchino, line 50: This looks good. Or you can replace s1 and s2 with s. If you reuse s in a for...yield construct, it creates a new s that hides the old s. This is a useful pattern for "assigning" to s. It avoids using the wrong version of s1, s2, etc.

compiler/lib/src/main/scala/codegen/CppWriter/CppWriter.scala
```diff
@@ -25,7 +25,12 @@ object CppWriter extends AstStateVisitor with LineUtils {
   ) = {
     val node = aNode._2
     val data = node.data
-    visitList(s, data.members, matchComponentMember)
+    val cppDoc = ComponentCppWriter(s, aNode).write
+    for {
+      _ <- writeCppDoc(s, cppDoc)
```
bocchino, line 30: I think the LHS should be s in line 30 (we should thread the state through the computation). Even if the state isn't changed in line 30, this is good defensive programming.

compiler/lib/src/main/scala/codegen/CppWriter/FormatCppWriter.scala
```diff
@@ -56,11 +56,23 @@ object FormatCppWriter {
...
   def write(f: Format, tn: AstNode[Ast.TypeName]): String = {
-    def escapePercent(s: String) = s.replaceAll("%", "%%")
     f.fields.foldLeft(escapePercent(f.prefix))((a, s) =>
       a + (s match {
         case (field, suffix) => writeField(field, tn) + escapePercent(suffix)
       })
     )
   }
+
+  def writeEventFormat(event: Event): String = {
```
bocchino, line 66: Does this function belong here? It seems event- and component-specific. Maybe the FormatCppWriter should just write formats.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriter.scala
```diff
@@ -0,0 +1,973 @@
...
+                      )
+                    )
+                  }
+                )
+              )
+            )
+          )
+        )
+      )
+    )
```
bocchino, line 515: This is a lot of close parentheses in a row. Consider breaking this up by assigning some of the inner expressions to variables.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentEvents.scala
```diff
@@ -0,0 +1,353 @@
...
+                    )
+                  )
+                ),
+                lines("#endif")
+              )
+            )
+          )
+        )
+      )
+    )
```
bocchino, line 304: This is a lot of close parentheses in a row.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentTelemetry.scala
```diff
@@ -0,0 +1,187 @@
...
+                        |);
+                        |"""
+                  )
+                )
+              )
+            )
+          )
+        )
+      )
+    )
```
bocchino, line 169: A lot of close parentheses in a row.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCommands.scala
```diff
@@ -0,0 +1,338 @@
...
+            List(
+              lines(s"FW_ASSERT(this->${portVariableName(cmdRegPort.get)}[0].isConnected());"),
+              intersperseBlankLines(
+                sortedCmds.map((_, cmd) =>
+                  lines(
+                    s"""|this->${portVariableName(cmdRegPort.get)}[0].invoke(
+                        |  this->getIdBase() + ${commandConstantName(cmd)}
+                        |);
+                        |"""
+                  )
```
bocchino, line 110: This is a lot of close parentheses in a row. Consider breaking it up.

compiler/tools/fpp-to-cpp/test/component/test/check-cpp
```diff
@@ -0,0 +1,40 @@
...
+for file in `find . -name '*Ac.cpp'`
+do
+  echo "compiling $file with default guards"
+  $fprime_gcc -I../../.. -c $file -Wno-gnu-zero-variadic-macro-arguments -Wno-unused-parameter
+
+  g=`echo $guards_on`
+  echo "compiling $file with $g"
+  $fprime_gcc -I../../.. -c $file $g -Wno-gnu-zero-variadic-macro-arguments -Wno-unused-parameter
+
+#  g=`echo $guards_off`
```
bocchino, line 33: Why is this code commented out? Can we delete it?

bocchino (commented):
Looks good! Just a few comments.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriterUtils.scala
```diff
@@ -0,0 +1,612 @@
...
+      s,
+      s.a.typeMap(param.typeName.id),
+      Some("Fw::LogStringArg")
+    )
+
+  /** Write a channel type as a C++ type */
+  def getChannelType(t: Type): String =
+    TypeCppWriter.getName(s, t, Some("Fw::TlmString"))
+
+  /** Add an optional string separated by two newlines */
```
bocchino, line 404: Can we move this function into CppWriterUtils? It seems generic.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriterUtils.scala
```diff
@@ -0,0 +1,612 @@
...
+  def addSeparatedString(str: String, strOpt: Option[String]): String = {
+    strOpt match {
+      case Some(s) => s"$str\n\n$s"
+      case None => str
+    }
+  }
+
+  /** Add an optional pre comment separated by two newlines */
+  def addSeparatedPreComment(str: String, commentOpt: Option[String]): String = {
```
bocchino, line 413: This function seems generic too.

compiler/lib/src/main/scala/codegen/CppWriter/ComputeCppFiles.scala
tiffany1618: Ok, I changed this. I think that's better.

compiler/lib/src/main/scala/codegen/CppWriter/FormatCppWriter.scala
tiffany1618: Good point -- I moved it to ComponentCppWriterUtils.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriter.scala
tiffany1618: Ok, fixed!

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriterUtils.scala
tiffany1618: This function is channel-specific, so I think it belongs here.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriterUtils.scala
tiffany1618: Ok, I moved this function to CppWriterUtils.

compiler/tools/fpp-to-cpp/test/component/test/check-cpp
tiffany1618: This is an old version of the script from an old test directory that I created to compile files generated by the Python Autocoder, and should have been deleted. The current version of this script is here: [compiler/tools/fpp-to-cpp/test/component/check-cpp ](https://github.com/fprime-community/fpp/pull/240/files#diff-f663fb671471e8802cc40fa486edfb4761642c0ec5ce509644c427863fae9cd6)

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriter.scala
bocchino: Looks great!

bocchino (approved):
Great work! This PR is a big accomplishment.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriterUtils.scala
bocchino: I agree.

PR #247, opened 2023-04-10
Improve support for manual topology setup

### LINKED ISSUE #239: Improve support for manual topology setup
The topology code generator currently optimizes away functions that are not called by the auto-generated `setup` function -- e.g., no `loadParameters` function is generated if there are no parameters and `setup` doesn't call `loadParameters`. This causes issues when users want to write a manual `setup` function, because the setup function has to change if the functions appear and disappear. The fix is to generate empty functions and call them.

bocchino:
Closes #239.

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/TopologyCppWriter/TopHelperFns.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/TopologyCppWriter/TopHelperFns.scala
@@ -289,9 +289,15 @@ case class TopHelperFns(
     name: String,
     params: List[CppDoc.Function.Param],
     body: List[Line]
-  ) = body match {
-    case Nil => None
-    case ll => Some(
+  ): Option[CppDoc.Member.Function] = {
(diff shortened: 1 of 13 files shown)


bocchino:
@LeStarch When you get a chance, can you look at this PR? It implements a feature that you requested.

LeStarch (approved):
@bocchino I read this change as the following:

Originally, we matched the body (lines of the function), if empty/nil then nothing was produced, otherwise the `Some` thing was produced (presumably with the function implementation per `ll`).

Now we are *always* setting `ll` to the result of the body match.  Nil produces a single comment line, and anything else produces the original lines. `Some` always runs but may just be passed a comment.

The effect is now a function is always produced, sometimes just a comment, otherwise the implementation. 

If this analysis is correct, then I approve this change!

LeStarch:
@bocchino I wrote-up my understanding in the approval. If that is the gist of what is happening, then this is approved!

bocchino:
Yes, that is correct. Thanks!

PR #253, opened 2023-05-14
Fix generation of struct containing only arrays

SMorettini:
### Problem

The generation of the equal operator of `.cpp` for the following fpp
```
struct Modules4 {
  arr1: [3] M.Modules2
  arr2: [6] M.Modules2
}
```
is 
```
bool Modules4 ::
  operator==(const Modules4& obj) const
{
  // Compare non-array members
  if (!(

  )) {
    return false;
  }

  // Compare array members
...
```

The first if is empty and doesn't compile.

# Solution

I change the fpp-to-cpp to generate correct code also if there are no non-array members. 

### Note: it's my first time working with Scala, let me know if I should make the change using a different approach


compiler/lib/src/main/scala/codegen/CppWriter/StructCppWriter.scala
```diff
@@ -261,22 +261,25 @@ case class StructCppWriter(
...
-          lines("if (!("),
-          nonArrayMemberCheck.map(indentIn),
-          lines(
-            """|)) {
-               |  return false;
-               |}"""
-          ),
-        ).flatten,
-        Line.blank :: lines("// Compare array members"),
+        if nonArrayMemberNames.length >0 then List(
```
bocchino, line 264: Please add a space between > and 0.

bocchino (changes requested):
Good catch! Looks good. Just a couple of formatting issues.

SMorettini:
I apply the formatting changes suggested

bocchino (approved):
Looks good!

PR #257, opened 2023-05-31
odroid runner for building aarch64 packages, rhel8.7 runner to replace ubuntu-18.04

jwest115:
Updated workflow for building native images to include:
- "odroid" label for building on a self-hosted runner
- condition for determining the architecture type (used when installing GraalVM in env-setup)
- condition for publishing binaries and building PyPI packages ONLY when triggered by a release 
- "rhel-8.7" label for building on self-hosted runner (replaces ubuntu-18.04)
- scope PR trigger to those merging to the "main" branch

### DIFF
--- a/.github/workflows/build-native.yml
+++ b/.github/workflows/build-native.yml
@@ -17,11 +17,9 @@ jobs:
     - name: Image Build Step
       working-directory: ./compiler
       if: ${{ matrix.os == 'odroid' }}
-        run: |
-          echo "architecture=aarch64" >> "$GITHUB_ENV"
+        run: echo "architecture=aarch64" >> "$GITHUB_ENV"
       if: ${{ matrix.os != 'odroid' }}
-        run: |
-          echo "architecture=amd64" >> "$GITHUB_ENV"
+        run: echo "architecture=amd64" >> "$GITHUB_ENV"
       run: |
         . ${GITHUB_WORKSPACE}/.github/workflows/env-setup ${{ env.architecture }}
         ./release
--- a/.github/workflows/build-native.yml
+++ b/.github/workflows/build-native.yml
@@ -14,12 +14,14 @@ jobs:
       uses: actions/checkout@v2
       with:
         fetch-depth: 0
+    - name: Set architecture type (aarch64)
+      if: ${{ matrix.os == 'odroid' }}
+      run: echo "architecture=aarch64" >> "$GITHUB_ENV"
+    - name: Set architecture type (amd64)
+      if: ${{ matrix.os != 'odroid' }}
+      run: echo "architecture=amd64" >> "$GITHUB_ENV"
     - name: Image Build Step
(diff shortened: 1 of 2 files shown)


.github/workflows/build-native.yml
```diff
@@ -2,28 +2,37 @@ name: Build Native Images
 on:
   release:
     types: [published]
+  pull_request:
```
LeStarch, line 5: Do we need this empty line any more?

.github/workflows/build-native.yml
```diff
@@ -2,28 +2,37 @@ name: Build Native Images
...
     - name: Image Build Step
       working-directory: ./compiler
       run: |
-        . ${GITHUB_WORKSPACE}/.github/workflows/env-setup
+        . ${GITHUB_WORKSPACE}/.github/workflows/env-setup ${{ env.architecture }}
         ./release
-        mv native-fpp-*.tar.gz "${GITHUB_WORKSPACE}"
+        mv native-fpp-*.tar.gz "${GITHUB_WORKSPACE}" 
     - name: Publish Release Binaries
+      if: ${{ github.event_name == 'release' }}
```
LeStarch, line 30: Since this is guarded, should we run the build on ever PR anyway? That way we'd know the native compile does not break.

LeStarch (changes requested):
We should choose one of the following:
1. Remove the on/pull_request empty flag
2. Set on/pull_request: "trigger" to test the build on each PR without publishing it.

Thoughts?

LeStarch:
Otherwise this is **awesome**!!!

.github/workflows/build-native.yml
jwest115: I think from previous discussion we talked about only running the section for publishing the release if the build was triggered via the "release" trigger. Are you referring to running this regardless of trigger type (so for PRs and releases)?

bocchino: > Since this is guarded, should we run the build on ever PR anyway? That way we'd know the native compile does not break.

I'd rather not run the native build on every PR during development. However, it would be nice to be able to test the native build near release time without creating a test release. Is there a middle ground here?

jwest115: Maybe we can include the release "pre-release" or "unpublished" triggers so that the build will run once a release is created (but not yet in the published state)?

LeStarch: The native builds are a required feature of this code, and thus it seems that building natively should be a required test.  Is there a downside of this that I am not seeing?

.github/workflows/build-native.yml
jwest115: Scoped the pull_request trigger to PRs targeting the "main" branch

LeStarch: ok!

LeStarch (approved):

PR #258, opened 2023-06-05
Remove unused serial input port to typed output port connector function

### LINKED ISSUE #256: Remove unused serial input port to typed output port connector function
Serial ports should never be connected to typed ports that have a return type. Therefore, serial input port to typed output port connector functions should not be generated when the typed output port has a return type.

For example, this function should not be generated:

https://github.com/fprime-community/fpp/blob/71932e3f0a58fa60046273776e0b2f401e005505/compiler/tools/fpp-to-cpp/test/component/ActiveTestComponentAc.ref.cpp#L1171-L1183

tiffany1618:
Closes #256.

Also fixes the comment formatting in `addSeparatedPrecomment()`  in `CppWriterUtils`.


bocchino (approved):
Looks great! This is a nice optimization to reduce dead code in the generated C++.

PR #262, opened 2023-06-15
Parser Bug Fix

elliotfayman:
Resolving a bug that more accurately identifies sources of errors within arithmetic operations.


bocchino:
This issue has to do with parsing and location info for left-associative expressions. The Scala parser combinator can't express left-recursive parse rules, so we need to parse a left-associative expression like `1 + 2 + 3` as follows:
1. Parse an initial expression and a list of operations, e.g., `1, + 2, + 3`.
2. Use the expression and list to reconstruct the left-recursive parse tree, e.g., `(1 + 2) + 3`.

In step 2 we need to interpolate the location info for the new AST nodes.  The existing code uses the location of 1 for the locations of + 2 and + 3, which is not as precise as it could be. The updated code uses the locations of the tokens + to resolve the locations of + 2 and + 3. This is better.

bocchino (approved):
Looks good!

PR #265, opened 2023-06-16
Revise check-xml

bocchino:
In the checker for the reference XML in the unit tests, disable Python autocoding and C++ compilation in new cases, due to regressions in F Prime. Specifically, this commit:

https://github.com/nasa/fprime/pull/1734

causes new failures in cases where there is an XML enum outside of a namespace.

This probably isn't worth fixing in F Prime, since we have phased out this XML-to-C++ code gen. However, the fprime-check script should work.

@LeStarch FYI.


bocchino:
I added FPP modules in a few cases to make the XML code generation work.

tiffany1618 (approved):
Looks good!

PR #267, opened 2023-06-21
Remove private parameter getter in C++ component writer

tiffany1618:
Remove the `getParam` function in component base classes. This is a private function that gets parameters by ID, but it is private and not used elsewhere in the component base class. It is also not used in any of the F Prime framework classes, and the F Prime framework successfully builds and passes all unit tests after the removal of this function.


bocchino (approved):
Looks good!

PR #268, opened 2023-06-27
C++ writer for component implementation templates

### LINKED ISSUE #218: Implement C++ component implementation templates
Implement C++ code gen for component implementation templates.

tiffany1618:
Closes #218

- Add `ComponentImplWriter` for writing C++ component implementation templates
- Add `template` parameter to `CppWriterState` to indicate whether to generate component base or implementation classes
- Add support for generating C++ files with arbitrary file extensions
- Reorganize unit tests for C++ component writer and add unit tests for implementation templates

Changes from Python autocoder:
- Add handler implementations for internal interfaces
- Change file extension of template files from `.{hpp,cpp}-template` to `.template.{hpp,cpp}`
- Fix formatting of Doxygen comments


compiler/lib/src/main/scala/codegen/CppWriter/CppWriterState.scala
```diff
@@ -16,6 +16,8 @@ case class CppWriterState(
   pathPrefixes: List[String] = Nil,
   /** The default string size */
   defaultStringSize: Int = CppWriterState.defaultDefaultStringSize,
+  /** Whether or not to emit template code */
+  template: Boolean = false,
```
bocchino, line 20: Let's make this a variant type, not a Boolean. We should have one variant for each mode. I think there are two modes:
1. AUTOCODE: Generate F Prime autocode
2. IMPL_TEMPLATE: Generate a template for the hand-coded implementation

For the variant type, in Scala 3 we can use an enum.

compiler/tools/fpp-to-cpp/src/main/scala/fpp-to-cpp.scala
```diff
@@ -62,6 +62,7 @@ object FPPToCpp {
           options.guardPrefix,
           options.pathPrefixes,
           options.defaultStringSize,
+          options.template,
```
bocchino, line 65: It's OK to use a boolean flag for the command-line option, but let's convert this to a mode variable here. See above.

compiler/lib/src/main/scala/codegen/CppWriter/ComputeCppFiles.scala
```diff
@@ -47,7 +50,8 @@ object ComputeCppFiles extends AstStateVisitor {
     val name = s.getName(Symbol.Component(aNode))
     val loc = Locations.get(node.id)
     for {
-      s <- addMappings(s, FileNames.getComponent(name), Some(loc))
+      s <- if s.template then addMappings(s, FileNames.getComponentImpl(name), Some(loc))
```
bocchino, line 53: I think we should have two different visitors here, one for each mode (autocode generation, template generation). Also I think CppWriter should have two different visitors. In autocode gen mode, we should have the same behavior as before. In template gen mode, we should generate only component implementation templates and nothing else (e.g. no types, ports, etc.). We should add tests to the `fpp-filenames` tool to make sure that ComputeCppFiles is working as intended.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentImplWriter.scala
```diff
@@ -0,0 +1,161 @@
...
+
+  def write: CppDoc = {
+    val includeGuard = s.includeGuardFromQualifiedName(symbol, fileName)
+    CppWriter.createCppDoc(
+      s"$name component implementation class",
+      fileName,
+      includeGuard,
+      getMembers,
+      s.toolName,
+      "hpp-template",
```
bocchino, line 28: This is what F Prime currently uses for the template suffix, but I think it would be better to use .template.{hpp,cpp}. That way when you open the file you get the syntax highlighting. Sometimes it’s useful to open the template file to see what’s in it, without first copying it over to a file with a correct suffix. @LeStarch does this sound OK?

bocchino (commented):
Looks great! I think we're now very close to doing an FPP release.

I made a few comments. Please take a look at them. We can discuss them tomorrow at the meeting.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentImplWriter.scala
bocchino: Also I notice that even in this PR, GitHub doesn't highlight the `.hpp-template` reference files as C++. So the files are harder to read than the could be.

tiffany1618: Yeah, I noticed that too. I think this is a good idea.

compiler/lib/src/main/scala/codegen/CppWriter/CppWriterState.scala
bocchino: Actually if we have two separate visitors for the two modes, we may not need a mode variable, right? We could just call the appropriate visitor from the top-level code.

bocchino: Maybe we could make CppWriter an abstract class and have two subclasses, AutocodeCppWriter and ImplCppWriter.

tiffany1618: I think we should have the two separate visitors, but leave CppWriter as is (minus the visitor functions), since it has some functions that are used by the other cpp writer classes.

bocchino: > it has some functions that are used by the other cpp writer classes

OK, could we move those functions to CppWriterUtils?

Some helper functions are private and I think those should be made protected members of CppWriter so that AutocodeCppWriter and ImplCppWriter can inherit from CppWriter and use them.

compiler/tools/fpp-to-cpp/src/main/scala/fpp-to-cpp.scala
bocchino: With two visitors I don't think we need a mode variable.

compiler/lib/src/main/scala/codegen/CppWriter/CppWriterState.scala
tiffany1618: Ok, this sounds good. I also realized those functions can just stay in a companion object of CppWriter.

compiler/lib/src/main/scala/codegen/ComputeGeneratedFiles.scala
```diff
@@ -17,10 +17,10 @@ object ComputeGeneratedFiles {
...
     for {
-      s <- ComputeCppFiles.visitList(
+      s <- ComputeAutocodeCppFiles.visitList(
         CppWriterState(a),
         tul,
-        ComputeCppFiles.transUnit
+        ComputeAutocodeCppFiles.transUnit
       )
     }
     yield s.locationMap.toList.map(_._1)
```
tiffany1618, line 26: This still needs some work -- I'm not sure what's the best way to indicate which filenames to compute. Should we add flags to the fpp-depend and fpp-filenames tools?

bocchino: Yes, good idea! I added draft specs for the `-t` options to the wiki. Please take a look and see what you think.

https://github.com/fprime-community/fpp/wiki/fpp-depend
https://github.com/fprime-community/fpp/wiki/fpp-filenames

tiffany1618: This looks good, thanks!

compiler/lib/src/main/scala/codegen/CppWriter/CppWriterState.scala
```diff
@@ -16,7 +16,7 @@ case class CppWriterState(
   pathPrefixes: List[String] = Nil,
   /** The default string size */
   defaultStringSize: Int = CppWriterState.defaultDefaultStringSize,
-  /** The name of the tool using the CppWriter */
+  /** Whether or not to emit template code */
```
bocchino, line 19: Looks like the wrong comment was deleted here.

compiler/tools/fpp-to-cpp/src/main/scala/fpp-to-cpp.scala
```diff
@@ -40,11 +40,16 @@ object FPPToCpp {
         Parser.parseFile (Parser.transUnit) (None) _
       )
       a <- CheckSemantics.tuList(a, tulFiles ++ tulImports)
-      s <- ComputeCppFiles.visitList(
-        CppWriterState(a),
-        tulFiles,
-        ComputeCppFiles.transUnit
-      )
+      s <- if options.template then ComputeImplCppFiles.visitList(
```
bocchino, line 43: Let’s match on a mode variable here. We can add code to CppWriter that converts the boolean flags to the mode variable. Then the tools that need this code can use it.

bocchino: This will also let us extend the code when we add unit test code gen.

compiler/tools/fpp-to-cpp/test/component/impl/ActiveCommands.ref.template.cpp
```diff
@@ -0,0 +1,380 @@
...
+// ----------------------------------------------------------------------
+
+ActiveCommands ::
+  ActiveCommands(const char* const compName) :
+    ActiveCommandsComponentBase(compName)
+{
+
+}
+
+void ActiveCommands ::
```
bocchino, line 21: @LeStarch Did you want us to remove this init function from the code gen? I seem to recall a change in F Prime along these lines.

LeStarch: Yes please!  The we now recommend that projects use the existing `init` function as it is all boilerplate.  Thus we've been removing them from new components by hand.

bocchino (commented):
Looks great! Let's just remove the init function in the component template, per the comment from Michael.

compiler/tools/fpp-to-cpp/test/component/impl/ActiveCommands.ref.template.cpp
bocchino: Thanks! This is done.

bocchino (approved):
Looks great!

PR #271, opened 2023-06-29
Remove redundant serial connector function

tiffany1618:
This serial connector function:

https://github.com/fprime-community/fpp/blob/22dc4207987e0f30ac8c75a6c91fd9fd14ccb3e7/compiler/tools/fpp-to-cpp/test/component/ActiveSerialComponentAc.ref.cpp#L1289-L1301

is a less general version of this serial connector function:

https://github.com/fprime-community/fpp/blob/22dc4207987e0f30ac8c75a6c91fd9fd14ccb3e7/compiler/tools/fpp-to-cpp/test/component/ActiveSerialComponentAc.ref.cpp#L1441-L1453

so it is redundant to have both versions. The less general version can be safely removed.


bocchino (commented):
This looks good to me! Tim, Michael, can you take a look?

bocchino (approved):
Looks good! Approved after discussion with Tim, Michael, and Thomas.

PR #272, opened 2023-07-05
Install FPL as part of FPP pip install

thomas-bc:
## Context

Addresses https://github.com/nasa/fprime/issues/2084

In continuation of the `fprime-util visualize` effort, which aims to seamlessly visualize an FPP model with a single command. We need an easy way to get the FPL tools (https://github.com/fprime-community/fprime-layout) installed on the user's machine.\
It was proposed that these should be installed in the FPP pip installation process. This is what this PR is introducing.

## Depends on
- https://github.com/fprime-community/fprime-layout/pull/1
(_My first commit was `cd`'ing into the project to install it, but that was not the cleanest solution. Might as well keep the FPL install process in line with FPP's._)

## Testing recommendation
Wait for the above PR to be merged in, or replace the `/fprime-community/` org with `/thomas-bc/` [in the repo URL here](https://github.com/thomas-bc/fpp/blob/41658c0f735d75d46cab3ae2779a8151711136cc/fprime_fpp/fprime_fpp_install.py#L269), as I have introduced the change in my fork.

### DIFF
--- a/fprime_fpp/fprime_fpp_install.py
+++ b/fprime_fpp/fprime_fpp_install.py
@@ -265,24 +265,18 @@ def install_fpl_via_git(installation_directory: Path, subprocess_environment: di
             file=sys.stderr,
         )
         sys.exit(-1)
-    # Retrieve fprime-layout source
-    step = ["git", "clone", "https://github.com/fprime-community/fprime-layout"]
-    print(f"-- INFO  -- Running { ' '.join(step) }")
-    completed = subprocess.run(step, env=subprocess_environment)
-    if completed.returncode != 0:
-        print(
-            f"-- ERROR -- Failed to run { ' '.join(step) }", file=sys.stderr
-        )
-        sys.exit(-1)
-    # Install fprime-layout
-    step = ["./install", str(installation_directory)]
-    print(f"-- INFO  -- Running { ' '.join(step) }")
-    completed = subprocess.run(step, env=subprocess_environment, cwd="fprime-layout")
-    if completed.returncode != 0:
-        print(
-            f"-- ERROR -- Failed to run { ' '.join(step) }", file=sys.stderr
-        )
-        sys.exit(-1)
+    steps = [
+        ["git", "clone", "https://github.com/fprime-community/fprime-layout"],
+        ["fprime-layout/install", str(installation_directory)]
+    ]
+    for step in steps:
+        print(f"-- INFO  -- Running { ' '.join(step) }")
+        completed = subprocess.run(step, env=subprocess_environment)
+        if completed.returncode != 0:
+            print(
+                f"-- ERROR -- Failed to run { ' '.join(step) }", file=sys.stderr
+            )
+            sys.exit(-1)
 
 def iterate_fpp_tools(working_dir: Path) -> Iterable[Path]:
     """Iterates through FPP tools"""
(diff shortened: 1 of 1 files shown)


fprime_fpp/fprime_fpp_install.py
```diff
@@ -244,15 +245,38 @@ def safe_extract(tar, path=".", members=None, *, numeric_owner=False):
...
         for step in steps:
             print(f"-- INFO  -- Running { ' '.join(step) }")
             completed = subprocess.run(step, env=subprocess_environment)
             if completed.returncode != 0:
                 print(f"-- ERROR -- Failed to run { ' '.join(step) }", file=sys.stderr)
                 sys.exit(-1)
-
+        # FPL install (https://github.com/fprime-community/fprime-layout)
+        if (os.getenv("FPL_NO_INSTALL") != "TRUE"):
+            install_fpl_via_git(installation_directory, subprocess_environment)
```
thomas-bc, line 257: FPL install has to be ran within the tools_directory context manager
https://github.com/fprime-community/fpp/blob/22dc4207987e0f30ac8c75a6c91fd9fd14ccb3e7/fprime_fpp/fprime_fpp_install.py#L214
since SBT is downloaded in this ephemeral directory.
So I am running the FPL install here, within the FPP install. This makes it a simpler PR, however another easy (and maybe cleaner?) solution would be to move the context manager up at this level: 
https://github.com/fprime-community/fpp/blob/22dc4207987e0f30ac8c75a6c91fd9fd14ccb3e7/fprime_fpp/fprime_fpp_install.py#L186
and run those "independently" (instead of nesting FPL within FPP) like such:
```
except urllib.error.HTTPError:
  with tempfile.TemporaryDirectory() as tools_directory:
    install_fpp_via_git(working_dir, tools_directory)
    install_fpl_via_git(working_dir, tools_directory)
```

I went with the PR that introduces the less changes. But let me know what you think, and I can quickly make the change if needed.

bocchino: This makes sense. I am in favor of the cleaner solution. I'm not too concerned about minimizing changes.

fprime_fpp/fprime_fpp_install.py
```diff
@@ -199,6 +199,7 @@ def install_fpp_via_git(installation_directory: Path):
     Should FPP not be available as a published version, this will clone the FPP repo, checkout, and build the FPP tools
     for the given version. This requires the following tools to exist on the system: git, sh, java, and sbt. These tools
     will be checked and then the process will run and intall into the specified directory.
+    Additionally installs the FPL (https://github.com/fprime-community/fprime-layout) tools unless FPL_NO_INSTALL is set to TRUE.
```
bocchino, line 202: Can we change the flag name from FPP_NO_INSTALL to FPP_INSTALL? In general, double-negative flags of the form X_IS_NOT_TRUE = TRUE are confusing and should be avoided. You could do the installation unless FPP_INSTALL = FALSE.

fprime_fpp/fprime_fpp_install.py
```diff
@@ -212,48 +213,90 @@ def install_fpp_via_git(installation_directory: Path):
...
-            
-                for member in tar.getmembers():
-                    member_path = os.path.join(path, member.name)
-                    if not is_within_directory(path, member_path):
-                        raise Exception("Attempted Path Traversal in Tar File")
+        with safe_chdir(tools_directory):
+            # Install SBT into ephemeral tools directory
+            sbt_env = install_sbt(tools_directory)
+            # Install FPP tools in installation_directory
+            install_fpp_via_git(installation_directory, tools_directory, sbt_env)
```
LeStarch, line 220: This particular call does not use git anymore....should we change the name?

LeStarch (commented):
Just one comment.

fprime_fpp/fprime_fpp_install.py
thomas-bc: Addressed

PR #273, opened 2023-07-07
Made error message clickable

SMorettini:
## Change

This pull request I changed the path of the file printed in the  error message from 
```
/my_path/folder/folder/my_file.fpp: 61.66
```
to
```
/my_path/folder/folder/my_file.fpp:61.66
```

## Motivation

The new chance make the error message clickable from vs code and maybe for other editor too. 

Moreover it follows the changes done in Fprime for unit tests and asserts:

* https://github.com/nasa/fprime/pull/2022 -> assert pull request
* https://github.com/nasa/fprime/pull/1829 -> unit tests pull request

## Note

If the Fpp community agree with this change I will update the tests accordingly because this changes broke a lot of tests as for example:
```
--- missing_ports.ref.txt	2023-06-06 13:21:23.924335286 +0000
+++ missing_ports.out.txt	2023-07-07 07:13:09.288195548 +0000
@@ -1,5 +1,5 @@
 fpp-check
-[ local path prefix ]/compiler/tools/fpp-check/test/event/missing_ports.fpp: 1.1
+[ local path prefix ]/compiler/tools/fpp-check/test/event/missing_ports.fpp:1.1
 passive component C {
 ^
 error: component with event specifiers must have event port
```


bocchino:
I like the idea, but why are Python files being added in this PR? It seems that only Location.scala and the unit test reference files should be changed.

SMorettini:
> I like the idea, but why are Python files being added in this PR? It seems that only Location.scala and the unit test reference files should be changed.

I didn't notice that when I push. I'll remove the python file and I'll proceed with the update of the tests.

SMorettini:
@bocchino the python files has been removed and the unit tests updated, the pull request is ready for review.

P.s.: thanks to @Blabla51 for helping to update the unit tests

bocchino (approved):
Looks good! Thanks for this update.

PR #275, opened 2023-07-11
Update release to include layout tools

LeStarch:


bocchino:
@LeStarch Can we close this PR? I think the functionality has been moved to the fprime-layout repo.

PR #280, opened 2023-07-13
Make sure comments are generated consistently

### LINKED ISSUE #278: [C++ templates] Make sure comments are generated consistently
Annotations are not currently turning into comments for command handlers. We should look at what the base class code gen is doing, and do something similar.

tiffany1618:
Closes #278


bocchino (approved):
Looks good! It's consistent now.

PR #283, opened 2023-07-17
Fpp to json

elliotfayman:
Introduces a new FPP tool fpp-to-json, that enables the conversion of Scala data structures to JSON files. The tool specifically targets the abstract syntax tree, location map, and analysis data structures, enabling the generation of JSON representations. This functionality allows external libraries to seamlessly ingest and utilize the compiler data.


compiler/tools/fpp-to-json/test/python/locationMapValidator.py
```diff
@@ -0,0 +1,50 @@
...
+    
+    for number in numbers:
+        if str(number) not in data:
+            return f"ID in ast not found in location map: {number}"
+    
+    return True
+
+
+
+with open('fpp-ast.json') as json_file:
```
bocchino, line 39: I think we can put the two open constructs together and have just one level of indenting. See https://stackoverflow.com/questions/26731720/how-to-avoid-nested-with-statements-when-working-with-multiple-files-in-python.

compiler/tools/fpp-to-json/test/python/locationMapValidator.py
```diff
@@ -0,0 +1,50 @@
+import json
+import sys
+import re
+
+def collect_ids(json_data):
```
bocchino, line 5: Let's add a short comment to document each function

compiler/tools/fpp-to-json/test/python/locationMapValidator.py
```diff
@@ -0,0 +1,50 @@
...
+
+    elif isinstance(json_data, dict):
+        for key, value in json_data.items():
+            if key == 'id':
+                # Check if the value is an integer
+                if isinstance(value, int):
+                    ids.append(value)
+                # Check if the value is a string and matches the format 'id : number'
+                elif isinstance(value, str) and re.match(r"id\s*:\s*\d+$", value):
+                    ids.append(int(value.split(':')[1].strip()))
```
bocchino, line 20: This code seems to assume that we can have a string "id: 5" in the output that actually represents an object with a field id with value 5. This should never happen in our JSON output. Can you confirm this never happens? If so, let's remove this code. If not, let's fix it in the JSON encoder.

compiler/lib/src/main/scala/codegen/JsonEncoder.scala
```diff
@@ -0,0 +1,698 @@
...
+  implicit val formalParamKindEncoder: Encoder[Ast.FormalParam.Kind] =
+    Encoder.encodeString.contramap(getUnqualifiedClassName(_))
+  implicit val generalPortInstanceKindEncoder: Encoder[PortInstance.General.Kind] =
+    Encoder.encodeString.contramap(getUnqualifiedClassName(_))
+  implicit val patternEncoder: Encoder[Ast.SpecConnectionGraph.Pattern.Kind] =
+    Encoder.encodeString.contramap(getUnqualifiedClassName(_))
+  implicit val severityEncoder: Encoder[Ast.SpecEvent.Severity] =
+    Encoder.encodeString.contramap(getUnqualifiedClassName(_))
+
+  implicit val typeEncoder: Encoder[Type] =
```
bocchino, line 54: I don't think we want to use toString here. We should preserve the representation of these objects as data structures.

bocchino: I also think these encoders belong with the analysis encoders below. Also, if possible we should use object {...} blocks to group the Ast encoders separately from the Analysis encoders.

bocchino: Can we move the Ast, Analysis, and LocMap encoders each into its own file? Then JsonEncoder would become a subdirectory, like CppWriter, etc.

bocchino (approved):
Looks good!

PR #294, opened 2023-07-24
Revise handling of incorrect options to FPP tools

bocchino:
Ensure that the tools return nonzero status when the command-line options are invalid. Previously, if the option parser returned a None value indicating a parse failure, the tools would exit with status zero.

### DIFF
--- a/compiler/tools/fpp-format/src/main/scala/fpp-format.scala
+++ b/compiler/tools/fpp-format/src/main/scala/fpp-format.scala
@@ -28,17 +28,18 @@ object FPPFormat {
     result match {
       case Left(error) => {
         error.print
-        System.exit(1)
+        errorExit
       }
-      case Right(_) => ()
+      case _ => ()
     }
   }
 
+  def errorExit = System.exit(1)
+
   def main(args: Array[String]) = {
-    val options = OParser.parse(oparser, args, Options())
-    options match {
+    OParser.parse(oparser, args, Options()) match {
       case Some(options) => command(options)
-      case None => ()
+      case None => errorExit
     }
   }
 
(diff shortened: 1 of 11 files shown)


LeStarch:
Ohhhhh I have studied and can review this now!

compiler/tools/fpp-format/src/main/scala/fpp-format.scala
```diff
@@ -28,17 +28,18 @@ object FPPFormat {
     result match {
       case Left(error) => {
         error.print
-        System.exit(1)
+        errorExit
```
LeStarch, line 31: Why is the error handled here, inside the command, and not up a level like all the other tools?

LeStarch (approved):
The code looks correct.  Here are some minor comments/questions but are non-blocking if you want to merge:

1. Should "errorExit" be defined in a standard place, instead of redefined in every tool?  e.g. `import fpp.lib.util.errorExit`?
2. Some tools handle the error in `command` and others handle it in the oparse matching expression. Both will work, but why not drive the error out all the way to the top-level so it is always handled in the same place.
3. I have a mild preference for:
```
val parsed_args = Oparse.....
parsed_args match {
   Some(options): command(options
   None: errorExit
}
```
It makes the code slightly less hard to read because you make the temporary value explicit. Might not be worth changing at this point.

compiler/tools/fpp-format/src/main/scala/fpp-format.scala
bocchino: No particular reason, it's just the way it's written.

bocchino:
Thanks, these are good points! I will address them.

bocchino:
I factored the boilerplate code for the main method into a function in `util.Tool`. This avoids repeating the boilerplate in each tool.

PR #295, opened 2023-07-25
Add capability to select the cpp file

### LINKED ISSUE #286: [CppDoc] Add capability to select the cpp file
Add an option field to each CppDoc member for selecting the cpp file to use when writing cpp. If the value of the field is None, we can use the default cpp file provided when the CppDoc object is constructed.

tiffany1618:
Closes #286


bocchino (approved):
Looks good! I did a little refactoring to consolidate repetitions of the matching logic.

PR #297, opened 2023-07-31
Make cpp file names consistent

tiffany1618:
Just one more fix to make cpp file names and file name bases consistent.

- `CppDocWriter.Input` and `CppDoc` structures store full file names for default and output cpp files
- To specify a cpp file other than the default, the user only needs to provide the file name base, both in `CppDoc` members and when calling helper functions for writing `CppDoc`s (e.g. `CppWriter.writeCppDoc`)


bocchino (approved):
Looks good.

PR #301, opened 2023-08-01
Fix conditional compilation in generated C++ code

### LINKED ISSUE #299: Fix conditional compilation in generated C++ code
Guards for functions that include `|| BUILD_UT` can cause compilation issues if the functions depend on other guards being defined. For example, this function:

https://github.com/fprime-community/fpp/blob/00c54e0b78357514cea9e27d6e25f50f222324f2/compiler/tools/fpp-to-cpp/test/array/AArrayAc.ref.cpp#L177-L201

will fail to compile if  `BUILD_UT` is on and `FW_ARRAY_TO_STRING` is off because `FW_ARRAY_TO_STRING_BUFFER_SIZE` is guarded by `FW_ARRAY_TO_STRING` (see https://github.com/nasa/fprime/blob/92c9f0712358ee34789c69a4cdc013f17748ad86/config/FpConfig.h#L311-L315).

To fix this, `|| BUILD_UT` should be removed from these guards.

tiffany1618:
Closes #299.

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ArrayCppWriter.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ArrayCppWriter.scala
@@ -466,7 +466,7 @@ case class ArrayCppWriter (
     )
     ) ++
       wrapClassMembersInIfDirective(
-        "\n#if FW_ARRAY_TO_STRING || BUILD_UT",
+        "\n#if FW_ARRAY_TO_STRING && BUILD_UT",
         List(
           functionClassMember(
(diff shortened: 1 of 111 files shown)


bocchino:
Before merging this PR, we need to remove all the instances of `|| BUILD_UT` and `&& BUILD_UT`.

bocchino:
Looks good! We should just confirm that it works with F Prime.

tiffany1618:
This passes the F Prime unit tests and the F Prime FPP tests.

bocchino (approved):
Looks good!

PR #305, opened 2023-08-03
Switch to new native-image building

LeStarch:
This changes the native build to use the standard native image, and provides binary wheels.

### DIFF
--- a/compiler/test
+++ b/compiler/test
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-# Fix when run from a differen directory
+# Ensure the script can be run from any directory
 wd=`dirname $0`
 cd $wd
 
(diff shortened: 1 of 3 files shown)


compiler/test
```diff
@@ -1,4 +1,8 @@
 #!/bin/sh
 
-export COMPILER_ROOT=$PWD
+# Fix when run from a differen directory
+wd=`dirname $0`
+cd $wd
+
+export COMPILER_ROOT=`cd $wd; pwd`
```
bocchino, line 7: Do we need the cd $wd in line 7? It looks redundant with line 5.

bocchino (commented):
Looks good! Just one comment.

compiler/test
LeStarch: Not needed, this is fixed!

bocchino (approved):
Looks great! It works because CI passes.

Issue #315, opened 2023-08-19
`fpp-to-cpp` Generates Broken C++


LeStarch:
In a component with a guarded command, the C++ will not compile.  See: https://github.com/fprime-community/fprime-baremetal-reference/issues/12

LeStarch:
Minimal reproduction test case:

**Component FPP**
```fpp
module Ref {
    @ Component for F Prime FSW framework.
    passive component MyComponent {
        @ Broken port
        guarded command COMMAND_NAME(param_name: U32)


        ###############################################################################
        # Standard AC Ports: Required for Channels, Events, Commands, and Parameters  #
        ###############################################################################

        @ Port for requesting the current time
        time get port timeCaller

        @ Port for sending command registrations
        command reg port cmdRegOut

        @ Port for receiving commands
        command recv port cmdIn

        @ Port for sending command responses
        command resp port cmdResponseOut

    }
}
```

**Error**
```bash
[100%] Building CXX object Ref/MyComponent/CMakeFiles/Ref_MyComponent.dir/MyComponentComponentAc.cpp.o
/Users/mstarch/code/fprime-infrastructure/fprime/Ref/build-fprime-automatic-native/Ref/MyComponent/MyComponentComponentAc.cpp:432:11: error: no member named 'lock' in 'Ref::MyComponentComponentBase'
    this->lock();
    ~~~~  ^
/Users/mstarch/code/fprime-infrastructure/fprime/Ref/build-fprime-automatic-native/Ref/MyComponent/MyComponentComponentAc.cpp:439:11: error: no member named 'unLock' in 'Ref::MyComponentComponentBase'
    this->unLock();
    ~~~~  ^
2 errors generated.
```

LeStarch:
Fixed!!!!

PR #316, opened 2023-08-19
Fix case where C++ does not compile on guarded command without guarded port

LeStarch:
Fix for #315


LeStarch:
Need to update UTs still, but the implementation works on the minimal test case shown in #315

LeStarch:
Unit tests failed to catch this issue because of this line:

https://github.com/fprime-community/fpp/blob/535c30de029f55612f4200d1987003ff086be178/compiler/tools/fpp-to-cpp/test/component/passive.fpp#L31

In the "passive command test", a set of superfluous  typed ports are included. This included a `guarded` port which fixes the issue.

We should consider removing that line such that the "pasive command unit test" just focuses on commands and nothing else.   However, this decision is likely outside the scope of this PR.

bocchino (approved):
This change looks good to me! Thanks for submitting the fix.

Issue #326, opened 2023-09-20
Fix shadow variable for enum generation


JohanBertrand:
Currently, when compiled with `-wShadow`, the auto generated enums are generating some warnings.

For example:
```
XEnumAc.hpp: warning: declaration of 'e' shadows a member of 'X::E_X' [-Wshadow]
   71 |             const T e //!< The raw enum value
      |             ~~~~~~~~^
note: shadowed declaration is here
  164 |         T e;
      |           ^
XEnumAc.hpp: warning: declaration of 'e' shadows a member of 'X::E_X'[-Wshadow]
  108 |         bool operator==(T e) const
      |                         ~~^
note: shadowed declaration is here
  164 |         T e;
      |           ^
XEnumAc.hpp: warning: declaration of 'e' shadows a member of 'X::E_X' [-Wshadow]
  114 |         bool operator!=(T e) const
      |                         ~~^
note: shadowed declaration is here
  164 |         T e;
      |           ^
```

Since the enum value are used in fprime using `my_enum.e`, I think that it is not recommended to change the member name `e` to `m_e`.

However, we could have a prefix/suffix on the function arguments which are shadowing the member value.

For example:
```cpp
      //! Constructor (user-provided value)
      E(
          const T e //!< The raw enum value
      )
      {
        this->e = e;
      }

      /* ... */

      //! Equality operator
      bool operator==(T e) const
      {
        return this->e == e;
      }

      //! Inequality operator
      bool operator!=(T e) const
      {
        return !(*this == e);
      }
```
With the `a_` prefix (for "argument") would become:
```cpp
      //! Constructor (user-provided value)
      E(
          const T a_e //!< The raw enum value
      )
      {
        this->e = a_e;
      }

      /* ... */

      //! Equality operator
      bool operator==(T a_e) const
      {
        return this->e == a_e;
      }

      //! Inequality operator
      bool operator!=(T a_e) const
      {
        return !(*this == a_e);
      }
```

I am open to suggestion if there is a better option/a better prefix or suffix.

bocchino:
I think the easiest fix is to rename `e` to `e1` everywhere it appears as an argument.

JohanBertrand:
Alright, I made the changes and created a pull request. 

However, I can see that the same problem can happen with the files generated by the `StructCppWriter` file, where the different members are having the same name as function argument.

Adding a suffix here might would not work, because a member variable called `m1` would possibly shadow a member called `m11`. This case is happening on the unit tests.

[...]

PR #327, opened 2023-09-21
Fix Enum shadow variable

JohanBertrand:
Renamed function argument `e` to `e1` to avoid shadow variable warnings, as mentioned in #326 

Also fixed `clean` script, to be able to run it from other directories, like `install` and `test`

### DIFF
--- a/compiler/clean
+++ b/compiler/clean
@@ -1,5 +1,9 @@
 #!/bin/sh -e
 
+# Ensure the script can be run from any directory
+wd=`dirname $0`
+cd $wd
+
 . scripts/utils.sh
 
 sbt clean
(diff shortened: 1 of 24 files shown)


bocchino:
You can run `./update-ref` to update the reference output. You may have already figured this out.

bocchino (approved):
Looks good!

JohanBertrand:
Good to know that there is a script for that, I've been always doing the changes manually.

Also, for some reasons, I am having some files generated by the `compiler/test` script, not ignored, in the folder `compiler/tools/fpp-to-cpp/test/component/test-impl/` and `compiler/tools/fpp-to-cpp/test/component/test-base/`. It seems like they are missing the `.out` suffix.
I'm not sure if it's a problem from my side or not. I'm just executing `compiler/install` followed by `compiler/test`.

bocchino:
> Also, for some reasons, I am having some files generated by the compiler/test script, not ignored

Yes, there is some missing `.gitignore` in those tests. This is fixed on the `data-products` branch. We expect to merge these changes to main soon.

PR #328, opened 2023-09-25
Fix incorrect generation in GTestBase

LeStarch:
Noticed this issue when integrating with F´.  In the process of testing it now.


bocchino (approved):
Looks good!

PR #335, opened 2023-10-03
Revise component cpp code gen for async buffers

### LINKED ISSUE #334: [C++ code gen] Optimize union type for async buffers
Currently the union type used to size async buffers includes the argument sizes of all typed input ports (in addition to commands and internal ports). It should be limited to async typed input ports, since sync port arguments never go on the async queue.

bocchino:
This PR fixes a small regression (minimal impact, performance only) in the FPP-to-C++ code gen for components. It also does some incidental code cleanup.

Closes #334.


compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriter.scala
```diff
@@ -288,42 +288,48 @@ case class ComponentCppWriter (
...
+      typedAsyncInputPorts.flatMap(p => {
+        val portName = p.getUnqualifiedName
+        val portTypeName = getQualifiedPortTypeName(p, p.getDirection.get)
+        lines(s"BYTE ${portName}PortSize[${portTypeName}::SERIALIZED_SIZE];")
+      }),
+      // Command input port
+      guardedList (cmdRecvPort.isDefined) (lines(s"BYTE cmdPortSize[Fw::InputCmdPort::SERIALIZED_SIZE];")),
+      // Internal ports
+      internalPorts.flatMap(p =>
+        line(s"// Size of ${p.getUnqualifiedName} argument list") ::
```
LeStarch, line 303: I followed the code up to this point.  Not sure what this is doing, nor why it doesn't look like the async block above.

bocchino:
@LeStarch It looks like the checks passed! Is this PR ready to merge?

PR #347, opened 2023-11-09
Making ping-entries an extern declaration

LeStarch:
This PR exposes `pingEntries` as an `extern` and adds a constant representing the number of entries it has.  This makes it easy for users to lift and use `pingEntries` from the autocoder.


compiler/lib/src/main/scala/util/Version.scala
```diff
@@ -3,6 +3,6 @@ package fpp.compiler.util
 /** The compiler version */
 object Version {
 
-  val v = "[unknown version]"
+  val v = "v2.0.0-4-gc8371371"
```
LeStarch, line 6: Revert this.

compiler/tools/fpp-to-cpp/test/top/HealthTopologyAc.ref.cpp
```diff
@@ -8,33 +8,31 @@
...
-      }
 
+  // ----------------------------------------------------------------------
+  // Component configuration objects
+  // ----------------------------------------------------------------------
+
+  namespace ConfigObjects {
+
+    namespace health {
+      Svc::Health::PingEntry pingEntries[] = {
```
LeStarch, line 19: Add num entries as NUM_PING_ENTRIES

bocchino (approved):
Looks good! Thanks for making this PR. It closes a loophole in the alternate way of configuring component instances.

PR #351, opened 2023-11-15
Fix bug in resolution of qualified identifiers

### LINKED ISSUE #350: Using a Topology Symbol as a Qualifier Causes a Crash
Adding the following line to the F´ `Ref` topology causes the below stack trace.  It should instead say "topology Ref.MyTopology does not exist"

```
   topology Ref {
+    import Ref.MyTopology
```

bocchino:
Properly report the error when a non-qualifier symbol is used as a qualifier.

**Example:**
```
topology M {
  import M.T
}
```
Note that the topology symbol `M` is incorrectly used as a qualifier in the qualified identifier `M.T`.

**Old behavior:** Crash.

**New behavior:**
```
fpp-check
/Users/bocchino/test.fpp:2.10
  import M.T
         ^
error: invalid symbol M: not a qualifier
symbol is defined here:
/Users/bocchino/test.fpp:1.1
topology M {
^
```

Closes #350.

### DIFF
--- a/compiler/lib/src/main/scala/analysis/CheckSemantics/CheckUses.scala
+++ b/compiler/lib/src/main/scala/analysis/CheckSemantics/CheckUses.scala
@@ -151,9 +151,21 @@ object CheckUses extends UseAnalyzer {
     ) = {
       for {
         a <- visitQualIdentNode (ng) (a, qualifier)
-        symbol <- {
+        scope <- {
           val symbol = a.useDefMap(qualifier.id)
-          val scope = a.symbolScopeMap(symbol)
+          a.symbolScopeMap.get(symbol) match {
+            case Some(symbol) => Right(symbol)
+            case None => Left(
+              SemanticError.InvalidSymbol(
+                symbol.getUnqualifiedName,
+                Locations.get(node.id),
+                "not a qualifier",
+                symbol.getLoc
+              )
+            )
+          }
+        }
+        symbol <- {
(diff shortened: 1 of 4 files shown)


LeStarch:
While this does fix the crash, it does not fix my expected use case:

Consider this:
```
module A {
   topology B {}
   topology A {
      import  A.B
   }
}
```

This fails with the error above (no crash, hurray), but I still expected to be able to refer to topology B by the fully qualified name `A.B` even within the context of the module `A`.

The following works:

```
module A {
   topology B {}
   topology A {
      import  B
   }
}
```

Here I can only refer to topology B in module A by the local name `B` because it is used in a topology `A`.  I'd have expected I could refer to `A.B` by its fully qualified names in any given context.

LeStarch (approved):
Although this does not full fix to what my expectations are, it does:

1. Remove the crash
2. Has a work-around

So I give my approval to merge, if you want to spin the other problem into another issue.

bocchino:
Sounds good.

>I'd have expected I could refer to A.B by its fully qualified names in any given context.

As discussed, you can't write A.B in a scope where the symbol A that you intend to use is hidden by a different symbol, also named A. This behavior is (1) correct according to the FPP spec (see s. 14.6 of the spec) and (2) standard for languages with scopes and name qualification, e.g., C++, Java. The recommended fix is to avoid symbol hiding, e.g., don't use the same name for a topology and its enclosing module.

PR #356, opened 2023-11-28
Fixing implementation include paths

### LINKED ISSUE #355: Fix Template Generation Include Paths
When generating template include paths, the relative path to the header is based on the current working directory of the execution.  However, this is inconsistent with other code generation where the include path is relative to the model file (symbol location).

This inconsistency should be fixed!

LeStarch:
This fixes #355!

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentImplWriter.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentImplWriter.scala
@@ -48,12 +48,9 @@ case class ComponentImplWriter(
   }
 
   private def getCppIncludes: CppDoc.Member = {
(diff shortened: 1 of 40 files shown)


LeStarch:
@thomas-bc here is the fix for the issue we discovered!  It was not as simple as I first indicated, as we needed to:

1. Acquire the Component symbol's location (FPP model file location defining symbol)
2. Convert that to a full path
3. Strip prefixes
4. Add .hpp
5. Pass to "make include statement" function (this was done already, but on erroneous input)

bocchino (approved):
Looks good! I made the corresponding changes to the headers in the test implementation.

bocchino:
I also refactored a bit to eliminate code duplication.

bocchino:
The CI failure is due to some issue with Python packages.

PR #357, opened 2023-12-07
Fixing issues regarding packaging and Python 12

LeStarch:
This should fix the build issues plaguing us of late!


bocchino (approved):
Looks great! Everything passed.

PR #362, opened 2023-12-08
Revise banner comments in generated C++ code

### LINKED ISSUE #360: Fix generated comments in C++ templates
* File name should be `Foo.hpp`, not `Foo.template.hpp`.
* Author should be the active user name, not `Generated by fpp-to-cpp`.

bocchino:
This PR fixes some issues with the banner comments in the generated "template" files for component implementations and unit test code.

Incorporating this as a "hot fix" against v2.0.1 because it is a small change, and it will be useful to include this in the upcoming point release of F Prime.

Closes #360.


compiler/lib/src/main/scala/codegen/CppWriter/ComputeCppFiles.scala
```diff
@@ -105,9 +105,35 @@ object ComputeCppFiles {
...
+      fileName.matches(".*TesterHelpers\\.cpp")
+
     /** Gets the C++ file name for generated component main test harness classes */
     def getComponentTestMain(baseName: String) = s"${baseName}TestMain"
 
+    /* Maps generated template patterns to handcoded file patterns */
+    val templateMap = Map(
+      "TestMain\\.cpp$" -> "TestMain.cpp",
+      "Tester\\.cpp$" -> "Tester.cpp",
+      "Tester\\.hpp$" -> "Tester.hpp",
```
tiffany1618, line 119: Are these mapping to the same thing?

tiffany1618 (commented):
Looks good!

compiler/lib/src/main/scala/codegen/CppWriter/ComputeCppFiles.scala
bocchino: Yes, the map is the identity map in those cases.

LeStarch (approved):

LeStarch:
The new output looks correct as shown in the updated UT correctness files.

PR #366, opened 2023-12-19
Add index and size of array when asserting index is smaller than the SIZE

SMorettini:
### Change

Added in the assert the index and the SIZE. Since the assert accepts multiple parameters I think it would help to know the index that makes the assert fail.


bocchino (approved):
Looks good!

PR #367, opened 2023-12-20
Made the size of the buffer storing the port name less arbitrary

SMorettini:
### Problem

In the initialization of each component, a buffer of size `120` was used to store the name of the port(component name + port name + some other characters). 

The value `120` was an arbitrary value. If the `FW_OBJ_NAME_MAX_SIZE` is set to a value bigger than `120` or near `120`, a lot of warning would be produced when building an Fprime executable. And there is no way to remove the warnings because the `120` is hard coded.

### Solution

Changed 120 with `FW_OBJ_NAME_MAX_SIZE*2`. 

### How to reproduce the warning

1. Apply the following changes:
```
diff --git a/Ref/CMakeLists.txt b/Ref/CMakeLists.txt
index 766689ec3..b1b3b4805 100644
--- a/Ref/CMakeLists.txt
+++ b/Ref/CMakeLists.txt
@@ -54,6 +54,7 @@ set(MOD_DEPS ${PROJECT_NAME}/Top)
 register_fprime_deployment()
 # The following compile options will only apply to the deployment executable.
 # The extra warnings trigger in core F Prime so we don't apply them there.
+target_compile_options("${PROJECT_NAME}" PUBLIC -pedantic)
 target_compile_options("${PROJECT_NAME}" PUBLIC -Wall)
 target_compile_options("${PROJECT_NAME}" PUBLIC -Wextra)
 target_compile_options("${PROJECT_NAME}" PUBLIC -Werror)
diff --git a/Ref/SignalGen/SignalGen.fpp b/Ref/SignalGen/SignalGen.fpp
index 61110fa53..e0ca3899e 100644
--- a/Ref/SignalGen/SignalGen.fpp
+++ b/Ref/SignalGen/SignalGen.fpp
@@ -29,7 +29,7 @@ module Ref {
     # General Ports
     # ----------------------------------------------------------------------
 
-    sync input port schedIn: Svc.Sched
+    sync input port schedIn_long: Svc.Sched
 
     # ----------------------------------------------------------------------
     # Special ports
diff --git a/config/FpConfig.h b/config/FpConfig.h
index 12e92bd7d..06322ec37 100644
--- a/config/FpConfig.h
+++ b/config/FpConfig.h
@@ -174,7 +174,7 @@ typedef U16 FwTlmPacketizeIdType;
 #if FW_OBJECT_NAMES
 #ifndef FW_OBJ_NAME_MAX_SIZE
 #define FW_OBJ_NAME_MAX_SIZE \
-    80  //!< Size of object name (if object names enabled). AC Limits to 80, truncation occurs above 80.
+     120  //!< Size of object name (if object names enabled). AC Limits to 80, truncation occurs above 80.
 #endif
```
2. Run `fprime-util generate`
3. Run `cd SignalGen; fprime-util build`

One of the warning produced will look like 
```
/workspaces/fprime-with-devcontainer/fprime/Ref/build-fprime-automatic-native/Ref/SignalGen/SignalGenComponentAc.cpp:123:9: warning: ‘_schedIn_long_InputPort[’ directive output may be truncated writing 24 bytes into a region of size between 1 and 120 [-Wformat-truncation=]
  123 |         "%s_schedIn_long_InputPort[%" PRI_PlatformIntType "]",
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/workspaces/fprime-with-devcontainer/fprime/Ref/build-fprime-automatic-native/Ref/SignalGen/SignalGenComponentAc.cpp:120:22: note: ‘snprintf’ output between 27 and 156 bytes into a destination of size 120
  120 |       (void) snprintf(
      |              ~~~~~~~~^
  121 |         portName,
      |         ~~~~~~~~~     
  122 |         sizeof(portName),
      |         ~~~~~~~~~~~~~~~~~
  123 |         "%s_schedIn_long_InputPort[%" PRI_PlatformIntType "]",
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  124 |         this->m_objName,
      |         ~~~~~~~~~~~~~~~~
  125 |         port
      |         ~~~~          
  126 |       );
      |       ~    
```
Produced from 
```
...
#if FW_OBJECT_NAMES == 1
      char portName[120];
      (void) snprintf(
        portName,
        sizeof(portName),
        "%s_cmdIn_InputPort[%" PRI_PlatformIntType "]",
        this->m_objName,
        port
      );
      this->m_cmdIn_InputPort[port].setObjName(portName);
#endif
...
```


timcanham:
The `120` magic number is certainly wrong, but to be consistent, why wouldn't we just make it `FW_OBJ_NAME_MAX_SIZE `? That constant is meant to bound the size of an object name, and the port name is an object name.

SMorettini:
Because it would not solve the warning. The warning is caused by the fact that the target buffer should be bigger than the size of `objName`(that is of size `FW_OBJ_NAME_MAX_SIZE`) + the size of the string generated like `_productSendOut_OutputPort[`.

That's why I suppose previously was 120 that was bigger than  `FW_OBJ_NAME_MAX_SIZE`

timcanham:
To capture it, can you post the warning?

timcanham:
Could we see the code it is flagging?

SMorettini:
> Could we see the code it is flagging?

I improve the pull request description with all the information, let me know if you need more :)

timcanham:
Fair enough! I'll let @LeStarch and @bocchino weigh in too, but it seems to me that we should allow truncation without warnings, since even `FW_OBJ_NAME_MAX_SIZE*2` could get small enough to trigger the warning on small platforms. Is this a warning we want to disable? Truncating doesn't result in bad behavior, just incomplete strings.

bocchino:
The code generator knows the size of everything but the object name, right? So it seems the generated code can have an array that's known to be big enough, i.e., "max object size + known size of everything else."

SMorettini:
If you just do  `char portName[FW_OBJ_NAME_MAX_SIZE];` there will be a warning for every port because the output string is of size `FW_OBJ_NAME_MAX_SIZE` and the input will be of size `FW_OBJ_NAME_MAX_SIZE+ the size of the hardcoded text`, in the previous example the hardcoded text I'm referring is `s_schedIn_long_InputPort`. 

`portName` is just a temporary storage for a string that will be stored lately in a buffer of `FW_OBJ_NAME_MAX_SIZE` so if the string contained in `portName` is bigger than `FW_OBJ_NAME_MAX_SIZE`, the `setObjName` will truncate it without a warning.

SMorettini:
> The code generator knows the size of everything but the object name, right? So it seems the generated code can have an array that's known to be big enough, i.e., "max object size + known size of everything else."

Actually these seems a better solution, I'll implement it

bocchino:
Correct. So for example, in the case the string is "%s123%d456" the code generator can generate
```
char portName[FW_MAX_OBJ_NAME_SIZE + 27]
```
and the string is guaranteed to fit if I've done the math right, because the %s fits in FW_MAX_OBJ_NAME_SIZE characters and the "123%d456" fits in 27 characters -- 3 characters for 123, 20 characters for %d (conservatively estimating the max decimal string as 2^64 in decimal which is 18446744073709551616), 3 characters for 456 and one character for the terminating zero.

bocchino:
Or we could reserve 10 characters for %d, or even a smaller number, since these are port indices and are not likely to go over a few hundred.

SMorettini:
@bocchino I updated my change. Is this good or do you want me to sum the four integers directly in Scala? Let me know when you think so then I'll update the tests.

bocchino:
Please do as much computation as you can (e.g., sum all integer constants) before generating code. Note: To update the unit tests, you can run `./update-ref`.

bocchino:
Thanks. After looking at the generated code, I changed my mind. I think it's better to generate code that makes clear what the sum is, rather than generating code that has magic numbers in it. I made that change.

bocchino (approved):
Looks good!

bocchino:
Comments from @LeStarch:
1. Use single quotes instead of double quotes where possible, to avoid the terminating zero.
2. Check that `sizeof "abcdef"` returns the size of `"abcdef"` (i.e., 7).

bocchino:
With regard to point 2, this C++ program, when compiled and run, prints `7`:
```
#include <cstdio>

int main(int argc, char **argv) {
  printf("%lu\n", sizeof "abcdef");
  return 0;
}
```

bocchino:
So after all this I actually agree with Tim's initial observation: the simplest and most consistent approach is just to use `FW_OBJ_NAME_MAX_SIZE` as the size of the char array. With regard to this comment:

> If you just do  char portName[FW_OBJ_NAME_MAX_SIZE]; there will be a warning for every port because the output string is of size FW_OBJ_NAME_MAX_SIZE and the input will be of size FW_OBJ_NAME_MAX_SIZE+ the size of the hardcoded text

If that's true for your deployment, then it seems the solution is to increase `FW_OBJ_NAME_MAX_SIZE` for your deployment so everything fits. I think Tim is correct that the port name is an object name. "Object name" here means "name of a C++ object" not "name of an F Prime component."

bocchino:
Also, the code that sets the object name does a strncpy into an array of size `FW_OBJ_NAME_MAX_SIZE`. So any extra bytes beyond that will be thrown away during the copy in any event.

See https://github.com/nasa/fprime/blob/05ff4447fec19bfb1b1b8070e15f9f750528e6cc/Fw/Obj/ObjBase.cpp#L44-L47.

SMorettini:
In a previous [comment](https://github.com/fprime-community/fpp/pull/367#issuecomment-1864878068), I explained already how the solutions selected don't solve the warning, actually the current solution create always warning whatever the size of `FW_OBJ_NAME_MAX_SIZE`:

> Because it would not solve the warning. The warning is caused by the fact that the target buffer should be bigger than the size of objName(that is of size FW_OBJ_NAME_MAX_SIZE) + the size of the string generated like _productSendOut_OutputPort[.
> 
> That's why I suppose previously was 120 that was bigger than FW_OBJ_NAME_MAX_SIZE

To be very clear: **_now everyone will have the warning if compiles with the pedantic flag_**.

So the solution of @bocchino :
>If that's true for your deployment, then it seems the solution is to increase FW_OBJ_NAME_MAX_SIZE

is not a valid solution if the objective is to build my software with the pedantic flag. Or better is not a solution to any Fprime project.

Here is the warning using the code merged in main:
```
/workspaces/Fprime/build-fprime-automatic-native/F-Prime/Svc/ActiveLogger/ActiveLoggerComponentAc.cpp:132:9: error: ‘_LogRecv_InputPort[’ directive output may be truncated writing 19 bytes into a region of size between 1 and 80 [-Werror=format-truncation=]
  132 |         "%s_LogRecv_InputPort[%" PRI_PlatformIntType "]",
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/workspaces/Fprime/build-fprime-automatic-native/F-Prime/Svc/ActiveLogger/ActiveLoggerComponentAc.cpp:129:22: note: ‘snprintf’ output between 22 and 111 bytes into a destination of size 80
  129 |       (void) snprintf(
      |              ~~~~~~~~^
  130 |         portName,
      |         ~~~~~~~~~     
  131 |         sizeof(portName),
      |         ~~~~~~~~~~~~~~~~~
  132 |         "%s_LogRecv_InputPort[%" PRI_PlatformIntType "]",
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  133 |         this->m_objName,
      |         ~~~~~~~~~~~~~~~~
  134 |         port
      |         ~~~~          
  135 |       );
      |       ~   
```
I put only one warning but there are hundreds more, one for each port in Fprime.

About:
> Also, the code that sets the object name does a strncpy into an array of size FW_OBJ_NAME_MAX_SIZE. So any extra bytes beyond that will be thrown away during the copy in any event.
> 
> See https://github.com/nasa/fprime/blob/05ff4447fec19bfb1b1b8070e15f9f750528e6cc/Fw/Obj/ObjBase.cpp#L44-L47.

 I agree and I think that if want to catch the cutting I could add an `FW_ASSERT` there 

@timcanham  @LeStarch @bocchino  Now the question is: Do Fprime target to have a code that compiles with no warning when using the `-pedantic` flag? 

1. If yes, we need to reopen this pull request;
2. If not, let me know so I'll continue maintaining my internal fork.

bocchino:
OK, I think I understand the issue now. It looks like in this code:

https://github.com/fprime-community/fpp/blob/182eea658546550890aa87350c0b74b928a01bec/compiler/tools/fpp-to-cpp/test/component/base/ActiveAsyncProductPortsOnlyComponentAc.ref.cpp#L89-L96

the compiler is looking at the argument `this->m_objName` and since `m_objName` is an array, it is using the size of the array (i.e., FW_OBJ_NAME_MAX_SIZE) as the max possible size of the string argument. So it computes the max string size as FW_OBJ_NAME_MAX_SIZE + literal string size, which exceeds FW_OBJ_NAME_MAX_SIZE. I mistakenly thought the issue was that the literal character string was overflowing the buffer. In that case the solution would be to increase `FW_OBJ_NAME_MAX_SIZE`.

Given what the compiler is actually doing, I think the original solution of `2*FW_OBJ_NAME_MAX_SIZE` is a good choice for the buffer size. It's simple and it's correct, since we expect everything to fit inside `FW_OBJ_NAME_MAX_SIZE`. It's just that the compiler doesn't know that.

With regard to compiling with -Wpedantic, that is a larger issue. I suggest you open an issue on this on nasa/fprime. It seems to me that if our code is close to compiling with -Wpedantic, then it would be a good idea to turn that flag on in the default build, and maintain the code with that flag on.

bocchino:
With regard to this part of Tim's comment:

> but it seems to me that we should allow truncation without warnings, since even FW_OBJ_NAME_MAX_SIZE*2 could get small enough to trigger the warning on small platforms.

My opinion is that in this case, the solution is to increase FW_OBJ_NAME_MAX_SIZE.

bocchino:
And in that case I think the warnings are a good thing (they are telling us that FW_OBJ_NAME_MAX_SIZE is set too small).

PR #368, opened 2023-12-20
Resolve warning array of size zero

SMorettini:
### Problem

In `BuffUnion` arrays of size zero were created. This is bad practice and it can create warnings when compiling

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriter.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriter.scala
@@ -300,7 +300,7 @@ case class ComponentCppWriter (
     // For each one, add a byte array of that size as a member
     val members = List.concat(
       // Data product and typed async input ports
-      (dataProductAsyncInputPorts ++ typedAsyncInputPorts).filterNot(p => p.size > 0).flatMap(p => {
+      (dataProductAsyncInputPorts ++ typedAsyncInputPorts).filter(p => getPortParams(p).size > 0).flatMap(p => {
         val portName = p.getUnqualifiedName
         val portTypeName = getQualifiedPortTypeName(p, p.getDirection.get)
         lines(s"BYTE ${portName}PortSize[${portTypeName}::SERIALIZED_SIZE];")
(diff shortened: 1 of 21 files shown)


compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriter.scala
```diff
@@ -352,8 +350,8 @@ case class ComponentCppWriter (
          |
          |    enum {
          |      // Max. message size = size of data + message id + port
-         |      SERIALIZATION_SIZE =
-         |        sizeof(BuffUnion) +
+         |      SERIALIZATION_SIZE =${if (getBuffUnion.nonEmpty) """
```
bocchino, line 353: I think this change is not necessary. This function is called only if the component is not passive. In that case, BuffUnion cannot be empty. See https://github.com/fprime-community/fpp/blob/26e5ace8d17807fa5caf726f70636fea4d0a6e36/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriter.scala#L261-L278.

bocchino (changes requested):
Looks good, please see the comment.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriter.scala
SMorettini: Why it cannot be empty? 

For example, a component could have only ports with no arguments. --> This will cause the `getBuffUnion` to return nothing. --> `sizeOf(BuffUnion)` cannot exist because `BuffUnion` doesn't exist in the code.
Maybe I can add a UT to show the case

bocchino: OK, good point. Yes, can you add a unit test to cover the case you identified? Thanks.

SMorettini: I added the test

bocchino:
Comments from @LeStarch:
1. Use single quotes instead of double quotes where possible, to avoid the terminating zero.
2. Check that `sizeof "abcdef"` returns the size of `"abcdef"` (i.e., 7).

bocchino:
With regard to point 2, this C++ program, when compiled and run, prints `7`:
```
#include <cstdio>

int main(int argc, char **argv) {
  printf("%lu\n", sizeof "abcdef");
  return 0;
}
```

bocchino (approved):

bocchino:
Looks good! I refactored the code a bit and addressed the case of internal ports with no formal parameters.

PR #372, opened 2024-01-03
Revise code generation for data products

bocchino:
This PR revises the component code generation for data products. It conforms to changes in the data products design documented in docs/Design/data-products.md at https://github.com/bocchino/fprime/commit/70601e3b3650cbde59655a8c8109f816851947e7.

* When requesting a data product buffer, users now request the data size, not the buffer size. The auto-generated code computes the buffer size.
* The data product header spec has changed, based on feedback from the DpWriter design review.
* Improve the fpp-to-json test to catch more errors.

check-cpp passes against https://github.com/bocchino/fprime/commit/70601e3b3650cbde59655a8c8109f816851947e7.


bocchino:
@LeStarch Build Native Images for Linux reports that the command `python3` does not exist. See https://github.com/fprime-community/fpp/actions/runs/7400045898/job/20133303295?pr=372. This error may have been masked before because of the way the unit test was written. Now the test is more robust, but the CI issue is exposed, causing a failure. Can you investigate?

bocchino:
For now I worked around the issue by skipping the check if python3 is not there.

PR #381, opened 2024-01-13
Fix shadow variables

JohanBertrand:
More details in https://github.com/nasa/fprime/pull/2482

Fixes in the code generation to remove -Wshadow warnings:
- Added `m_` for fprime component member variables


LeStarch (commented):
I see what I expect....but it seems like certain models where updated.  I like those changes, but will defer to @bocchino to determine if they are desired.

bocchino (approved):
Looks good! I reverted two changes: (1) I restored the old behavior of the toString conversion. The names of the logical struct variables are the names shown in FPP. The m_ names are only for the C++ implementation. (2) I restored the FPP variable names in the unit tests.

JohanBertrand:
Thanks!

I originally changed the names of the variables in the fpp files, because I thought we should not advise users to add a `m` prefix in front of the variables they defined.

bocchino:
I think it's OK in the context of a unit test. In real code, a struct member name should say what the variable does. In a unit test context, it seems fine to call a U32 member `mU32`, since it really has no purpose except to be a U32 member in the context of that test.

PR #383, opened 2024-01-18
Increase size of temporary buffer in C++ code gen

bocchino:
* Increase FW_OBJ_NAME_MAX_SIZE to 2*FW_OBJ_NAME_MAX_SIZE as suggested by @SMorettini 
* Add a comment explaining why we do this

This change fixes a warning that can occur with -Wpedantic enabled. See the comments in #367.

### DIFF
--- a/compiler/tools/fpp-to-cpp/test/component/base/ActiveGetProductsComponentAc.ref.cpp
+++ b/compiler/tools/fpp-to-cpp/test/component/base/ActiveGetProductsComponentAc.ref.cpp
@@ -250,7 +250,13 @@ void ActiveGetProductsComponentBase ::
     this->m_cmdIn_InputPort[port].setPortNum(port);
 
 #if FW_OBJECT_NAMES == 1
-    char portName[FW_OBJ_NAME_MAX_SIZE];
+    // The port name consists of this->m_objName and some extra info.
+    // We expect all of this to fit in FW_OBJ_NAME_MAX_SIZE bytes.
+    // However, the compiler may assume that this->m_objName fills
+    // the entire array, whose size is FW_OBJ_NAME_MAX_SIZE. So to
+    // avoid a compiler warning, we provide an extra FW_OBJ_NAME_MAX_SIZE
+    // bytes to cover the extra info.
+    char portName[2*FW_OBJ_NAME_MAX_SIZE];
     (void) snprintf(
       portName,
       sizeof(portName),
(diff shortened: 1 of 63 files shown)


SMorettini (approved):

compiler/tools/fpp-to-cpp/test/component/base/ActiveGetProductsComponentAc.ref.cpp
```diff
@@ -250,7 +250,13 @@ void ActiveGetProductsComponentBase ::
...
 
 #if FW_OBJECT_NAMES == 1
-    char portName[FW_OBJ_NAME_MAX_SIZE];
+    // The port name consists of this->m_objName and some extra info.
+    // We expect all of this to fit in FW_OBJ_NAME_MAX_SIZE bytes.
+    // However, the compiler may assume that this->m_objName fills
+    // the entire array, whose size is FW_OBJ_NAME_MAX_SIZE. So to
+    // avoid a compiler warning, we provide an extra FW_OBJ_NAME_MAX_SIZE
+    // bytes to cover the extra info.
+    char portName[2*FW_OBJ_NAME_MAX_SIZE];
```
LeStarch, line 259: The logic holds, per my brief review, as long as `sizeof "_noArgsAsync_InputPort[%" ... "] < FW_OBJ_NAME_MAX_SIZE`.  Otherwise compiler warnings come back....which was an advantage of the previous implementation we saw with the `sizeof` operators.  It ensured enough space for the literals and then let it get cut down to size in the `.setObjName` call.  It also used less stack space when  `sizeof "_noArgsAsync_InputPort[%" ... "] <<< FW_OBJ_NAME_MAX_SIZE`  (by `<<<` I mean much less than).

I don't know if that precision is worth the original complexity...but I can say that it does get annoying when compiler warnings come out by just changing config.

On smaller systems, doubling the stack allocation for a port call here could be significant.  I guess the argument is that object names are not required and as such could be turned off? I can buy that argument.  @timcanham do you agree that turning off FW_OBJECT_NAMES is appropriate for smaller systems?

LeStarch: I saw a comment that addresses this in the other PR.  Essentially the argument is that the warning in the case I describe *is valid* because `FW_OBJECT_NAMES` is too small to hold the literal and break.  There are two valid non-warning solutions:

1. Turn off object names
2. Increase object name size.

LeStarch (approved):

bocchino:
I chatted with @timcanham  about this. He agrees with this solution in the near term. In the longer term, he would like to avoid compiler warnings about truncating object names, even in cases where the names don't fit. He envisions that small systems may want a small object name size, and may want to tolerate the truncation.

We think a good way to implement that behavior is to create an `Fw::ObjectName` implementation of `Fw::StringBase`, sized to `FW_OBJ_NAME_MAX_SIZE`. Then instead of using `snprintf` into a `char` array we can use the `format` method of `Fw::StringBase`. This is better practice anyway.

This solution requires integrated changes to F Prime and FPP, so I'll open separate issues for these propose changes.

compiler/tools/fpp-to-cpp/test/component/base/ActiveGetProductsComponentAc.ref.cpp
bocchino: I agree these are the near-term solutions. Tim and I discussed a longer-term solution that allows silent truncation of long object names, and avoids complex coding. See below.

PR #388, opened 2024-02-01
Fix regression in queue size check

bocchino:
Closes nasa/fprime#2508.


LeStarch (approved):
Code looks right to me. Here is what I see:

1. In the case of "passive" do nothing.  In the other case (active/queued) produce an invalid right (aka error)
2. I assume the other is plumbing for clearer error messages
3. The UT updates look right

PR #389, opened 2024-02-05
Remove BSD-like text from NOTICE.txt

thomas-bc:
Addressing https://github.com/fprime-community/fpp/issues/232

We had the same NOTICE in core F´ and addressed the issue (https://github.com/nasa/fprime/issues/1904) in https://github.com/nasa/fprime/pull/1915


LeStarch (approved):

bocchino (approved):
Looks good!

PR #395, opened 2024-03-07
Add FPP JSON Dictionary

jwest115:
This PR adds:
- Dictionary data structure and construction ([#2274](https://github.com/nasa/fprime/issues/2274))
- Ability to generate JSON dictionaries from FPP topologies ([#2479](https://github.com/nasa/fprime/issues/2479))
- `fpp-to-dict` command line tool ([#2276](https://github.com/nasa/fprime/issues/2276))
- Unit tests for `fpp-to-dict` tool ([#2501](https://github.com/nasa/fprime/issues/2501))


compiler/tools/fpp-to-dict/test/scripts/run.sh
```diff
@@ -53,11 +53,16 @@ validate_json_schema()
   # Check to see if the dictionary JSON schema is valid
   if which python3 > /dev/null 2>&1
```
bocchino, line 54: This looks good, but I think we can clean up the logic. I don't think we need two nested if statements: the outer one tests for `python3`, but if there's no `python3`, then the inner one will fail. So I'd delete the outer one and keep the inner one.

thomas-bc:
Currently, using a type defined in FPP looks like this:
```
{
    "name": "MyModule.MyType",
    "kind": "qualifiedIdentifier",
}
```
To figure out what MyType actually is (enum, struct, array) one has to go find the type MyType defined in one of those 3 sub-dictionary which isn't ideal, both for a machine and for human-readability.
What if we added an entry to have something like this
```
{
    "name": "MyModule.MyType",
    "kind": "qualifiedIdentifier",
    "kind_type": "array"
}
```

Probably this could be sufficient too, since we know any of those 3 has to be looked up at a qualified identifier;
```
{
    "name": "MyModule.MyType",
    "kind": "array|struct|enum"
}
```


I believe this would be implemented [here](https://github.com/jwest115/fpp/blob/bbb09c65400d29618ec7e9831cc9a73d2bea7c2d/compiler/lib/src/main/scala/codegen/DictionaryJsonWriter/DictionaryJsonEncoder.scala#L154-L171) ?

bocchino:
> To figure out what MyType actually is (enum, struct, array) one has to go find the type MyType defined in one of those 3 sub-dictionary

I think we got rid of those sub-dictionaries, so you just look up the name in a single dictionary of type definitions.

bocchino:
Confirmed, there are no sub-dictionaries. See here: https://github.com/nasa/fprime/blob/a771b3540bcfa58c507b9d368fc91a3f0debaf2e/docs/Design/fpp-json-dict.md#dictionary-content.

thomas-bc:
Ok that helps. Still - do we not think it's irrelevant to add a field saying which of the three it is? It would help for both human readability and parsing.
Probably just need an extra field [there](https://github.com/jwest115/fpp/blob/bbb09c65400d29618ec7e9831cc9a73d2bea7c2d/compiler/lib/src/main/scala/codegen/DictionaryJsonWriter/DictionaryJsonEncoder.scala#L154-L174) ?

bocchino:
I don't think we need that extra field repeating partial info about the type. The qualified type name points to the type definition, which has all the info, e.g., not just whether it's an array but the entire shape of the array. This is also how it works in FPP source.

bocchino:
Also, the JSON dictionary is not primarily designed for human readability. We can always write a script that inlines the entire type definition at every type name, if we think that is useful. It's a tradeoff -- lookup vs. verbosity. I think this is a concern for a downstream tool, not the dictionary spec.

bocchino:
Recording the array/struct/enum info does not help with machine processing. It's just a name lookup in a table.

bocchino (approved):
Looks great! Thanks @jwest115 and @thomas-bc for all your hard work on this!

PR #400, opened 2024-04-08
Fix missing assert conversion

JohanBertrand:
This is rising an error when compiling with -wConversion (conversion from unsigned int to int)


LeStarch:
@bocchino would you review this?  It is short, and looks reasonable to me.

LeStarch:
@JohanBertrand we may need to update the UT reference files too.  @bocchino is that process written somewhere we can link to?

JohanBertrand:
I updated the UT, it's supposed to be ready for review

bocchino (approved):
Looks good!

PR #408, opened 2024-04-12
Fix C++ code gen for string records

### LINKED ISSUE #392: [Data products] String type records don't work properly
~The code generator is not generating the string type that's needed for the argument to the serialization function. This issue is related to #387. We should use the same strategy for generating string type names.~

We are using the strategy described in #406. This is our go-forward strategy for strings in FPP.

bocchino:
Closes #392.


compiler/tools/fpp-to-cpp/test/component/base/ActiveGetProductsComponentAc.ref.cpp
```diff
@@ -142,17 +144,80 @@ Fw::SerializeStatus ActiveGetProductsComponentBase::DpContainer ::
...
+    sizeof(FwDpIdType) +
+    sizeof(FwSizeStoreType);
+  for (FwSizeType i = 0; i < size; i++) {
+    const Fw::StringBase *const sbPtr = array[i];
+    FW_ASSERT(sbPtr != nullptr);
+    sizeDelta += sbPtr->serializedTruncatedSize(stringSize);
+  }
+  // Serialize the elements if they will fit
+  Fw::SerializeStatus status = Fw::FW_SERIALIZE_OK;
+  if (this->m_dataBuffer.getBuffLength() + sizeDelta <= this->m_dataBuffer.getBuffCapacity()) {
```
LeStarch, line 166: Prefer () instead of relying on order of operations:  `(this->m_dataBuffer.getBuffLength() + sizeDelta)`

LeStarch (approved):
Minor non-blocking comment.

compiler/tools/fpp-to-cpp/test/component/base/ActiveGetProductsComponentAc.ref.cpp
bocchino: Good point! I will fix it.

bocchino: Fixed!

PR #430, opened 2024-05-22
Just spec changes for state machine

garthwatney:


bocchino (changes requested):
This branch can't be merged as is. It's based on a branch that's out-of-date with fpp/main, and it collects all the changes into a single commit. But we want some of the changes and not others. In particular, merging this PR would wipe out all the info in the spec that was recently added to describe data products. I think this PR will have to be closed and another PR made based on a branch that is up-to-date with fpp/main.

PR #431, opened 2024-05-23
FPP State machine behavior spec

garthwatney:
State machine behavior spec


bocchino (approved):
Looks good! I made some minor revisions.

PR #443, opened 2024-06-07
Add unit test for duplicate file names

jwest115:
Write `fpp-to-dict` output to txt file and compare against reference (similar to other tools). Added unit test case for duplicate file names.

### DIFF
--- a/compiler/tools/fpp-to-dict/test/scripts/run.sh
+++ b/compiler/tools/fpp-to-dict/test/scripts/run.sh
@@ -20,7 +20,7 @@ run_test()
   else
     outfile=$infile
   fi
-  $fpp_to_dict $args $infile.fpp 2>&1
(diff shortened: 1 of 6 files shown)


bocchino (commented):
Looks good, thanks! Why are the empty files dataProducts.ref.txt and multipleTops.ref.txt checked in here? Do we need them?

jwest115:
I don't think we need them, I decided to include them to keep things consistent with other tool unit tests, for example `fpp-to-cpp` https://github.com/nasa/fpp/blob/main/compiler/tools/fpp-to-cpp/test/enum/default.ref.txt.

bocchino:
OK, I agree that makes sense now that we are checking for console output. Thanks!

bocchino (approved):
Looks good!

PR #444, opened 2024-06-08
Fix issues in fpp-depend unit tests

bocchino:
* Regenerate test output
* Fix bug in tests that was masking test failures


jwest115 (approved):
Looks good to me!

PR #447, opened 2024-06-18
Remove FW_ARRAY_TO_STRING #375

### LINKED ISSUE #375: Remove FW_ARRAY_TO_STRING
The compilation that is now conditioned on FW_ARRAY_TO_STRING should be conditioned on FW_SERIALIZABLE_TO_STRING. It doesn't make sense to treat arrays separately from structs, because structs can have arrays in them. In other words, FW_SERIALIZABLE_TO_STRING should not govern just the code gen for structs: it should govern the code gen of all serializable types.

Related to #303.

jwest115:
- Updated array cpp codegen to use `FW_SERIALIZABLE_TO_STRING` instead of `FW_ARRAY_TO_STRING`
- Removed `|| FW_ENABLE_TEXT_LOGGING` in enum code gen. Per the static assertions in the F Prime config, if text logging is on, then serializable strings are on. See https://github.com/nasa/fprime/blob/af76f089e5aaee774d129e9005e19aaeab36e2f0/Fw/Cfg/ConfigCheck.cpp#L39-L42.
- Regenerated fpp-to-cpp test reference files
- Updated examples in the User's Guide and removed outdated code.

Closes #375.


compiler/lib/src/main/scala/codegen/CppWriter/ArrayCppWriter.scala
```diff
@@ -477,7 +477,7 @@ case class ArrayCppWriter (
     )
     ) ++
       wrapClassMembersInIfDirective(
-        "\n#if FW_ARRAY_TO_STRING",
+        "\n#if FW_SERIALIZABLE_TO_STRING",
```
bocchino, line 480: Can we change the guard to `#ifdef FW_SERIALIZABLE_TO_STRING` per the discussion in the other issue?

jwest115: Yes, can do!

bocchino: I think we decided to implement on/off switches as an ifdef check.

jwest115: This should be resolved in recent commit!

compiler/lib/src/main/scala/codegen/CppWriter/EnumCppWriter.scala
```diff
@@ -355,12 +355,12 @@ case class EnumCppWriter(
...
         List(
           linesClassMember(
-            lines("\n#if FW_SERIALIZABLE_TO_STRING"),
+            lines("\n#ifdef FW_SERIALIZABLE_TO_STRING"),
             CppDoc.Lines.Cpp
           )
         ),
         wrapClassMembersInIfDirective(
-          "\n#if FW_SERIALIZABLE_TO_STRING || FW_ENABLE_TEXT_LOGGING",
+          "\n#ifdef FW_SERIALIZABLE_TO_STRING || FW_ENABLE_TEXT_LOGGING",
```
bocchino, line 363: I think we should delete ` || FW_ENABLE_TEXT_LOGGING` here. If `FW_SERIALIZABLE_TO_STRING` is off and `FW_ENABLE_TEXT_LOGGING` is on, then the build is broken. I confirmed this with a quick test.

jwest115: good catch, I just removed it

bocchino: Thanks! On further consideration, I reverted this change. I think the intent of the or condition is to enable display of enums in text logging even when serializable-to-string is turned off.

compiler/lib/src/main/scala/codegen/CppWriter/ArrayCppWriter.scala
bocchino: Thanks! On further consideration, I reverted this change. Replacing `#if` with `#ifdef` changes the configuration procedure: instead of always defining a symbol to 1 or 0, you have to define or not define a symbol. I'm not sure if this change is what we want. In this PR I think we should be conservative and just address #447. Let's continue to have the discussion about configuration procedure.

bocchino (approved):
Looks good!

compiler/lib/src/main/scala/codegen/CppWriter/EnumCppWriter.scala
bocchino: However, now I see this in the F Prime config code:

https://github.com/nasa/fprime/blob/af76f089e5aaee774d129e9005e19aaeab36e2f0/Fw/Cfg/ConfigCheck.cpp#L39-L42

This suggests that `FW_ENABLE_TEXT_LOGGING != 0` implies `FW_SERIALIZABLE_TO_STRING != 0`. So maybe we should revert the reversion.

PR #455, opened 2024-06-25
Updated XML and GDS limitations #402

### LINKED ISSUE #402: Update XML and GDS limitations in the User's Guide
Some of the issues marked XML and GDS limitations in the User's Guide are or will be fixed. Go through the User's Guide and delete or update them as necessary.

jwest115:
Removed XML and GDS limitations that are no longer relevant and/or fixed.

I removed any limitations that are related to XML to CPP since we no longer use the XML to generate CPP.

Also, for the XML limitation regarding limits for telemetry channels of array or struct types - I verified that the limits for these types are in the XML and JSON dictionary, however, the limits do not appears in the telemetry tab in the GDS. Based on [#2754](https://github.com/nasa/fprime/issues/2754), this appears to be a bug in the GDS, so I think we can change this to a GDS limitation or leave it out... @bocchino let me know what you think!

Closes #402.

### DIFF
--- a/docs/fpp-spec.html
+++ b/docs/fpp-spec.html
@@ -5,7 +5,7 @@
 <meta http-equiv="X-UA-Compatible" content="IE=edge">
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
 <meta name="generator" content="Asciidoctor 2.0.20">
-<title>The F Prime Prime (FPP) Language Specification, Unreleased, after v2.0.2</title>
+<title>The F Prime Prime (FPP) Language Specification, v2.1.0</title>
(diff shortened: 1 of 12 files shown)


timcanham:
We don't want to lose color-coding. That is very important for live testing.

bocchino (approved):
Looks good! I updated the asciidoc, updated the FPP version to v2.1.0, and regenerated the docs.

I reverted one change: we still use the convention that a symbol defined in a component `C` has a name prefixed with `C_`.

thomas-bc (approved):

jwest115:
@timcanham agreed - FPP includes the limit info/colors in the dictionary. My comment above was mainly referring to F´ GDS, since it looks like it does not render the appropriate colors in the channel tab. @thomas-bc created an issue for that https://github.com/nasa/fprime/issues/2754

Issue #456, opened 2024-06-26
CI depends on macOS-11, which is going away


bocchino:
From the CI error report:
```
[Package / build-native-images (macos-11, macosx_10_9_universal2)](https://github.com/nasa/fpp/actions/runs/9682674682/job/26718184337)
This is a scheduled macOS-11 brownout. The macOS-11 environment is deprecated and will be removed on June 28th, 2024.
```

PR #460, opened 2024-06-27
Upgrade to macos-12 in native builds CI

thomas-bc:
Fix https://github.com/nasa/fpp/issues/456

Hopefully this should be all that is needed, but I am not sure of if there could be side effects or incompatibilities.
We may want to wait for @LeStarch to confirm as he is much more familiar with the native build process than me.

### DIFF
--- a/.github/workflows/native-build.yml
+++ b/.github/workflows/native-build.yml
@@ -102,7 +102,7 @@ jobs:
             "run": [
               {
                 "runner": "macos-12",
-                "tag": "macosx_10_9_universal2"
+                "tag": "macosx_12_0_universal2"
               },
               {
                 "runner": "ubuntu-latest",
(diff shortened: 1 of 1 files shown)


bocchino:
Thanks, Thomas! This change looks good to me.

.github/workflows/native-build.yml
```diff
@@ -101,7 +101,7 @@ jobs:
           matrix = {
             "run": [
               {
-                "runner": "macos-11",
+                "runner": "macos-12",
                 "tag": "macosx_10_9_universal2"
```
LeStarch, line 105: Doesn't this tag need to change too?  New image means new build means new osx minimum version....

LeStarch (commented):
One question.

LeStarch (approved):
Fixed the tag for you.  Technically we could automate this step....but not sure it is worth the finesse over a few version points.

bocchino:
Thanks, Michael!

PR #462, opened 2024-07-03
Smbehaviour syntax

garthwatney:
FPP State machine Phase 2.
Completed fpp-syntax and fpp-format and added unit tests


compiler/lib/src/main/scala/ast/Ast.scala
```diff
@@ -587,7 +611,63 @@ object Ast {
 
   }
 
-  /** Port matching specifier */
+  /** Initial state Specifier */
+  final case class SpecInitial(
+    action: Option[Ident],
+    state: Ident
```
bocchino, line 617: State needs to be a qualified identifier, not a simple identifier. Also, in the case of an initial specifier, we revised the spec so that the syntax is `initial [enter-expression]`.

compiler/lib/src/main/scala/ast/Ast.scala
```diff
@@ -587,7 +611,63 @@ object Ast {
...
+  )
+
+  /** Guard definition */
+  final case class DefGuard(
+    name: Ident,
+    typeName: Option[QualIdent]
+  )
+
+  /** Junction definition */
+  final case class DefJunction(
```
bocchino, line 639: This doesn't follow the syntax in the spec, which is described in terms of enter expressions. Let's follow the spec.

bocchino (changes requested):
I added a couple of comments. It looks good, but we should make sure that the syntax implementation matches the syntax described in the spec as closely as possible. Also, we need to use qualified identifiers to refer to states.

bocchino (approved):
Looks good!

PR #463, opened 2024-07-03
fpp-to-json testing refactor

mosa11aei:
This PR implements two changes to fpp-to-json tests:

1. The `patternedConnections` analysis-based test case was added
2. Test files were refactored such that instead of creating a single `.ref.txt` out of the multiple outputs of `fpp-to-json`, each file creates its own `.ref.txt`. This aims to provide more clarity into each reference output, and more granular diff checking if there are any differences between output and reference files.


mosa11aei:
https://github.com/nasa/fpp/issues/464

mosa11aei:
This PR is closing. After further discussion with @bocchino and @garthwatney, the fpp-to-json test structure refactor was decided not to be implemented at the moment.

A new PR will be opened for the patterned cg test case.

PR #469, opened 2024-07-15
fpp-to-json Patterned Connection Graph test case

### LINKED ISSUE #467: fpp-to-json bug with analyzing patterned connections
`fpp-to-json` errors out with a message that there is a missing bitmap entry when trying to analyze files with patterned connections. 

For example, in a file with:

```
module A {
    topology T {
        instance c
        telemetry connections instance c
    }
}
```
Assuming that `c` is properly defined with the patterned interface of the telemetry connections graph, `fpp-to-json` without the `-s` will fail.

This is similar to https://github.com/nasa/fpp/issues/445, however the analysis now causes an error rather than the syntax parsing.

mosa11aei:
This PR implements the test case for the patterned connection graphs. This in turn updates the fpp bitmap to solve the issue posed in #467. 

This PR closes #467.

### DIFF
--- a/compiler/lib/src/main/resources/META-INF/native-image/reflect-config.json
+++ b/compiler/lib/src/main/resources/META-INF/native-image/reflect-config.json
@@ -447,8 +447,52 @@
   "fields":[{"name":"0bitmap$166"}]
 },
 {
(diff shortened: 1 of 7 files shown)


bocchino (approved):
Looks good!

PR #470, opened 2024-07-17
Make component instance names qualified in topology c++ code gen and JSON dictionary

### LINKED ISSUE #349: [C++ code gen, dictionary] Make instance names more consistent
When an instance I is defined in a module M, the name of the instance is M_I. However, the code generation is not entirely consistent:
1. In the dictionary, the name becomes I instead of M_I. This makes the name shorter, but it can create ambiguity because it deletes the module prefix. E.g., M_I and N_I have the same name I in the dictionary.
2. In the generated C++, if a topology T is defined in module M, then the name becomes I when generating C++ code for T. The name should be consistent across topologies. It should not depend on the enclosing module of the topology.

We can fix (1) as we develop the new JSON dictionary. We can fix (2) by updating fpp-to-cpp.

jwest115:
- Updated component instance names so they are fully qualified
- Updated component instance names to use C++ qualification (rather than underscores)
- Refactored C++ code gen to use component instance qualified names rather than short names
   - Change also applied to topology config objects, constants, and helper functions for consistency
- Updated C++ code gen so component instances are defined inside their own namespaces
- Regenerated reference test files and updated fpp test files to match these changes

Closes #349


jwest115:
After talking to @thomas-bc, this change will require an update to `fprime-gds`

thomas-bc (commented):
In the loaders, we parse the qualified names to infer name of component and channel/event/command
Logic will need to be adapted 

https://github.com/nasa/fprime-gds/tree/devel/src/fprime_gds/common/loaders

jwest115:
Created https://github.com/nasa/fprime/issues/2813 to track the GDS update, will work this in before we merge anything on the FPP side.

bocchino (approved):
Looks good!

PR #471, opened 2024-07-18
Attempt to fix native-builds

LeStarch:


LeStarch:
This seems to have gotten past the explosion.  Here is what went wrong:
1. Node is used to "checkout" actions
2. Node is mapped into docker actions for checkout purposes
3. Github unilaterally decided to update to node20 from node16
4. Node20 requires a newer version of GLIBC than node16
5. GLIBC is intentionally old in our native build container to provide more wide-reaching support
6. Node20 exploded in container

The fix, which is limited until they take node16 off their boxes, is to switch to node16 again using environment.

PR #485, opened 2024-07-31
Rename enter exprs to transition exprs

bocchino:
Calling _[_ _do-expr_ _]_ `enter` _qual-ident_ an "enter expression" seems confusing, since it's not just an enter; it's also an optional do. I propose to rename this construct a "transition expression." This renaming works better in the spec and in the algorithm descriptions. E.g., the flattening algorithm maps flattened transitions to these constructs.

This change affects the parser and AST, but it's a simple renaming change.


bocchino:
Merging this. The change is underway in #486.

PR #487, opened 2024-08-01
Phase1 state machine instance additions spec update

garthwatney:
Add priority and queue behavior to state machine instance specifiers


bocchino (approved):
Looks good!

PR #489, opened 2024-08-02
Phase1 state machine instance additions semantics update

garthwatney:
Add priority and queue behavior semantics checking for state machine instance priority and queue full behavior


bocchino (approved):
Looks good!

PR #493, opened 2024-08-12
Phase 1 state machines

garthwatney:
This pull request adds the Phase 1 of state machines to the FPP language.  In phase 1 the state machine behavior is specified outside FPP but the state machine definition names and instance names are specified in FPP and the appropriate support code is generated in the component base class.  There are also some incidental cleanup like the key words (ask Rob)


bocchino:
We held an in-person review on 8/20/24. At that review we agreed on the following changes:
* There should be one signal-sending function for each state machine instance, instead of one function for all instances.
* The signal-sending function should implicitly fill in the state machine instance ID and should have two arguments: a signal ID enumeration ID and SignalBuffer, which is a stack-allocated buffer. The SignalBuffer should just contain data bytes.
* The size of the data stored in the SignalBuffer should not be hard-coded. It should be configurable as part of the F Prime configuration.
* The pure virtual functions called when doing an action should have three arguments: state machine instance ID, signal ID, and signal buffer.

garthwatney:
All issues from the 8/20 review should be fixed.

bocchino:
Great! I confirmed that the `check-cpp` tests pass.

As discussed, let's exercise a few more capabilities in the state machines in the unit tests.

compiler/tools/fpp-to-cpp/test/fprime/Fw/Types/Types.fpp
```diff
@@ -13,6 +13,12 @@ module Fw {
     NO_ROOM_LEFT  @< No room left in the buffer to serialize data
   }
 
+  struct SMSignals {
```
bocchino, line 16: Do we still need this type?

compiler/tools/fpp-to-cpp/test/fprime/Fw/Types/Types.fpp
```diff
@@ -71,4 +77,10 @@ module Fw {
       FAILURE @< Representing failure
       SUCCESS @< Representing success
   }
+
```
bocchino, line 80: Why is this type being added?

bocchino:
> I confirmed that the check-cpp tests pass.

Scratch that, I forgot to pull the latest. It looks like the STARS-generated code needs to be updated in this repo in order for the `fpp-check` compilation to work.

bocchino:
After the updates, I ran fpp-check against this F Prime branch: https://github.com/garthwatney/fprime/commit/08d8135b341daf5531a0ed1c1c06bad6bfe636f5. I'm seeing this compilation error in fpp-to-cpp/test/component/impl:
```
In file included from ActiveStateMachines.cpp:7:
./ActiveStateMachines.hpp:40:17: error: use of undeclared identifier 'S2_Interface'
          const S2_Interface::S2_Signals signal, //!< The state machine signal
                ^
fatal error: too many errors emitted, stopping now [-ferror-limit=]
```

compiler/tools/fpp-to-cpp/test/fprime/Fw/Types/Types.fpp
bocchino: This type is coming in from F Prime.

compiler/tools/fpp-to-cpp/test/component/ActiveStateMachines_S1.cpp
```diff
@@ -0,0 +1,103 @@
...
+// ======================================================================
+// \title  ActiveStateMachines_S1.cpp
+// \author Auto-generated
+// \brief  cpp file for state machine ActiveStateMachines_S1
+//
+// ======================================================================            
+    
+#include "stdio.h"
+#include "assert.h"
+#include "Fw/Types/SMSignalsSerializableAc.hpp"
```
bocchino, line 11: We need to remove this header include. `SMSignals` is no longer used.

compiler/tools/fpp-to-cpp/test/component/ActiveStateMachines_S1.hpp
```diff
@@ -0,0 +1,96 @@
...
+// \title  ActiveStateMachines_S1.h
+// \author Auto-generated
+// \brief  header file for state machine ActiveStateMachines_S1
+//
+// ======================================================================
+           
+#ifndef ACTIVESTATEMACHINES_S1_H_
+#define ACTIVESTATEMACHINES_S1_H_
+                                
+#include <Fw/SMSignal/SMSignalBuffer.hpp>
```
bocchino, line 12: For consistency with the rest of F Prime, I think the path should be `Fw/Sm/SmSignalBuffer.hpp`. Compare `Fw/Com/ComPacket.hpp`, `Fw/Dp/DpContainer.hpp`, etc.

compiler/tools/fpp-to-cpp/test/fprime/Fw/Types/Types.fpp
bocchino: I removed this type and fixed a related bug in the code gen (a use of the type was still hanging around and needed to be replaced).

compiler/tools/fpp-to-cpp/test/component/ActiveStateMachines_S1.cpp
bocchino: I removed this.

bocchino: I fixed this. I also removed the dependencies on stdio and assert. These are not allowed in F Prime flight code. @garthwatney Please update the STARS code generator to match.

bocchino:
`fpp-check` is now passing.

compiler/tools/fpp-to-cpp/test/component/ActiveStateMachines_S1.cpp
```diff
@@ -0,0 +1,100 @@
+
+// ======================================================================
+// \title  ActiveStateMachines_S1.cpp
+// \author Auto-generated
+// \brief  cpp file for state machine ActiveStateMachines_S1
+//
+// ======================================================================            
+
+#include "ActiveStateMachines_S1.hpp"
```
bocchino, line 9: @garthwatney Can you make sure that STARS generates these header includes? Before I saw dependencies on C stdio and C assert. We shouldn't have those dependencies in flight code.

compiler/tools/fpp-to-cpp/test/component/ActiveStateMachines_S1.cpp
```diff
@@ -0,0 +1,100 @@
...
+            
+            switch (signal) {
+
+                default:
+                    break;
+            }
+            break;
+    
+        default:
+            FW_ASSERT(0);
```
bocchino, line 98: @garthwatney Can you make sure that STARS generates FW_ASSERT here? Before I saw a C-style assert, which we should not use in F Prime flight code.

compiler/tools/fpp-to-cpp/test/component/ActiveStateMachines_S1.hpp
garthwatney: Done

compiler/tools/fpp-to-cpp/test/component/ActiveStateMachines_S1.cpp
garthwatney: Done

compiler/tools/fpp-to-cpp/test/component/ActiveStateMachines_S1.cpp
garthwatney: Done

compiler/tools/fpp-to-cpp/test/component/ActiveStateMachines_S1.hpp
bocchino: Looks good!

bocchino (approved):
Looks good!

LeStarch (approved):

PR #494, opened 2024-08-14
Add Displayable Types

jwest115:


bocchino (approved):
Looks good! I made some revisions to the User's Guide. I added some semantic tests for displayable types, and I added some more error reporting for tracing through type definitions for non-displayable types.

PR #495, opened 2024-08-16
Fixing missing parameters in overflow hook calls

LeStarch:
Fixes for overflow hook call.  In short, the `portNum`  default argument was missing.


bocchino (approved):
Looks good. I made some additional fixes to get it to compile.

1. Internal ports were still broken. Fixed now.
2. To call the overflow hook for async commands, we have to deserialize the command arguments. The deserialization code is missing. For now I commented out the hook call.

I will open an issue on item 2. I want to defer work on this issue until #493 is merged. It is very difficult to simultaneously develop the overflow hook changes and the state machine changes to the same code. In the mean time, all the overflow behavior except for commands should be good to test in FppTest.

PR #497, opened 2024-08-22
Fix code gen for command overflow hooks

### LINKED ISSUE #496: Fix code generation for command overflow hooks
We need to deserialize the command arguments before calling the hook. See #495.

This will require some refactoring of the code gen, as we will need to do the same thing as in the code that calls the command handlers.

This issue is blocked on #493 to avoid conflicts.

bocchino:
* In the function prototypes and at the call sites, follow the pattern of the pre-message hooks. Don't include the command arguments. At the call sites, these are in serialized form and so are not available in typed argument form.
* Re-enable the call sites, which were commented out to get the code to compile.
* Update the unit tests.
* Update the version stamp to after v2.1.0.
* Re-generate the docs.

Closes #496.

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCommands.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCommands.scala
@@ -192,7 +192,7 @@ case class ComponentCommands (
                   priority,
                   MessageType.Command,
                   cmd.getName,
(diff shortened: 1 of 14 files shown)


thomas-bc (approved):

LeStarch (approved):

LeStarch:
@bocchino once merged, ready for a alpha?

bocchino:
Yes, I will merge it and then make an alpha release.

PR #498, opened 2024-08-27
Fix a glitch in CI

bocchino:
Scala 3.5 has broken the use of `scala` on the command line to run compiled Scala classes. `scala` now invokes the Scala CLI, which has different behavior. See https://www.scala-lang.org/blog/2024/08/22/scala-3.5.0-released.html. The new way to run compiled Scala classes, if any, is not documented online. Scala 3.5.0 is so new that it is not yet available in MacPorts.

For now, detect that Scala 3.5 is installed and disable the test in this case. This is OK for CI, because the checks done by this test are redundant with the checks done by the fpp-to-cpp tests. When the Scala docs and the distributions stabilize, we'll work out a better fix.


bocchino:
Scala 3.5 seems to have broken the use of `scala` on the command line to run compiled Scala classes. See https://www.scala-lang.org/blog/2024/08/22/scala-3.5.0-released.html.

LeStarch (approved):

PR #512, opened 2024-09-16
Allow unmatched connections to matched ports

### LINKED ISSUE #249: Improve support for matching connections through a hub
The matched port checker does not play well with the hub pattern. When it sees a pair (p1, p2) of matched ports, it expects each connection C with one end at p1 to have a unique instance I(C) at the other end, and it looks for a matching connection between p2 and I(C). This works fine for matched connections in a single physical deployment, but when matched connections cross a hub boundary everything is connected to a single hub instance, so the pattern doesn't work.

We discussed two possible fixes:
1. We relax the matching algorithm so it does not attempt to enforce matching if it sees a manually numbered connection with a serial port at the other end. 
1. We extend the matching algorithm so that two connections are matched if they go to the same component with the same port indices.
Solution 2 seems cleaner, but it requires a change to the way that the hub is currently used (currently, the pairs of matched connections typically do not have matching port indices at the hub).

### LINKED ISSUE #525: spec  on Component Graph Specifiers missing "text event" pattern-kind
`Ref/Top/topology.fpp` in fprime has `text event connections instance textLogger` but the `text event` part seems to be missing from the spec Section 7.3.

jwest115:
- Added `unmatched` keyword
- Updated parser to catch `unmatched` specifiers
- Updated connection AST and semantics to include whether the connection is unmatched. Added error handling for the case where unmatched is specified on connections that do not have matched ports.
- Updated automatic port numbering algorithm to consider unmatched connections
- Added unit tests
- Updated spec and user guide to include match constrained and unmatched concepts

Closes #249. Closes #525.


compiler/tools/fpp-to-xml/test/top_numbering/unmatched/unmatched_connections.fpp
```diff
@@ -0,0 +1,59 @@
...
+
+    instance c1
+    instance c2
+    instance c3
+    instance c4
+
+    connections P {
+      # Case 1: 2 ports that go to the same component
+      unmatched c1.pOut -> c1.pIn
+      unmatched c2.pOut -> c1.pIn
```
jwest115, line 36: should be c2.In

compiler/lib/src/main/scala/util/Error.scala
```diff
@@ -211,6 +215,10 @@ sealed trait Error {
         Error.print (Some(loc)) ("passive component may not have async input")
       case SemanticError.PassiveStateMachine(loc) =>
         Error.print (Some(loc)) ("passive component may not have a state machine instance")
+      case SemanticError.PortNumberAlreadyInUse(n) =>
+        Error.print (None) (s"port number $n is already in use")
+      case SemanticError.PortNumberOutOfRange(portNumber, port, arraySize) =>
```
bocchino, line 220: `PortNumberOutOfRange` appears to be redundant with `InvalidPortNumber`. Can we just use `InvalidPortNumber`?

compiler/lib/src/main/scala/analysis/Semantics/ResolveTopology/MatchedPortNumbering.scala
```diff
@@ -161,29 +187,52 @@ object MatchedPortNumbering {
...
+    // Computes the set of used ports for all connections at a specific port instance
+    def computeUsedPortNumbers(pi: PortInstance): Result.Result[Set[Int]] = {
+      val empty: Set[Int] = Set()
+      val pii = PortInstanceIdentifier(ci, pi)
+      val cs = t.getConnectionsAt(pii).toList.sorted
+      Result.foldLeft (cs) (empty) ((s, c) => {
+        val used = t.getUsedPortNumbers(pi, List(c))
+        val intersection = s.intersect(used)
+        // Do not allow duplicate port numbers
+        if intersection.nonEmpty
```
bocchino, line 218: Is it possible for this intersection to contain more than one element?

compiler/lib/src/main/scala/util/Error.scala
```diff
@@ -195,8 +195,12 @@ sealed trait Error {
       case SemanticError.MissingConnection(loc, matchingLoc) =>
         Error.print (Some(loc)) ("no match for this connection")
         printMatchingLoc(matchingLoc)
+      case SemanticError.MissingPortMatching(loc) =>
+        Error.print (Some(loc)) ("unmatched connection must go from or to a matched port")
       case SemanticError.MissingPort(loc, specMsg, portMsg) =>
         Error.print (Some(loc)) (s"component with $specMsg must have $portMsg")
+      case SemanticError.MultipleConnectionsAtPortIndex(portIndexString) =>
+        Error.print (None) (s"multiple connections appear at port indices $portIndexString")
```
bocchino, line 203: I think this error message should not refer to indices (plural). The common case (and maybe the only case?) is a single index.

compiler/lib/src/main/scala/analysis/Semantics/ResolveTopology/MatchedPortNumbering.scala
jwest115: I think a case where the intersection would contain more than one element is when a connections is using 2 or more port numbers that are already in the accumulator set. In this case I use `nonEmpty` to check if the intersection contains at least one element. If it does, then that means that there is a duplicate (multiple connections are using the same port number) and we throw and error.

compiler/lib/src/main/scala/util/Error.scala
jwest115: True! I updated the wording in the error to be "index" vs. "indices".

compiler/lib/src/main/scala/analysis/Semantics/ResolveTopology/MatchedPortNumbering.scala
bocchino: I don't see how a connection can use 2 port numbers. A connection has the form `a.b[i] -> c.d[j]`, where `i` and `j` are numbers.

bocchino: We are thinking of extending connections to allow ranges, e.g., a.b[0..5] -> c.d[0..5], but we don't allow that yet.

jwest115: I was thinking of the case where i and j are different numbers, so the connection would be using 2 port numbers. Does that make sense?

jwest115: Oh, actually never mind. Only one port number would be used since we are dealing with one port instance at a time.  Sorry for the confusion and to clear this up, the intersection would contain only one element!

compiler/lib/src/main/scala/util/Error.scala
```diff
@@ -195,8 +195,12 @@ sealed trait Error {
       case SemanticError.MissingConnection(loc, matchingLoc) =>
         Error.print (Some(loc)) ("no match for this connection")
         printMatchingLoc(matchingLoc)
+      case SemanticError.MissingPortMatching(loc) =>
+        Error.print (Some(loc)) ("unmatched connection must go from or to a matched port")
       case SemanticError.MissingPort(loc, specMsg, portMsg) =>
         Error.print (Some(loc)) (s"component with $specMsg must have $portMsg")
+      case SemanticError.MultipleConnectionsAtPortIndex(portIndexString) =>
+        Error.print (None) (s"multiple connections appear at port index $portIndexString")
```
jwest115, line 203: Re-word error, something along the lines of "duplicate connection at matched port"

compiler/lib/src/main/scala/analysis/Semantics/ResolveTopology/MatchedPortNumbering.scala
```diff
@@ -161,29 +187,52 @@ object MatchedPortNumbering {
...
       })
     }
+
+    // Computes the set of used ports for all connections at a specific port instance
+    def computeUsedPortNumbers(pi: PortInstance): Result.Result[Set[Int]] = {
+      val empty: Set[Int] = Set()
+      val pii = PortInstanceIdentifier(ci, pi)
+      val cs = t.getConnectionsAt(pii).toList.sorted
+      Result.foldLeft (cs) (empty) ((s, c) => {
+        val used = t.getUsedPortNumbers(pi, List(c))
```
jwest115, line 215: Folding over all the ports that are connected to a specific matched instance.

compiler/lib/src/main/scala/util/Error.scala
```diff
@@ -211,6 +215,8 @@ sealed trait Error {
         Error.print (Some(loc)) ("passive component may not have async input")
       case SemanticError.PassiveStateMachine(loc) =>
         Error.print (Some(loc)) ("passive component may not have a state machine instance")
+      case SemanticError.PortNumberAlreadyInUse(n) =>
+        Error.print (None) (s"port number $n is already in use")
```
jwest115, line 219: need to call out which port number is causing the issue and where it was already used in the other connection

compiler/lib/src/main/scala/util/Error.scala
jwest115: Update: we want a new error along the lines of "matched port numbering could not find an available port number on both sides of the matching"

compiler/lib/src/main/scala/analysis/Semantics/ResolveTopology/MatchedPortNumbering.scala
jwest115: Update: refactor to do something similar to `checkDuplicateOutputConnections`

bocchino (approved):
Looks great! Everything was functionally correct. I made some revisions for style, clarity, and efficiency.

PR #513, opened 2024-09-24
Refactors AC queue codegen to be inline with OSAL queue refactor

LeStarch:
This brings codegen inline with new OSAL refactors.


bocchino:
The corresponding PR to nasa/fprime is https://github.com/nasa/fprime/pull/2895.

bocchino (approved):
Looks good! I confirmed that the fpp-check tests pass.

PR #517, opened 2024-10-02
FP-2784: Add trace id to fpp

Shivaly-Reddy:
Adding trace id type to fpp to enable Tracing implementation


bocchino (approved):
Looks good!

PR #521, opened 2024-10-11
Revise code gen for DpGet

### LINKED ISSUE #520: DpGet does not clear out the DpContainer that's passed in
The DpGet interface requires the user to pass in a reference to a DpContainer. Currently, if that DpContainer has been used in a previous DpGet call, then some of the state from the previous call will remain in the container after the second DpGet call. That shouldn't happen. To support the reuse of DpContainer objects across different calls, this code should fully clear out the DpContainer:

https://github.com/nasa/fpp/blob/e4e137f3cf48350e2450e1a988f684597fddec6d/compiler/tools/fpp-to-cpp/test/component/base/ActiveGetProductsComponentAc.ref.cpp#L3430-L3441

This issue is related to https://github.com/nasa/fprime/issues/2938.

bocchino:
* Closes #520.
* Updates fpp-check scripts to clean up code, and to conform to changes in F Prime.
* Fixes issues with CI.

### DIFF
--- a/.github/workflows/native-build.yml
+++ b/.github/workflows/native-build.yml
@@ -53,7 +53,7 @@ jobs:
     name: "Build Standard JARs"
     steps:
       - name: "Checkout repository"
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           submodules: recursive
       - name: "Setup Native Image Tools"
@@ -64,7 +64,7 @@ jobs:
           ${{ inputs.build }} ${{ inputs.output-directory }}
         shell: bash
       - name: "Archiving JARs package"
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: build-jar
           path: ${{ inputs.output-directory }}/*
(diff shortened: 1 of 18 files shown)


bocchino:
@LeStarch CI is erroring out with `sbt: not found`. Can you investigate?

https://github.com/nasa/fpp/actions/runs/11286265572/job/31390351156?pr=521

.github/workflows/native-build.yml
```diff
@@ -131,27 +131,20 @@ jobs:
     container:
       image:  ${{ matrix.run.container || '' }}
     steps:
-      - name: "Fix environment"
-        run: |
-          # https://github.com/actions/checkout/issues/1809
-          # https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/
-          echo "ACTIONS_RUNNER_FORCED_INTERNAL_NODE_VERSION=node16" >> $GITHUB_ENV
-          echo "ACTIONS_RUNNER_FORCE_ACTIONS_NODE_VERSION=node16" >> $GITHUB_ENV
```
thomas-bc, line 139: @LeStarch these links seemed specific to older versions of the `actions/checkout` et al.

Please confirm this is ok to remove, if it passes CI

thomas-bc: nvm it doesn't work, I had to revert
See note in https://github.com/nasa/fpp/pull/521/commits/1407109cc8125315dfff66f8925800fa2bae9fde

LeStarch (approved):

PR #532, opened 2024-11-12
Add component state machine getter

### LINKED ISSUE #530: [Phase 1 state machines] Add a protected function for getting the state
In Phase 1 state machines, the state machine instance variables are private in the auto-generated component base class. This is so that the implementation can't access the state machine instance directly; it has to go through the interface provided by the component base class. However, the implementation needs to be able to query the state of the state machine. So we need to add a protected function for getting the state to the component interface.

- [ ] Revise code gen
- [ ] Update unit tests

Kronos3:
Adds a `protected` `*_getState` function for each state machine.

Closes #530

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentStateMachines.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentStateMachines.scala
@@ -205,11 +205,12 @@ case class ComponentStateMachines(
 
       val smiName = smi.getName
       val smName = s.writeSymbol(smi.symbol)
+      val smEnumName = s"$smName::${s.getName(smi.symbol)}_States";
       functionClassMember( 
         Some(s"Get the state of state machine instance $smiName"), 
         s"${smiName}_getState",
         Nil, 
-        CppDoc.Type(s"$smName::State", Some(s"$name::$smName::State")), 
+        CppDoc.Type(smEnumName, Some(smEnumName)), 
         lines(s"return this->m_stateMachine_$smiName.state;"), 
         CppDoc.Function.NonSV, 
         CppDoc.Function.Const 
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentStateMachines.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentStateMachines.scala
@@ -26,8 +26,8 @@ case class ComponentStateMachines(
   def getFunctionMembers: List[CppDoc.Class.Member] = {
     List.concat(
       getOverflowHooks,
-      getSignalSendMember,
-      getStateGetterFunction
+      getSignalSendFunctions,
+      getStateGetterFunctions
     )
   }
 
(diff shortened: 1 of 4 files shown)


bocchino (changes requested):
Looks good!

I forgot to mention that you can run `./check-cpp` in `fpp-to-cpp/test` or `fpp-to-cpp/test/component/base` to check that the generated code compiles. When I did that I saw that the state type in the generated state machine class is not `State` but `ActiveStateMachines_S1_States`. Can you fix it and run `./check-cpp` to make sure it compiles? Thanks!

bocchino:
The externally generated code is here: `fpp-to-cpp/test/component/ActiveStateMachines_S1.hpp`.

Kronos3:
Looks like this requires the `FPRIME` environment variable to be set. I tried this with the latest `devel` and `v3.5.0` of https://github.com/nasa/fprime which doesn't seem to work properly:

```
$ ../../../../../scripts/fprime-gcc -c ActiveAsyncProductPortsOnlyComponentAc.cpp -I../../../.. -I.. -I../../fprime -I../../fprime/config -Wno-sign-conversion -Wno-unused-parameter -Wno-vla-extension -Wno-zero-length-array
In file included from ActiveAsyncProductPortsOnlyComponentAc.cpp:12:
In file included from ../base/ActiveAsyncProductPortsOnlyComponentAc.hpp:12:
../../fprime/Fw/Dp/DpRequestPortAc.hpp:19:10: fatal error: 'config/FwDpIdType.hpp' file not found
   19 | #include "config/FwDpIdType.hpp"
      |          ^~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

When I use `FPRIME=.../compiler/tools/fpp-to-cpp/test/fprime` looks like I'm missing some configuration files. I'm betting I'm missing some untracked git files that you have in your working environment but I'm not sure how to set things up properly.

bocchino:
I think the problem is that `fpp-check` is using older FPP version that is in your PATH, instead of the one that's installed in the bin directory of your working repo. This patch should fix it:
```
diff --git a/compiler/tools/fpp-to-cpp/test/fprime/generate_cpp b/compiler/tools/fpp-to-cpp/test/fprime/generate_cpp
index d33ebbdc..ebd8417d 100755
--- a/compiler/tools/fpp-to-cpp/test/fprime/generate_cpp
+++ b/compiler/tools/fpp-to-cpp/test/fprime/generate_cpp
@@ -9,7 +9,7 @@ cd `dirname $0`
 echo "generating framework C++"
 
 fpp_files=`find . -name '*.fpp'`
-fpp-to-cpp -p $PWD $fpp_files
+../../../../bin/fpp-to-cpp -p $PWD $fpp_files
 
 # Move config files into place
 for base in FppConstantsAc ProcTypeEnumAc
```
Can you try applying that?

The FPRIME variable should be set to a local copy of nasa/fprime/devel.

Kronos3:
Ah yes good call, I had residual packages (including `fprime-fpp`) installed globally from an older version of Python3. I cleaned those up and pointed `FPRIME` to the `devel` checkout and it's working as expected now.

I just pushed a fix for the above.

bocchino (approved):
Looks good!

I updated the path to fpp-to-cpp in the generate script. That way the script will pick up fpp-to-cpp from the repo instead of the environment.

PR #534, opened 2024-11-15
State Machines Phase 2

bocchino:
This pull request integrates Phase 2 state machines from the feature branch to main.

There are a lot of changes here, so I'll schedule one or more meetings to go over them.


LeStarch:
- Review diagrams with Garth!
- General.md vs README.md
- Add UML diagram to represent base class/component relationship
- Style guide note: avoid name of objects with `_`
- Fix the hard to read names of the form m(italic) some token (greyed, no formatting)
- Document the rules with initial states:
    - Cannot perform port calls
    - Cannot emit events nor telemetry
- Move to “start” phase? (Post connectComponents)
- Add a comments to break up flattened transition codes
    - Before doing each thing
- What was the message buffer change?
    - To be reviewed
- Coverage reports for UTs

Kronos3 (approved):
I skimmed through the generated code and it looks pretty good to me. One minor note is that it may be better in the future to build in the implementation of the state machine into the component class itself (i.e. dump the state machine logic into the owner class)

At the moment we are calling into the state machine from the component which in turns calls out actions back to the component. Another option might be to put the state machine action implementations in the `hpp` so that they may be inlined.

bocchino:
Thanks for your comments.

> it may be better in the future to build in the implementation of the state machine into the component class itself

Do you mean replicate the state machine base class in every component where the state machine is instantiated? Wouldn't that cause code duplication if the same state machine is used in more than one component? Or do you mean eliminate the inner class that implements the state machine? We know a way to do that, but it uses multiple inheritance and to me seems less clean than the solution we have adopted. It also has slightly worse type safety.

> Another option might be to put the state machine action implementations in the hpp so that they may be inlined.

I think the action implementations can be inlined. There are two levels of action implementation (in the inner class and in the component class) but they are both defined in the same auto-generated cpp file, so the compiler can see them both.

PR #544, opened 2024-12-04
Add unit tests for state machine semantics

### LINKED ISSUE #541: Add tests for type and constant uses in state machine definitions
We should add a few tests to verify that type and constant uses are being picked up when they occur inside state machine definitions.

- [x] Revise tests for `fpp-depend`
- [x] Revise tests for `fpp-locate-uses`

bocchino:
* Add unit tests for `fpp-depend` and `fpp-locate-uses`.
* Fix bug in output regeneration for `fpp-depend`.

Closes #541.


compiler/tools/fpp-depend/test/def_state_machine.fpp
```diff
@@ -0,0 +1,11 @@
+locate type A at "A.fpp"
+locate type G at "G.fpp"
+locate constant s at "s.fpp"
+
+state machine M {
+
+  action a: A
+  guard g: G
+  signal s: string size s
```
Kronos3, line 9: Does it make sense to add warnings about unused definitions inside state machines?

Kronos3 (approved):
Looks good, just one minor comment

compiler/tools/fpp-depend/test/def_state_machine.fpp
bocchino: Yes, that makes sense. Right now we don't have any "warning report" capability for FPP, we just have detect-error-and-halt. We would need to design an interface to the warnings, e.g., can we turn them on and off, do they just display to the console, can they be promoted to errors?

PR #577, opened 2025-01-17
Require displayable types in records

### LINKED ISSUE #576: Require displayable types in records
Currently an internal error is thrown if an abstract type is used in a record "Exception in thread "main" internal error: type not supported in JSON dictionary spec". We can fix this by not allowing abstract types to be present in records:

- Update spec
- Update record semantics to include non-displayable type check

jwest115:
Updated spec, user guide, and record semantics so non-displayable types are not allowed.

Closes #576.


bocchino:
@jwest115 Could you update this PR to merge into this branch?

https://github.com/nasa/fpp/tree/release/v2.3.0

Thanks!

bocchino:
Actually, I think we need to close this PR and make a new one against the release branch with just these changes. I think we need to do that in order to get these changes into the next F Prime release.

jwest115:
Closing this PR since a new PR was created based on the release branch https://github.com/nasa/fpp/pull/580.

PR #580, opened 2025-01-23
Require displayable types in records

### LINKED ISSUE #576: Require displayable types in records

jwest115:
Updated spec, user guide, and record semantics so non-displayable types are not allowed.

Closes #576.


bocchino (approved):
Looks good! There was one broken link in the User's Guide, which I fixed. The output of `redo` tells you if there are any broken links.

PR #549, opened 2024-12-09
Type aliases

Kronos3:
Closes #113 

- [ ] [**In progress**] Revise the spec
- [ ] Implement and test syntax, including AstWriter and FppWriter
- [ ] Update the algorithm descriptions on the wiki
- [ ] [**Mostly done**] Implement and test semantics
- [ ] [**Partially done**]  Implement and test code gen
- [ ] Update User's Guide

Issue #113, opened 2022-04-01
Type aliases


bocchino:
Allow type aliases, e.g., `type FwPacketDescriptorType = U32`. That way we can generate code for the GDS and for the C++ implementation that keeps these types consistent.

**Note:** This issue covers the FPP language feature of type aliases. Improving the GDS configuration and checking type bounds should be separate issues.

bocchino:
We can also check type bounds as discussed in #198.

PR #591, opened 2025-02-03
Type alias syntax & semantics

Kronos3:
Part of #113. This finishes up some changes made in #549 for syntax & semantics implementation. It adds a test to generate a struct that references a type alias.

No C++ code is generated yet for type aliases. In model elements that use type aliases, the type aliases are converted to their underlying types before generating C++ code.


bocchino:
It looks like you need to regenerate the trace info for the native build. See the instructions here: https://github.com/nasa/fpp/tree/main/compiler#building-native-binaries.

bocchino:
Another issue I see is that the code generator is generating an empty TypesAc.cpp file. I think we should suppress the generation of this file. Does that require any changes to the CppDoc framework?

Kronos3:
@bocchino You may need to do the trace updates again, I can't install GraalVM for Java11 since the only version available from Homebrew is jdk23 and the [SDKMAN/official installation method](https://www.graalvm.org/downloads/#) seems to only have 17, 21, and 23.

We may also want to consider updating the Java version or putting the trace updates in CI

compiler/lib/src/main/scala/analysis/Semantics/Type.scala
```diff
@@ -345,8 +365,11 @@ object Type {
   }
 
   /** Check for type identity */
-  def areIdentical(t1: Type, t2: Type): Boolean = {
+  def areIdentical(t1Aliased: Type, t2Aliased: Type): Boolean = {
```
bocchino, line 368: This change doesn't conform to the spec. The spec says that an alias type is identical to itself, not to its underlying type. See here: https://nasa.github.io/fpp/fpp-spec.html#Type-Checking_Identical-Types.

compiler/lib/src/main/scala/analysis/Analyzers/TypeExpressionAnalyzer.scala
```diff
@@ -55,6 +55,14 @@ trait TypeExpressionAnalyzer
     opt(typeNameNode)(a, node.data.typeName)
   }
 
+  override def defAliasTypeAnnotatedNode(a: Analysis, node: Ast.Annotated[AstNode[Ast.DefAliasType]]) = {
+    val (_, node1, _) = node
+    val data = node1.data
+    for {
```
bocchino, line 61: This for...yield can be simplified to a call to typeNameNode.

compiler/lib/src/main/scala/codegen/CppWriter/CppWriterState.scala
```diff
@@ -136,7 +136,9 @@ case class CppWriterState(
       for {
         fileName <- sym match {
           case _: Symbol.AbsType =>
-            if isBuiltInType(name) then None else Some(name)
+              if isBuiltInType(name) then None else Some(name)
+          // TODO(tumbar) Compute the includes for alias types
```
bocchino, line 140: Can we resolve this TODO before merging this PR? I think we just need to and add `TypeAc` to the filename pattern.

bocchino: Or is the issue that we're not generating code for alias types, so we can't generate the include header? If so, can the comment be clearer?

compiler/lib/src/main/scala/codegen/CppWriter/TypeCppWriter.scala
```diff
@@ -16,6 +17,10 @@ case class TypeCppWriter(
     override def absType(s: CppWriterState, t: Type.AbsType) =
       s.writeSymbol(Symbol.AbsType(t.node))
 
+    override def aliasType(s: CppWriterState, t: Type.AliasType) =
+      // TODO(tumbar) Keep the aliased type name in codegen
```
bocchino, line 21: Please clarify that we're doing it this way for now because we're not generating the type alias code.

compiler/lib/src/main/scala/util/Version.scala
```diff
@@ -3,6 +3,6 @@ package fpp.compiler.util
 /** The compiler version */
 object Version {
 
-  val v = "[unknown version]"
```
bocchino, line 6: Please revert this change. This is a glitch in the way the installer works -- if you abort the installation, it can corrupt this file.

compiler/tools/fpp-to-cpp/test/top/check-cpp-dir/Basic/check
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh -e
 
 fpp_to_cpp=../../../../../../bin/fpp-to-cpp
+fprime_gcc=../../../../../../scripts/fprime-gcc
```
bocchino, line 4: Good catch!

bocchino (commented):
Looks good, I added some comments.

We also discussed adding a few tests for type aliases to the ScalaTest tests. Can we do that in this PR?

compiler/lib/src/main/scala/codegen/CppWriter/CppWriterState.scala
Kronos3: Ya we are not actually generating this header so we can't add it to the include yet. But I did realize that this still isn't right since the underlying type could still be a non-primitive which would mean we do need an include for the underlying type, it's updated now.

bocchino:
I opened an issue on upgrading Java: https://github.com/nasa/fpp/issues/605.

bocchino:
As discussed, I'll resolve the conflict in the native image trace file.

bocchino:
I resolved the conflict. It looks like there is a test failure for `fpp-to-cpp/test/struct/alias_type.fpp`. A map lookup is failing.

Kronos3:
I added some scala tests and found an issue in the `commonType()` implementation for type aliases. We now compute the "lowest common ancestor" of the type alias tree as the common type.

For example:
```
type T

type A1 = T
type A2 = A1
type A3 = A1
```

`commonType(A2, A3) => A1` not the underlying type `T` since A1 is more descriptive. I also had to update the spec to include this change

bocchino (approved):
Looks good!

Issue #605, opened 2025-02-12
Upgrade Java and GraalVM


bocchino:
In CI, our GraalVM build of FPP native binaries is based on Java 11. The Java 11 version is no longer supported on the GraalVM web site:

https://www.graalvm.org/latest/docs/

It seems we should move to Java 17 or later. We should assess what is needed for this change, e.g., testing, assessing impact on users.

**See update below:** It looks like the Graal VM community releases, which we use, have moved on from Java 11. However, the release that we are using is still available here: https://github.com/graalvm/graalvm-ce-builds/releases/tag/vm-22.3.3.

LeStarch:
Several notes: 

1. Java 17 is not available to us as it requires a paid license through Oracle.  We will need to switch to GraalVM with Java 21 or 23
2. The critical test is if graal runs on our aarch64 hardware in CI.  I will push a fix to test that.

LeStarch:
Also: F Prime development should use the community edition: https://github.com/graalvm/graalvm-ce-builds/releases?page=1, which maintains all historical releases.

LeStarch:
Correct.  I am closing this issue.  We can track another issue to eventually update and reconfigure `sbt` in order to fix the use of a removed flag. `CMSClassUnloadingEnabled`.

PR #606, opened 2025-02-13
Bumping Graal VM Version to Latest (JDK 23)

LeStarch:


LeStarch:
Close this PR when CI passes again.  This upgrade is not possible given the current `sbt` setup.

bocchino:
What is the sbt error? Is it complaining about the garbage collector settings? Can we remove those settings and try again? Note: I think this is low priority in light of #605.

LeStarch:
`sbt` is using a flag `CMSClassUnloadingEnabled` on the java command line that is not supported past Java17.

bocchino:
It looks like that setting is particular to an older garbage collector that is no longer supported in Java 17. See https://stackoverflow.com/questions/70214788/unrecognized-vm-option-cmsclassunloadingenabled-after-switching-to-java-17. Going forward we can remove `-XX:+CMSClassUnloadingEnabled` and add `-XX:+UseG1GC`. Apparently G1GC is the default garbage collector in Java 17+. This is probably a better garbage collector and it will also smooth the path to upgrading Java. I'll open an issue on this.

PR #607, opened 2025-02-13
Add telemetry packets to JSON dictionary

### LINKED ISSUE #574: Implement and test dictionary generation for telemetry packets
- [x] Implement
- [x] Test

jwest115:
Added telemetry packets and telemetry packet sets to the JSON dictionary.
Updated JSON dictionary so dictionary entries are sorted before output to JSON.

Closes #574.


compiler/lib/src/main/scala/codegen/DictionaryJsonWriter/DictionaryJsonEncoder.scala
```diff
@@ -30,8 +30,12 @@ case class DictionaryJsonEncoder(
     dictionaryState: DictionaryJsonEncoderState
 ) {
     /** Dictionary entry map to JSON */
-    private def dictionaryEntryMapAsJson[A, B] (f1: (A, B) => Json) (map: Map[A, B]): Json =
-        (map.map { case (key, value) => f1(key, value) }).toList.asJson
+    private def dictionaryEntryMapAsJson[A <: BigInt, B] (f1: (A, B) => Json) (map: Map[A, B]): Json =
```
bocchino, line 33: Instead of a type parameter `A <: BigInt`, can we just use `BigInt`? Similarly for `String` in the next function. Also, instead of `String`, can we use `Name.Unqualified`? The second type is an alias of the first.

jwest115: This is resolved in my latest commit

bocchino (approved):
Looks good!

PR #612, opened 2025-02-13
Configure JVM options

### LINKED ISSUE #610: Revise GC settings for Java build
Remove `-XX:+CMSClassUnloadingEnabled` and add `-XX:+UseG1GC`. This will update to a newer garbage collector, which is the default in Java 17. The first option is no longer supported in Java 17. See https://stackoverflow.com/questions/70214788/unrecognized-vm-option-cmsclassunloadingenabled-after-switching-to-java-17.

bocchino:
* Use the newer G1GC garbage collector. Remove the older GC option that doesn't work in Java 17.
* Increase the heap size to 2G.
* Revise `compiler/install` so it restores `version.scala` properly on failure.

Closes #610.

### DIFF
--- a/compiler/.jvmopts
+++ b/compiler/.jvmopts
@@ -1,2 +1,1 @@
--XX:+CMSClassUnloadingEnabled
--Xss2M
+-XX:+UseG1GC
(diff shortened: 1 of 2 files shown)


bocchino:
So it looks like with the newer garbage collector and without class unloading (which is not supported in the newer GC) we need to increase the heap size. It looks like it's the JAR assembly that blows up the heap.

bocchino:
Closing this PR. We'll keep using Java 11 and the existing GC configuration for now.

Issue #603, opened 2025-02-12
emacs editor indenting for foo[x] -> \ bar


mwette:
Small indenting issue with line continuations:

This works fine:
```
    # comment
    foo -> \
        bar
```

This does not:
```
    # comment
        foo[x] -> \
        bar
```
The TAB key on the line with bar forces the line with foo[x] to get extra indent.

rider: fpp forbids TAB characters in the source, so if indent-tabs-mode=nil can be added that would help.

bocchino:
> rider: fpp forbids TAB characters in the source, so if indent-tabs-mode=nil can be added that would help.

Do you think this would solve the issue?

mwette:
ident-tabs-mode=nil would not solve the issue; I already have that set for fpp-mode

mwette:
The `intersection` forms here  look a bit odd:
```
(defun fpp-mode-indent-function ()
  "A very simple-minded indentation algorithm for FPP."
  (interactive)
  (let ((point-offset (- (current-column) (current-indentation)))
        (has-closing-paren
         (save-excursion
           (beginning-of-line)
           (fpp-mode-re-search 'forward (rx (or (intersection (not "{") "}")
                                                (intersection (not "(") ")")
                                                (intersection (not "[") "]")))
                               (point-at-eol)))))
```
It says take the intersection of all non-{ chars and }, which yields just }.   I'll see if I can come up with something to fulfill the intent.

mwette:
On this specific case the following works.   Maybe  @dennisog could take a look.
```
$ diff -u fpp-mode.el-orig fpp-mode.el
--- fpp-mode.el-orig	2025-02-13 17:51:09.039440365 -0800
+++ fpp-mode.el	2025-02-14 05:31:56.349868942 -0800
@@ -110,17 +110,31 @@
   (while (fpp-mode-is-continued-line t))
   (beginning-of-line))
 
+(defun fpp-mode-nopen ()
+  "Return the number of opens on the current line."
+  (save-excursion
+    (named-let loop ((cnt 0) (pt (beginning-of-line)))
+      (if pt
+	  (if (re-search-forward "{([" (end-of-line) t)
+	      (loop (1+ cnt) (point))
+	    cnt)
+	cnt))))
+
+(defun fpp-mode-nclosed ()
+  "Return the number of closes on the current line."
+  (save-excursion
+    (named-let loop ((cnt 0) (pt (beginning-of-line)))
+      (if pt
[...]

PR #616, opened 2025-02-19
fixes issue 603: indenting for foo[x] -> \ bar

mwette:
Fixes issue #603 

also adds highlighting for 'group'


bocchino:
Thanks, this fix looks good to me! @dennisog, could you take a look at it?

bocchino (approved):

PR #619, opened 2025-02-19
Add port number to assertions caused by disconnected ports

jwest115:
Adds an assert (with port number arg) to check if an output port is connected before calling the port invoke function.
Closes [fprime/3000](https://github.com/nasa/fprime/issues/3000).


bocchino (approved):
Looks good! I merged in #620 so the `fpp-check` tests pass.

PR #620, opened 2025-02-19
Fix compilation error in fpp-to-cpp tests

bocchino:
After merging #591, `check-cpp` broke for the struct tests. The fix is to use the name `T` for the abstract type definition, so that the included header file is `T.hpp`, which is available in the test environment. The name that was being used was not available.


Kronos3 (approved):
Looks good, thanks for updating

PR #623, opened 2025-02-19
Add final specifier to component implementation classes

jwest115:
Closes [fprime/3038](https://github.com/nasa/fprime/issues/3038).


jwest115:
Ran cpp-check and got the following error and warning:

```
In file included from AliasTypeSerializableAc.cpp:7:
./AliasTypeSerializableAc.hpp:10:10: fatal error: 'AbT.hpp' file not found
   10 | #include "AbT.hpp"
      |          ^~~~~~~~~
1 error generated.
```

Warnings:
`./ActiveExternalStateMachines.hpp:14:9: warning: abstract class is marked 'final' [-Wabstract-final-class]`

jwest115:
Disabled abstract final class warnings for the component implementation tests since in the case of external state machines, developers need to provide implementations for any of the pure virtual methods that may exist in external abstract classes.

compiler/lib/src/main/scala/codegen/CppWriter/CppDoc.scala
```diff
@@ -42,9 +42,14 @@ object CppDoc {
     superclassDecls: Option[String],
     /** The class members */
     members: List[Class.Member],
+    /** The class qualifier, if any, before the colon */
+    qualifier: Class.Qualifier = Class.NonFinal
```
bocchino, line 46: Can we call this qualifier `FinalQualifier` for consistency with `ConstQualifier`?

compiler/tools/fpp-to-cpp/test/component/impl/check-cpp
```diff
@@ -7,6 +7,7 @@ export FPRIME_GCC_FLAGS="-I../../fprime"
 warning_flags="
 -Wno-return-type
 -Wno-unused-parameter
+-Wno-abstract-final-class
```
bocchino, line 10: Can we put the new flag first (in alphabetical order)?

compiler/lib/src/main/scala/codegen/CppWriter/CppDoc.scala
jwest115: Done!

compiler/tools/fpp-to-cpp/test/component/impl/check-cpp
jwest115: Done!

bocchino (approved):
Looks good!

PR #627, opened 2025-02-20
Add type aliases to JSON dictionary

### LINKED ISSUE #563: Add type aliases to JSON dictionary

### LINKED ISSUE #626: Alias of type U32 is not displayable
I would expect an alias type to be displayable if it's underlying type is displayable.

Currently, I am getting:

```
product record Record1: AliasT1
                            ^
error: type of record is not displayable

  type AliasT1 = U32
  ^
Type is defined here
```

jwest115:
Adds type aliases to the JSON dictionary. Also addresses #626 by setting the `isDisplayable` property of AliasType according to it's underlying type.

Closes #563 and #626.

### DIFF
--- a/compiler/lib/src/main/scala/analysis/Semantics/Type.scala
+++ b/compiler/lib/src/main/scala/analysis/Semantics/Type.scala
@@ -197,7 +197,7 @@ object Type {
     override def getDefNodeId = Some(node._2.id)
     override def toString = node._2.data.name
     override def isCanonical = false
-    override def isDisplayable = aliasType.getUnderlyingType.isDisplayable
+    override def isDisplayable = getUnderlyingType.isDisplayable
     override def getUnderlyingType = aliasType.getUnderlyingType
   }
 
--- a/compiler/lib/src/main/scala/analysis/Semantics/Type.scala
+++ b/compiler/lib/src/main/scala/analysis/Semantics/Type.scala
@@ -197,6 +197,7 @@ object Type {
     override def getDefNodeId = Some(node._2.id)
     override def toString = node._2.data.name
     override def isCanonical = false
+    override def isDisplayable = aliasType.getUnderlyingType.isDisplayable
     override def getUnderlyingType = aliasType.getUnderlyingType
   }
 
--- a/compiler/tools/fpp-to-dict/test/dictionary.schema.json
+++ b/compiler/tools/fpp-to-dict/test/dictionary.schema.json
@@ -279,6 +279,9 @@
                             },
                             "underlyingType": {
                                 "$ref": "#/$defs/typeDescriptor"
+                            },
+                            "annotation": {
+                                "type": "string"
                             }
                         },
                         "required": [
--- a/compiler/tools/fpp-to-dict/test/dictionary.schema.json
+++ b/compiler/tools/fpp-to-dict/test/dictionary.schema.json
@@ -261,6 +261,32 @@
                             "members",
                             "default"
                         ]
+                    },
+                    {
+                        "type": "object",
+                        "properties": {
+                            "kind": {
+                                "type": "string",
+                                "enum": [
+                                    "alias"
+                                ]
+                            },
+                            "qualifiedName": {
+                                "type": "string"
+                            },
+                            "type": {
+                                "$ref": "#/$defs/typeDescriptor"
+                            },
+                            "underlyingType": {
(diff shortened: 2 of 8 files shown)


compiler/tools/fpp-to-dict/test/dictionary.schema.json
```diff
@@ -261,6 +261,35 @@
...
+                                "type": "string",
+                                "enum": [
+                                    "alias"
+                                ]
+                            },
+                            "qualifiedName": {
+                                "type": "string"
+                            },
+                            "type": {
+                                "$ref": "#/$defs/typeDescriptor"
```
bocchino, line 278: @jwest115 Can you explain what is going on in line 278? It doesn't look right. Why are there dollar signs, a number sign, and slashes?

compiler/tools/fpp-to-dict/test/dictionary.schema.json
```diff
@@ -261,6 +261,35 @@
...
+                            },
+                            "underlyingType": {
+                                "$ref": "#/$defs/typeDescriptor"
+                            },
```
Kronos3, line 282: I don't feel too strongly about this but it seems redundant to include the underlying type since it's something the consumer of the dictionary should be able to compute for themselves.

compiler/tools/fpp-to-dict/test/dictionary.schema.json
jwest115: It's part of JSON Schema's syntax for referencing an internal sub schema (in this case, the typeDescriptor schema specified in line 8). More info at https://json-schema.org/understanding-json-schema/structuring#defs.

compiler/tools/fpp-to-dict/test/dictionary.schema.json
jwest115: Good point! I agree its redundant and can easily be looked up in the dictionary. We were initially thinking about being more "verbose" (ie: include everything and the user of the dictionary can just ignore the things not relevant to them), @bocchino do you have any thoughts?

compiler/tools/fpp-to-dict/test/dictionary.schema.json
bocchino: Thanks! I didn't realize this was part of the schema. It looked like garbled JSON.

compiler/tools/fpp-to-dict/test/dictionary.schema.json
bocchino: @LeStarch @thomas-bc As the customers, can you comment on this? Do you want the underlying type to be part of the type alias definition? For example, when reporting the type aliases `type A = B` and `type B = U32`, do you want the description of the alias `A` to say that its type is `B` and its underlying type is `U32` (as it does now), or do you just want it to say that the type is `B`, leaving the underlying type to be inferred?

bocchino: Note: The underlying type is the unique non-alias type that you get after following the chain of aliases.

compiler/lib/src/main/scala/analysis/Semantics/Type.scala
```diff
@@ -197,6 +197,7 @@ object Type {
     override def getDefNodeId = Some(node._2.id)
     override def toString = node._2.data.name
     override def isCanonical = false
+    override def isDisplayable = aliasType.getUnderlyingType.isDisplayable
```
jwest115, line 200: Simplify this to getUnderlyingType.isDisplayable

compiler/tools/fpp-to-dict/test/dictionary.schema.json
LeStarch: I like the latter.  Makes it easier to get the real type without multiple calculations.

bocchino (approved):
Looks good! Please address the one remaining comment, then merge.

compiler/tools/fpp-to-dict/test/dictionary.schema.json
jwest115: Sounds good! We'll leave the underlying type information in the dictionary.

compiler/lib/src/main/scala/analysis/Semantics/Type.scala
jwest115: Addressed in latest commit, will merge once CI is done running.

PR #631, opened 2025-02-26
Fix alias type value conversion & cycles

### LINKED ISSUE #625: Error received when creating an array with alias type
When trying to create an array with an alias type I receive an internal error (see below):

```
type AliasT1 = U32
array U32x4 = [4] AliasT1 default [1, 2, 3, 4]
```

Produces the following error:
```
Exception in thread "main" internal error: cannot convert value [ 1, 2, 3, 4 ] to type array U32x4
        at fpp.compiler.util.InternalError$.apply(Error.scala:4)
        at fpp.compiler.analysis.Analysis$.convertValueToType(Analysis.scala:565)
```

Kronos3:
This fixes two bugs:
- Values who's expected type was an alias type could not be initialized since the analysis wasn't looking at the underyling type Closes #625
- UseDef cycles were not checked for alias type defs.


bocchino (changes requested):
Looks good! Just a couple of comments to address.

bocchino (approved):
Looks good!

PR #632, opened 2025-02-27
C++ code generation for telemetry packets

### LINKED ISSUE #567: Implement and test C++ code generation for telemetry packets
- [x] Update fpp-filenames and fpp-depend
- [x] Test fpp-filenames and fpp-depend
- [x] Implement code gen
- [x] Test code gen

bocchino:
Closes #567.


jwest115 (approved):

compiler/tools/fpp-to-cpp/test/top/tlm_packets.fpp
```diff
@@ -0,0 +1,134 @@
...
+      }
+
+    } omit {
+      M.c2.T1
+      M.c2.T2
+      M.c2.T3
+      M.c2.T4
+      M.c2.T5
+      M.c2.T6
+    }
```
Kronos3, line 130: The codegen looks good though I am confused about the purpose of `omit`. This block seems a bit counter-intuitive since it's `omit`ing all the channels it put in the group. Reading https://github.com/nasa/fpp/blob/main/docs/spec/Specifiers/Telemetry-Packet-Set-Specifiers.adoc did not clarify the purpose of `omit`

Kronos3 (approved):

compiler/tools/fpp-to-cpp/test/top/tlm_packets.fpp
jwest115: > The codegen looks good though I am confused about the purpose of `omit`. This block seems a bit counter-intuitive since it's `omit`ing all the channels it put in the group. Reading https://github.com/nasa/fpp/blob/main/docs/spec/Specifiers/Telemetry-Packet-Set-Specifiers.adoc did not clarify the purpose of `omit`

In this case, there are multiple component instances. Component instance c1’s channels are included in the telemetry packet and we’re only omitting component instance c2’s channels from the packet.

PR #645, opened 2025-03-03
Implement Type Alias codegen

### LINKED ISSUE #622: Implement and test C++ code generation for type aliases
- [x] Implement
- [x] Test

After discussion with @bocchino we decided to perform type alias codegen with the following:

- Generate a `.hpp` file that has the proper namespaceing based on module
- If the type alias definition is "simple" and supports C header generation (see below):
  - Generate a `.h` file with a C-style `typedef`
  - Generate a `.hpp` file that includes the `.h` file
- Otherwise:
  - Generate a `.hpp` with the proper module/namespacing with the type alias `using` (C++11 syntax)

### Simple Type alias

Simple type aliases are those that support being defined as C types. They must meet all the following requirements
- Exist in the global namespace
- Alias of one of the following types:
  - `U8` `I8` `U16` `I16` `U32` `I32` `U64` `I64` `F32` `F64` 
  -  Another "simple alias type"

Related to #113

Kronos3:
Closes #622


compiler/lib/src/main/scala/codegen/ComputeGeneratedFiles.scala
```diff
@@ -10,7 +10,12 @@ object ComputeGeneratedFiles {
   /** Computes autocoded files (XML, C++ and JSON Dictionary) */
   def getAutocodeFiles(tul: List[Ast.TransUnit]): Result.Result[List[String]] =
     for {
+      // Perform analysis only up to the point we need
       a <- enterSymbols(tul)
+      a <- CheckUses.visitList(a, tul, CheckUses.transUnit)
```
bocchino, line 15: Why is this analysis added here? It looks like nothing else has changed in this file.

compiler/lib/src/main/scala/codegen/CppWriter/ComputeAutocodeCppFiles.scala
```diff
@@ -26,6 +26,26 @@ object ComputeAutocodeCppFiles extends ComputeCppFiles {
...
 
+  override def defAliasTypeAnnotatedNode(
+    s: State,
+    aNode: Ast.Annotated[AstNode[Ast.DefAliasType]]
+  ) = {
+    val node = aNode._2
+    val name = s.getName(Symbol.AliasType(aNode))
+    val loc = Locations.get(node.id)
+    val fileName = ComputeCppFiles.FileNames.getAliasType(name)
+    val t = s.a.typeMap(node.id)
```
bocchino, line 37: OK, maybe this is why we need the type analysis

compiler/lib/src/main/scala/codegen/ComputeGeneratedFiles.scala
bocchino: It looks like it's needed in ComputeAutocodeCppFiles

compiler/lib/src/main/scala/codegen/CppWriter/AliasCppWriter.scala
```diff
@@ -0,0 +1,188 @@
...
+      fileName,
+      includeGuard,
+      getHMembers,
+      s.toolName,
+      "h"
+    ))
+  }
+
+  private def getHppMembers: List[CppDoc.Member] = {
+    List(
```
bocchino, line 89: It's better to use List.concat here instead of List(...).flatten.

compiler/lib/src/main/scala/codegen/CppWriter/AliasCppWriter.scala
```diff
@@ -0,0 +1,188 @@
...
+      getHIncludes,
+      getHOpenCPPGuard,
+      getHDefinition,
+      getHCloseCPPGuard,
+    )
+  }
+
+  private def getHOpenCPPGuard: CppDoc.Member = {
+    linesMember(addBlankPrefix(lines(
+        """|#ifdef __cplusplus
```
bocchino, line 106: Can we omit the #ifdef __cplusplus? This is a C++-only header, right?

compiler/tools/fpp-to-cpp/test/alias/TAliasAc.ref.h
```diff
@@ -0,0 +1,23 @@
...
+// \author Generated by fpp-to-cpp
+// \brief  h file for T alias
+// ======================================================================
+
+#ifndef TAliasAc_H
+#define TAliasAc_H
+
+#include "Fw/Types/BasicTypes.h"
+
+#ifdef __cplusplus
```
bocchino, line 12: Let's omit `#ifdef __cplusplus extern "C"` here and add `extern "C"` in the hpp file.

compiler/lib/src/main/scala/codegen/CppWriter/AliasCppWriter.scala
```diff
@@ -0,0 +1,185 @@
...
+    linesMember(addBlankPrefix(headers.distinct.sorted.map(line)))
+  }
+
+  private def getHppIncludes: CppDoc.Member.Lines = {
+    val standardHeaders = List(
+      aliasType.aliasType match {
+        case Type.String(_) => "Fw/Types/String.hpp"
+        case Type.AbsType(node) =>
+           s.isBuiltInType(node._2._1.name) match {
+            case true => "FppConfig.hpp"
```
bocchino, line 135: Let's make FpConfig.hpp a system header. Also I think it's FpConfig.hpp and not FppConfig.hpp.

bocchino: It seems we need to add a test for this case.

compiler/lib/src/main/scala/codegen/CppWriter/CppWriterState.scala
Kronos3, line 1: Turns out I already had a test for this but it wasn't getting `cpp-check`ed because alias tests don't generate C++ source code (headers only). I added a struct that references the alias types to all the tests.

I found that the builtin type handling wasn't handling when the underlying type was a builtin so this change was needed.

compiler/lib/src/main/scala/codegen/CppWriter/CppWriterState.scala
```diff
@@ -192,7 +192,12 @@ case class CppWriterState(
   }
 
   /** Is t a primitive type (not serializable)? */
-  def isPrimitive(t: Type, typeName: String): Boolean  = isBuiltInType(typeName) || t.getUnderlyingType.isPrimitive
+  def isPrimitive(t: Type, typeName: String): Boolean  = (
+    isBuiltInType(typeName) ||
+    t.getUnderlyingType.isPrimitive ||
+    // See if this an alias of a builtin type
+    isBuiltInType(t.getUnderlyingType.toString())
+  )
```
Kronos3, line 200: Oops I meant to link to this code in the above comment

compiler/tools/fpp-to-cpp/test/component/test-impl/check-cpp
```diff
@@ -9,7 +9,7 @@ warning_flags="
 -Wno-unused-parameter
 "
 include_flags="
--I$FPRIME/gtest/googletest-src/googletest/include
+-I$FPRIME/googletest/googletest/include/
```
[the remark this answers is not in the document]
Kronos3: We found out that this directory is generated when building  FPrime. This change just points to the base source header which doesn't require a build.

compiler/tools/fpp-to-cpp/test/component/base/TypedReturnPortAc.ref.cpp
```diff
@@ -49,7 +49,7 @@ namespace Ports {
         bool b,
         const Fw::StringBase& str2,
         const E& e,
-        const A& a,
```
bocchino, line 52: What is the reason for changing A to AA here and elsewhere?

Kronos3: Oops, I added a `array AA = [3] A` in the `types.fpp` during testing but forgot to change it back. I fixed it and expanded my test cases for recursive aliasing of array types.

bocchino (approved):
Looks good!

bocchino:
The `fpp-check` tests pass against this F Prime branch: https://github.com/nasa/fprime/pull/3393.

PR #650, opened 2025-03-05
Update User's Guide

### LINKED ISSUE #646: Document autocoded serialization and deserialization
The behavior of the autocoded ser/deser functions for structs/arrays/enums should be well-defined and guaranteed not to change.
As discussed with @bocchino , for serialization, the definition should be:
1. struct members are serialized in the order defined in fpp
2. no padding
3. big endianness

Additionally, notes should be made about potentially ambiguous serialization behavior for types like Fw::String and the various buffer types. It should be made clear that serializing a primitive type (int, float or char) behaves exactly as you'd expect.

### LINKED ISSUE #648: Implementation of Abstract FPP Type in CPP
According to the following text taken from the FPP User's Guide, an abstract type MUST implement the following methods:
image

However, there is at least one abstract type implementation that does not include all of these methods: `CmdArgBuf` (and it also doesn't superclass Fw::Serializable)
image
As you can see, it does not implement `operator==`. In the superclass, `operator==` is not defined if we aren't building UTs:
image

`CmdArgBuf` works perfectly fine when used in a port:

Image

But not when used in a struct, producing the following error:
image

Can the requirements for implementing an abstract type be clarified?

### LINKED ISSUE #649: Replace builtin types with type aliases
We need to remove the special handling used in our Codegen with builtinTypes:

https://github.com/nasa/fpp/blob/37b53031ec9926e7a4249e407052d547b2922155/compiler/lib/src/main/scala/codegen/CppWriter/CppWriterState.scala#L201-L218

This will need a sister PR on the mainline FPrime repo to point the default config to aliases

bocchino:
* Document the use of `Fw::SerializeBufferBase` to implement abstract types.
* Document the rules for serializing FPP values.

Closes #646. Closes #648.

### DIFF
--- a/docs/fpp-users-guide.html
+++ b/docs/fpp-users-guide.html
@@ -15017,17 +15017,16 @@ with no padding.</p>
 each member is serialized using these rules.</p>
 </li>
 <li>
-<p>Each member array is serialized as described above
-for a value of array type.</p>
+<p>Each member array is serialized as stated in rule 5.</p>
 </li>
 </ol>
 </div>
 </li>
 <li>
-<p>A value of <a href="#Defining-Enums">enum type</a> is serialized according
-to its <a href="#Defining-Enums_The-Representation-Type">representation type</a>.
-The representation type is serialized as described above
-for primitive integer values.</p>
+<p>A value of <a href="#Defining-Enums">enum type</a> is converted to a primitive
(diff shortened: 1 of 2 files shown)


zimri-leisher:
Thanks for fixing these! Now we can finally use FPP autocoded deserialization to read ICD-defined structs.

LeStarch (approved):

PR #653, opened 2025-03-11
Fix C++ code generation for ports with string return type 

### LINKED ISSUE #651: Returning a string from a port does not work
The following FPP model causes fpp-to-cpp to produce bad C++:
```
port P -> string

passive component C {

  sync input port pIn: P

}
```
The type used to represent concrete string values is `Fw::StringBase` when it should be `Fw::String`.

This case has slipped through all our tests (FPP unit tests, FppTest in F Prime) because it's not covered. There is some code in FppTest to cover this case, but it's commented out.

- [x] Update code gen for ports
- [x] Update unit tests for ports
- [x] Update code gen for component bases
- [x] Update unit tests for component bases
- [x] Update code gen for component implementations
- [x] Update unit tests for component implementations
- [x] Update code gen for component tester bases
- [x] Update unit tests for component tester bases
- [x] Update code gen for component test implementations
- [x] Update unit tests for component test implementations
- [x] Update FppTest

bocchino:
* Update the code generators for ports and components
  * Incidental: Revise the `"FpConfig.hpp"` header include to `<FpConfig.hpp>` to conform to F Prime conventions
  * Regenerate unit tests
* Add unit tests to cover the new behavior

Closes #651.

A small change to F Prime was required. To run the `check-cpp` tests, use this branch: https://github.com/bocchino/fprime/commit/b6cf5ccff4e844cbac270226af5b375c35572a6b.


Kronos3 (commented):
Looks good. I wonder if the if this still works if you return a type with a string nested in another type like a struct or an array?

bocchino:
It should work because structs and arrays use `Fw::ExternalString` for string members. Are you suggesting that we add a test case for this kind of return type?

bocchino:
Note that there are fpp-to-cpp tests for structs and arrays with string members, in the `struct` and `array` subdirectories.

Kronos3 (approved):
Ah right, and `Fw::ExternalString` will have a buffer managed by the port handler in the derived component code so it doesn't concern the base class at all.

PR #655, opened 2025-03-14
Add telemetry packets to fpp-from-xml

### LINKED ISSUE #630: Add support for telemetry packets to fpp-from-xml
So users can translate their existing XML packet specifications to FPP. We can also use this to convert the telemetry packets in the Ref application in the F Prime repository.

- [x] Implement
- [x] Test

bocchino:
Closes #630.

### DIFF
--- a/compiler/tools/fpp-from-xml/test/top/fprime_ref_packets.fpp
+++ b/compiler/tools/fpp-from-xml/test/top/fprime_ref_packets.fpp
@@ -1,0 +1,5 @@
+topology T {
+
+  include "fprime_ref_packets.txt"
+
+}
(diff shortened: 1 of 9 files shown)


compiler/tools/fpp-from-xml/test/top/fprime_ref_packets.fpp
```diff
@@ -0,0 +1,5 @@
+topology T {
+
+  include "fprime_ref_packets.txt"
```
LeStarch, line 3: I don't understand this change.  It includes a `.txt`?  I don't see that file anywhere in this change set.

LeStarch (approved):
I had a brief question, but I see the output test reference output I'd expect. I appreciate the refactors to!

No blockers.

compiler/tools/fpp-from-xml/test/top/fprime_ref_packets.fpp
bocchino: Good catch! The file name was wrong. There was another mistake that masked this mistake. I fixed both.

PR #658, opened 2025-03-17
Mark Generated C/C++ files as generated

Kronos3:
Before:
```
$ github-linguist
86.58%  11423656   C++
10.58%  1395476    Scala
1.39%   183882     Fortran
1.32%   173503     Shell
0.08%   10088      Emacs Lisp
0.03%   4287       Vim Script
0.02%   3215       Python
```

After:
```
$ github-linguist
76.57%  1395476    Scala
10.09%  183882     Fortran
9.52%   173503     Shell
2.85%   52011      C++
0.55%   10088      Emacs Lisp
0.24%   4287       Vim Script
0.18%   3215       Python
```

I believe the remaining C++ files are our `.hpp` for abstract type definitions. We can also mark the `.fpp` files as any language other than Fortran if we wanted to (though the language must exist according to github-linguist unfortunately).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# Mark generated reference files as vendored
+*.ref.cpp linguist-vendored
+*.ref.hpp linguist-vendored
+*.ref.h linguist-vendored
+
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.tasty
 *~
 .*
+!.gitattributes
 __SHADOW__
 __pycache__
 dist/


bocchino:
Thanks, this looks good! Question: Do you know what "vendored" is intended to be used for? It's a bit of an unusual term. I suppose we don't use "generated" for the C++ test output files because they are checked into the repo.

Kronos3:
Here's the set of overrides that linguist allows:

https://github.com/github-linguist/linguist/blob/main/docs/overrides.md#summary


 | Git attribute                                  | Defined in            | Effect on file                                                  |
 |:-----------------------------------------------|:----------------------|:----------------------------------------------------------------|
 | `linguist-detectable`                          | [`languages.yml`]     | Included in stats, even if language's type is `data` or `prose` |
 | `linguist-documentation`                       | [`documentation.yml`] | Excluded from stats                                             |
 | `linguist-generated`                           | [`generated.rb`]      | Excluded from stats, hidden in diffs                            |
 | `linguist-language`=<var><ins>name</ins></var> | [`languages.yml`]     | Highlighted and classified as <var><ins>name</ins></var>        |
 | `linguist-vendored`                            | [`vendor.yml`]        | Excluded from stats                                             |


There is "generated" but that means it won't show up in the diffs. That's usually used for code thats generated in someones tree that just does boilerplate. Since the generated files are part of the output, we definitely want the diffs.

bocchino (approved):
Looks good!

PR #660, opened 2025-03-19
Expand fpp-to-json tests 

### LINKED ISSUE #468: Expand fpp-to-json tests to run against fpp-check
Doing this may uncover issues with the native build that are not caught by the FPP unit tests.

jwest115:
Run fpp-to-json on fpp-check test cases to improve code coverage.
Update trace file to prevent crashes when running the fpp-to-json native build.

Closes #468.


bocchino (approved):
Looks good! I refactored the new fpp-to-json/fpp-check tests a bit. I moved those tests to a subdirectory. I used `find` instead of iterating over directories to find the fpp-check test input files. It's more efficient, and it finds files at all directory levels; before we were only going down one level. I revised the format of the test output so that the longer test names don't crash into the PASSED or FAILED output.

I also ran `dostounix` on the top-level source file for `fpp-to-json`. There were some stray ^M characters in there, probably from back when the file was created.

PR #669, opened 2025-03-26
Remove builtin types

Kronos3:
This PR removes the special handling of "builtin" types, which are type aliases that were previously handled inside `FpConfig.h` by C-level typedefs.

Some notes:
- I removed the `#include <FpConfig.h[pp]>` from generated alias headers since it's redundant now. It was previously there in case there was an alias of a built-in
- I removed the tests that tested builtin types since they are no longer relevant as there is no additional code coverage over the simple type alias tests.
- I found a bug in the `#define PRI` code that was not referencing built-in PRIs properly
  - We had not hit this yet because C++ code did not use an alias type's `#define` in generated code. With builtin's removed now the old builtin `PRI`s  were being hit.
  - The common pattern was like this:
```
// Queue priority type
typedef PlatformQueuePriorityType FwQueuePriorityType;
#define PRI_FwQueuePriorityType PRI_PlatformQueuePriorityType

// The type used to serialize a size value
typedef U16 FwSizeStoreType;
#define PRI_FwSizeStoreType PRIu16
```

Notice that builtin types like `U16` have PRIs defined `PRI[typeName.lower()` where-as aliases are `PRI_[typeName]`. That underscore was always assumed in previously generated code (see diffs for what I'm talking about).


Kronos3:
I've opened a sister PR in https://github.com/nasa/fprime/pull/3417 that updates the `FpConfig.h` which this PR's `check-cpp` will depend on.

compiler/tools/fpp-to-cpp/test/alias/SimpleCTypeAliasAc.ref.h
```diff
@@ -7,12 +7,10 @@
 #ifndef SimpleCTypeAliasAc_H
 #define SimpleCTypeAliasAc_H
 
-#include <FpConfig.h>
-
 #include "Fw/Types/BasicTypes.h"
```
bocchino, line 10: I think the current pattern is to include `<FpConfig.h>` and not include `Fw/Types/BasicTypes.h`. However, Michael may be changing how this works. @LeStarch can you comment?

Kronos3: This will introduce an include cycle since builtin aliases will depend on `FpConfig.h` and vis versa. I'm not sure if that that'll cause compilation issues since I haven't tried it though.

bocchino: I don't think that `FpConfig.h` should depend on type aliases. By introducing the aliases, we are moving the configuration out of `FpConfig.h` and into FPP. Maybe the issue is how these type aliases are included in the C++ code that needs them. The ordinary way that we include headers in generated code is from the symbols that are used in the code. Now we have some implied uses of type aliases that are not tracked, e.g., a command specifier represents an implied use of `FwCmdOpcodeType`. Should we just add these implied uses to the analysis? That seems the most straightforward way. Otherwise it seems we need to be conservative and include all the basic type aliases in every generated hpp file. But I think we should do that directly in the generated code, not by including a framework header like `FpConfig.h` or `BasicTypes.h` and making those headers depend on generated code.

bocchino: I guess there's also an issue of how these type aliases are included in handwritten code. Currently when you include `FpConfig.h` you get all these types. So maybe we should have a header that users can include to get them all. But I don't think it should be `FpConfig.h`, which is for hand-configuring definitions in C/C++.

bocchino: I conferred with Michael. He is working on a change in F Prime where we will include a types header (similarly to what you have) that will include the config header. I think we should hold off merging this PR until that change is done in F Prime.

bocchino: Also, eventually we can do both kinds of includes: (1) the types header will include the generated header for `FwOpcodeType` and (2) each use of `FwOpcodeType` seen by the analyzer will cause an include of the header, as usual for definitions and uses. The include guard will prevent the include from actually occurring twice.

Kronos3:
Future note to self, the PRIs for scalars are defined twice with slightly different names:
https://github.com/nasa/fprime/pull/3422/files#diff-4a4e7f315f68d77a9f724564407a9a79e284175fbd14fd0e81a428a00dca0293R51-R88

LeStarch:
@bocchino should be ready for re-review, and once finished a point-release!

compiler/lib/src/main/scala/codegen/CppWriter/AliasCppWriter.scala
```diff
@@ -120,27 +120,16 @@ case class AliasCppWriter (
...
     val headers = (standardHeaders ++ symbolHeaders).distinct.sorted.map(line)
-    linesMember(List.concat(
-      addBlankPrefix(systemHHeaders),
-      addBlankPrefix(headers)
-    ))
+    linesMember(addBlankPrefix(headers))
   }
 
   private def getHppIncludes: CppDoc.Member.Lines = {
-    val systemHppHeaders = List(
```
bocchino, line 140: I think we need to include BasicTypes.hpp here.

compiler/tools/fpp-to-cpp/test/component/base/ActiveAsyncProductsComponentAc.ref.hpp
```diff
@@ -7,7 +7,7 @@
 #ifndef ActiveAsyncProductsComponentAc_HPP
 #define ActiveAsyncProductsComponentAc_HPP
 
-#include <FpConfig.hpp>
+#include <Fw/FPrimeBasicTypes.hpp>
```
bocchino, line 10: Let's write `#include "Fw/FPrimeBasicTypes.hpp"` instead of `#include <Fw/FPrimeBasicTypes.hpp>`, as elsewhere.

bocchino:
@LeStarch. It looks good so far. Just a couple of comments:

1. I think there may be a missing header include for `Fw/FPrimeBasicTypes.hpp`. I noted this in an inline comment.

2. Let's be consistent and write `#include "Fw/FPrimeBasicTypes.hpp"` everywhere. Sometimes it is written `#include <Fw/FPrimeBasicTypes.hpp>`. I started noting this in comments, but then there were several cases. Let's sweep through and make everything consistent. `grep -r "#include <Fw/FPrimeBasicTypes.hpp>"` should come up clean.

Where is the corresponding F Prime branch, so I can run `check-cpp` and verify that it passes?

bocchino:
@LeStarch I'm not seeing `Fw/FPrimeBasicTypes.hpp` on that branch.

compiler/tools/fpp-to-cpp/test/component/base/ActiveAsyncProductsComponentAc.ref.hpp
LeStarch: Agreed!

compiler/lib/src/main/scala/codegen/CppWriter/AliasCppWriter.scala
bocchino: OK, then I think we should include `Fw/Type/BasicTypes.hpp`. I think you are right, if the alias refers to a basic type, then the header for that specific type should be automatically included. Can we also add a comment explaining why we are using `Fw/Type/BasicTypes.hpp` instead of `Fw/FPrimeBasicTypes.hpp` here?

LeStarch:
@bocchino, I have fixes the issues, except for the one I left a comment about.  I have also pushed to the FPrime branch so you can run cpp-check.

I have:
1. Fixed suggestions
2. Updated refs
3. Confirmed No angle bracket includes for the new header
4. Ran cpp-check

bocchino (approved):
Looks good! I added a few comments explaining what is going on with the headers in the alias code gen. I confirmed that check-cpp runs cleanly.

bocchino:
Before merging this branch into fpp/main, I would like to get https://github.com/nasa/fprime/pull/3434 reviewed and merged. It fixes problems in FppTest, which we need for testing the integration of this branch with F Prime.

bocchino:
I confirmed that everything works with the updated FppTest. Merging now.

PR #670, opened 2025-03-27
Adds in `doDispatch` Helpers and Tester Base Friend Class

LeStarch:
This paves the way for removes `PRIVATE` and `PROTECTED`.

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriter.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriter.scala
@@ -217,7 +217,7 @@ case class ComponentCppWriter (
 
       // Protected/private function members
       getDispatchFunctionMember,
-      guardedList (data.kind == Ast.ComponentKind.Queued) (getDispatchCurrentMemebers),
+      guardedList (data.kind == Ast.ComponentKind.Queued) (getDispatchCurrentMembers),
 
       // Private function members
       portWriter.getPrivateFunctionMembers,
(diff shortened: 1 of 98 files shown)


LeStarch:
I failed to bump UT sources of truth.  Will do shortly.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriter.scala
```diff
@@ -905,6 +908,44 @@ case class ComponentCppWriter (
     }
   }
 
+  private def getDispatchCurrentMemebers: List[CppDoc.Class.Member] = {
+    val body = intersperseBlankLines(
```
bocchino, line 912: I think we don't need the `intersperseBlankLines` and we don't need the leading spaces in the body lines.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/TestWriter/ComponentTesterBaseWriter.scala
```diff
@@ -627,6 +630,71 @@ case class ComponentTesterBaseWriter(
...
+      List(
+        lines(
+          """|    // Dispatch one message returning status
+             |    return component.doDispatch();
+             |"""
+        ),
+      )
+    )
+    val allBody = intersperseBlankLines(
+      List(
```
bocchino, line 645: I think we don't need the `intersperseBlankLines` and we don't need the leading spaces in the body lines.

bocchino (changes requested):
Please regenerate the unit test output. Also, should we revise the `check-cpp` script so that it compiles with `PROTECTED=protected` and `PRIVATE=private` here?

https://github.com/nasa/fpp/blob/6e976469ef31e734f694ce615b6f009f861e386c/compiler/tools/fpp-to-cpp/test/component/test-base/check-cpp#L24-L25

I think we should, to verify that the friend classes are working as expected. Then later we can take out the `PUBLIC` and `PRIVATE` symbols altogether.

LeStarch:
> Please regenerate the unit test output. Also, should we revise the `check-cpp` script so that it compiles with `PROTECTED=protected` and `PRIVATE=private` here?
> 
> https://github.com/nasa/fpp/blob/6e976469ef31e734f694ce615b6f009f861e386c/compiler/tools/fpp-to-cpp/test/component/test-base/check-cpp#L24-L25
> 
> I think we should, to verify that the friend classes are working as expected. Then later we can take out the `PUBLIC` and `PRIVATE` symbols altogether.
> 
> Also here:
> 
> https://github.com/nasa/fpp/blob/6e976469ef31e734f694ce615b6f009f861e386c/compiler/tools/fpp-to-cpp/test/component/test-impl/check-cpp#L20

Agreed.  I am taking a similar path on F prime itself!

LeStarch:
@bocchino:
1. Fixed your suggestions, thank you!
2. Updated UT refs
3. Confirmed `check-cpp` works against `nasa/devel`

bocchino (approved):
Looks good!

PR #673, opened 2025-04-03
Changing port formatting to use proper formatter PRI_FwIndexType

LeStarch:
Port name formatting code was using `PRI_PlatformIntType` instead of the new port index type formatter `PRI_FwIndexType`.  This change fixes that slight issue.


LeStarch:
Some details.  The variable type is seen on line 1054 of the changed scala file: `"FwIndexType port = 0",`.

Refs updated, UTs passed, and cpp-check passed!

PR #675, opened 2025-04-07
Document alias types in User's Guide

### LINKED ISSUE #636: Add type aliases to the User's Guide
Update the User's Guide to cover type aliases.

bocchino:
Closes #636.

### DIFF
--- a/docs/users-guide/Analyzing-and-Translating-Models.adoc
+++ b/docs/users-guide/Analyzing-and-Translating-Models.adoc
@@ -373,6 +373,8 @@ translated to {cpp} files.
 |State machine _M_ in component _C_|_C_ `_` _M_ `StateMachineAc.{hpp,cpp}`
 |Struct _S_ outside any component|_S_ `SerializableAc.{hpp,cpp}`
 |Struct _S_ in component _C_|_C_ `_` _S_ `SerializableAc.{hpp,cpp}`
+|Alias type _A_ outside any component|_A_ `AliasAc.{hpp,h,cpp}`
+|Alias type _A_ in component _C_|_C_ `_` _A_ `AliasAc.{hpp,h,cpp}`
 |Port _P_|_P_ `PortAc.{hpp,cpp}`
 |Component _C_|_C_ `ComponentAc.{hpp,cpp}`
 |Topology _T_|_T_ `TopologyAc.{hpp,cpp}`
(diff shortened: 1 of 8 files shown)


Kronos3 (changes requested):
Looks good, one minor thing: I think it would be good to explain the semantics of `string` aliases (i.e. `size` is ignored from the alias)

bocchino:
Whoops, I pushed my changes to the wrong branch.

bocchino:
OK, it should be ready for re-review now.

docs/users-guide/Analyzing-and-Translating-Models.adoc
```diff
@@ -564,6 +601,30 @@ the qualified name of the definition.
 For example, a component whose qualified name in FPP is `A.B.C`
 uses the name `A_B_CComponentAc_HPP` in its include guard.
 
+*Alias types:*
+The generated code for an <<Defining-Types_Alias-Type-Definitions,alias type
+definition>> is either a C-style `typedef` (in a `.h` file) or a {cpp}11-style
+`using` alias (in a `.hpp` file).  The generated code for an alias to an FPP
+`string` type is always an alias to the F Prime type `Fw::String`; the size in 
+the `string` type is ignored. For example, the following alias type definitions
```
Kronos3, line 609: In #676 size is not _always_ ignored. If it is used in storage (i.e. struct serialization) it keep the size sematics of the underlying string definition.

docs/users-guide/Defining-Types.adoc
```diff
@@ -6,11 +6,12 @@ model and that may generate code in the target language.
 For example,
 an FPP type definition may become a class definition in {cpp}.
 
-There are three kinds of type definitions:
+There are four kinds of type definitions:
 
 * Array type definitions
 * Struct type definitions
 * Abstract type definitions
+* Alias type definitions
```
Kronos3, line 14: What about enum definitions?

Kronos3 (changes requested):

docs/users-guide/Analyzing-and-Translating-Models.adoc
bocchino: OK, I added "when generating the alias" after "is ignored." Is this sufficient clarification?

Kronos3: Ya that works

bocchino:
> I think you still need to but up the last update to the string size.

I'm not sure what this means. Can you clarify?

Kronos3:
> OK, I added "when generating the alias" after "is ignored." Is this sufficient clarification?

I didn't see a commit for this change

bocchino:
Got it! I made the same mistake as before when juggling the branches. Fixed now.

Kronos3 (approved):

Issue #659, opened 2025-03-18
Add type aliases to FppTest


bocchino:
Add tests in the FppTest project of nasa/fprime to cover type aliases.

PR #676, opened 2025-04-10
Fix string alias codegen

Kronos3:
#645 did not get the string alias codegen quite right.  We need to keep the same underlying C++ type in all locations that utilize FPP strings (and their aliases). 

~~This PR will only utilize the aliased name if the codegen expects `Fw::String`. There was a missing testcase that caused issues while working on #659 (see updated `.fpp` test).~~

We decided that the original (pre-alias) behavior should be maintained for string aliases. A C++ `using AliasName = Fw::String` is still generated


compiler/tools/fpp-to-cpp/test/struct/alias_type.fpp
```diff
@@ -1,4 +1,9 @@
 type U16Alias = U16
 type T
 type TAlias = T
-struct AliasType { x: U16Alias, y: TAlias }
+type SAlias = string size 50
+struct AliasType {
+    x: U16Alias,
+    y: TAlias,
+    z: [50] SAlias
```
bocchino, line 8: Maybe we can reduce this array size from 50 to 10? To cut down on the amount of generated test code.

compiler/lib/src/main/scala/codegen/CppWriter/TypeCppWriter.scala
```diff
@@ -19,7 +19,13 @@ case class TypeCppWriter(
 
     override def aliasType(s: CppWriterState, t: Type.AliasType) =
       (stringTypeName, t.getUnderlyingType) match {
-        case ("Fw::StringBase", Type.String(_)) => stringTypeName
+        // String aliases are aliases for Fw::String
+        // If the parent is asking for this we can keep alias source semantics
+        // and still utilize the C++ type alias
+        case ("Fw::String", Type.String(_)) => s.writeSymbol(Symbol.AliasType(t.node))
```
bocchino, line 25: This looks right, but I wonder if it would be clearer to just remove the first case here, and always use the underlying string type. It might be confusing to use the alias name just in the case that the underlying type happens to be Fw::String.

bocchino (commented):
Good catch! This PR looks good to me. Please look at the comments I made.

compiler/lib/src/main/scala/codegen/CppWriter/TypeCppWriter.scala
bocchino: In general, using the alias name in the generated code is a good policy, but for strings it seems problematic, since we don't fully represent string type aliases in the C++ code.

Kronos3: I think the only spot we will see this will be in getter and port returns. Works for me either way.

I was also thinking that since string aliases are always `Fw::String` there's no need to generate the C++ type either. Essentially string type aliases would be there purely to propagate string sizes to uses in other parts of the FPP model.

bocchino:
Looks good! I'm just running check-cpp now.

compiler/lib/src/main/scala/codegen/CppWriter/TypeCppWriter.scala
```diff
@@ -18,14 +18,8 @@ case class TypeCppWriter(
...
-      (stringTypeName, t.getUnderlyingType) match {
-        // String aliases are aliases for Fw::String
-        // If the parent is asking for this we can keep alias source semantics
-        // and still utilize the C++ type alias
-        case ("Fw::String", Type.String(_)) => s.writeSymbol(Symbol.AliasType(t.node))
-        // This is a string type that is something other than what our alias uses
-        // as its underlying type. Use the type requested by the calling code
-        case (_, Type.String(_)) => stringTypeName
+      t.getUnderlyingType match {
+        case Type.String(_) => stringTypeName
```
bocchino, line 22: It might be good to add a comment here saying why Type.String is a special case. It's because all FPP string type aliases are represented as Fw::String in the generated C++ code.

bocchino (approved):
Looks good! check-cpp passed.

compiler/lib/src/main/scala/codegen/CppWriter/TypeCppWriter.scala
Kronos3: Added!

PR #677, opened 2025-04-10
Fix bug in syntax writers

bocchino:
The AstWriter and the FppWriter were ignoring the optional default expression for enum definitions.

* Fix writers
* Add tests to cover this case


Kronos3 (approved):
Looks good to me!

PR #680, opened 2025-04-12
Fix bug in code gen for C++ constant defs

bocchino:
Unit test and implementation generators were generating constant defs. Only the autocode generator should do that.


Kronos3 (approved):
Looks good to me!

PR #683, opened 2025-04-17
Fixed assert in from_prmSetOut_static

Brian-Campuzano:
There is a serial buffer being reset in the ${component}TesterBase, specifically in the `from_prmSetOut_static()` function. After the serial buffer is reset, it is deserialized. The buffer being reset causes the deserialize function to return `FW_DESERIALIZE_BUFFER_EMPTY` which trips an assert.

The solution is to remove the serial buffer reset.


bocchino (approved):
Looks good! I confirmed that check-cpp passes and the F Prime unit tests pass against this branch.

PR #684, opened 2025-04-22
Fast lexer

Kronos3:
This PR reimplements the lexing strategy that produces a hugely faster lexer (8-10x speedup). It uses the same design pattern as can be found in the Scala compiler using a manually written recursive descent parser that moves through the file character-by-character. I also included some fixes in the parser that speed up the parsing by ~30% using IntelliJ's static analysis and auto-fixes. Mainly the fixes removed some redunant syntax and added `private` to functions that could be internalized (allows inlining and other optimizations by Scala).

## Notable changes

There are some very minor discrepancies between the old and new lexer, @bocchino I didn't get exact feature parity because some behavior seems a little under defined so I'll wait for your input:

1. Trailing spaces after a `\` line continuation are disallowed. There is a single file in the FPrime repo that has some trailing spaces which throws an error in the lexer. I'm leaning toward keeping the functionality as is since it enforces cleaner syntax.
2. The lexer uses an error `Context` to track invalid tokens or tokenization errors.
3.  The string trimming of multi-line strings (3-strings) is pretty confusing (or bugged). The old behavior had odd behavior:

```
constant F: string = """
    indented
    keeps
    trailing newline
    """
```

Generates a string with value: `indented\nkeeps\ntrailing newline\n`

```
constant F: string = """
non-indented
does not keep
trailing newline
```

Generates string with value: `non-indented\ndoes not keep\ntrailing newline`

For now the new lexer will not string trailing newlines ever but I can easily add that in.

## Benchmarks

To benchmark I ran a JVM-based profiler to compare lexing times on the locs generation step of the FppTest build (this is a parsing intensive build stage). The images screenshots of the runtime of the lexer before and after this PR. I also `time`ed the runtime of the native GraalVM build of `fpp-locate-defs` before and after this PR. The stats are included below.

Before:
image

```
real	0m3.460s
user	0m3.389s
sys	0m0.071s
```

After:
image
```
real	0m0.335s
user	0m0.284s
sys	0m0.052s
```

### DIFF
--- a/compiler/lib/src/main/resources/META-INF/native-image/reflect-config.json
+++ b/compiler/lib/src/main/resources/META-INF/native-image/reflect-config.json
@@ -3762,14 +3762,6 @@
   "fields":[{"name":"0bitmap$1"}]
 },
 {
-  "name":"fpp.compiler.syntax.Lexer$",
-  "fields":[{"name":"0bitmap$2"}]
-},
-{
-  "name":"fpp.compiler.syntax.Lexer$$anon$1",
-  "fields":[{"name":"0bitmap$1"}]
-},
-{
   "name":"fpp.compiler.syntax.Parser$",
   "fields":[{"name":"0bitmap$1"}]
 },
--- a/compiler/lib/src/main/scala/analysis/Analysis.scala
+++ b/compiler/lib/src/main/scala/analysis/Analysis.scala
@@ -261,7 +261,7 @@ case class Analysis(
     val Value.Integer(v) = Analysis.convertValueToType(
       valueMap(id),
       Type.Integer
-    )
+    ): @unchecked
     v
   }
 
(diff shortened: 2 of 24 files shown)


bocchino:
This is great! The performance gain is impressive, and it should help speed up the F Prime builds. Let's discussed the proposed changes to the spec.

Kronos3:
I have pushed the following changes
- Clarified whitespace handling of multi-line string literals in the spec. The old behavior did not match the spec. We decided to update the spec to match the behavior.
- Updated lexer to allow trailing spaces after a line continuation, any other character will trigger an error
- Added a testcase for line continuation and multi-error reporting (updated `illegal-character` test in `fpp-syntax`)

Kronos3:
More interesting benchmarks from our CI jobs:

Before
Screenshot 2025-04-29 at 11 34 00 AM

After
Screenshot 2025-04-29 at 11 33 52 AM

Before
Screenshot 2025-04-29 at 12 47 04 PM

After
image

bocchino (approved):
Looks great! I made a few minor changes. When CI passes, I will merge it.

Kronos3:
I neglected to put the most important benchmark here which is the cmake configuration time. I purged FppTest before each run and ran a native GraalVM build before/after:

Before:
```
-- Configuring done (23.9s)
-- Generating done (1.7s)
```

After:
```
-- Configuring done (10.4s)
-- Generating done (1.7s)
```

PR #686, opened 2025-04-25
Require Fw Types in dictionary

### LINKED ISSUE #666: Add Fw types to dictionary
Update dictionary to include Fw types. Some possible implementations may be:
- Associate the types with the uses of values that should have that type. For example, in the UseAnalyzer, we could say that every command specifier has an implicit use of `FwCmdOpcodeType`. This may be more preferable since its a step towards providing type checking for these values (would help with #102).
- Store the types in the dictionary analysis. If we are generating a dictionary, then look up the types (ex: analyzer would treat the opcode specified in FPP as FwOpcodeType)

jwest115:
Closes #666.


compiler/lib/src/main/scala/analysis/Analysis.scala
```diff
@@ -70,7 +70,11 @@ case class Analysis(
   /** The dictionary under construction */
   dictionary: Option[Dictionary] = None,
   /** The telemetry packet set under construction */
-  tlmPacketSet: Option[TlmPacketSet] = None
+  tlmPacketSet: Option[TlmPacketSet] = None,
+  /** The dictionary generation mode */
+  dictionaryMode: Boolean = false,
```
bocchino, line 75: Let's rename this variable to be a predicate:
```suggestion
  /** Whether a dictionary is needed in code generation */
  dictionaryNeeded: Boolean = false,
```

compiler/lib/src/main/scala/analysis/CheckSemantics/CheckUses.scala
```diff
@@ -117,6 +117,33 @@ object CheckUses extends UseAnalyzer {
...
+          "FwTlmPacketizeIdType"
+        )
+        val (_, node1, _) = node
+        val mapping = a.nestedScope.get (NameGroup.Type) _
+        for {
+          a <- Result.foldLeft (impliedTypeUses) (a) ((a, t) => {
+            for {
+              symbol <- Result.annotateResult(
+                helpers.getSymbolForName(mapping)(node1.id, t), 
+                s"Dictionary requires type ${t} to be defined.")
```
bocchino, line 137: Let's add another check here to ensure that the type has an integer type as its underlying type. If not, we can return a TypeError saying that an integer type is required.

compiler/lib/src/main/scala/util/Error.scala
```diff
@@ -301,11 +301,16 @@ sealed trait Error {
...
       case XmlError.SemanticError(file, msg) => Error.printXml (file) (msg)
+      case AnnotatedError(error, note) => 
+        error.print
+        System.err.println(s"note: ${note}")
     }
   }
 
 }
 
+/** Annotated error for including additional notes */
```
bocchino, line 312:
```suggestion
/** An error with a note */
```

docs/users-guide/Analyzing-and-Translating-Models.adoc
```diff
@@ -1452,6 +1452,23 @@ The JSON format is specified in the
 https://fprime.jpl.nasa.gov/devel/docs/reference/fpp-json-dict/[F Prime design
 documentation].
 
+When generating dictionaries with `fpp-to-dict`, the following types must be defined
+in FPP (either in the input model itself or in a dependency file):
```
bocchino, line 1456: Let's say also that each of the defined types has to be an alias of an integer type.

bocchino (changes requested):
Looks good! Just a few comments to address.

compiler/lib/src/main/scala/analysis/CheckSemantics/CheckUses.scala
jwest115: Added this check to FinalizeTypeDefs.

thomas-bc:
What's the status here? We should coordinate to make sure we merge/release in parallel with https://github.com/nasa/fprime-gds/pull/199

jwest115:
> What's the status here? We should coordinate to make sure we merge/release in parallel with [nasa/fprime-gds#199](https://github.com/nasa/fprime-gds/pull/199)

I'm wrapping up a few minor changes, should have things committed shortly!

bocchino (approved):
Looks good!

PR #690, opened 2025-05-07
Generate loop for toString() on Struct member arrays

### LINKED ISSUE #689: Omit `toString()` on array larger than max length
There is no size bound on array members inside struct definitions. If the size is very large, the `toString()` becomes boundlessly large and hangs compilation. I think it is reasonable to limit the `toString()` codegen to arrays smaller than `256`. We could probably put in some default format or `...` syntax for arrays larger than a certain size in-place of the `toString()`.

Related to #688.

Kronos3:
This cleans up `toString()` for struct member arrays by concatenating a string to format the array. I ran FppTest in addition to our normal testing to validate the toString() formatting still matches.

Closes #689

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/StructCppWriter.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/StructCppWriter.scala
@@ -402,7 +402,7 @@ case class StructCppWriter(
                       |${fillTmpString}
                       |
                       |FwSizeType size = ${n}Tmp.length() + (i > 0 ? 2 : 0);
-                      |if (size + ${n}Str.length() < ${n}Str.maxLength()) {
+                      |if ((size + ${n}Str.length()) <= ${n}Str.maxLength()) {
                       |  if (i > 0) {
                       |    ${n}Str += ", ";
                       |  }
--- a/compiler/lib/src/main/scala/codegen/CppWriter/StructCppWriter.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/StructCppWriter.scala
@@ -402,8 +402,7 @@ case class StructCppWriter(
                       |${fillTmpString}
                       |
                       |FwSizeType size = ${n}Tmp.length() + (i > 0 ? 2 : 0);
-                      |FwSizeType remaining = ${n}Str.maxLength() - ${n}Str.length();
-                      |if (size < remaining) {
+                      |if (size + ${n}Str.length() < ${n}Str.maxLength()) {
                       |  if (i > 0) {
                       |    ${n}Str += ", ";
                       |  }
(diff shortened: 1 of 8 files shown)


compiler/tools/fpp-to-cpp/test/array/S3SerializableAc.ref.cpp
```diff
@@ -160,17 +160,33 @@ namespace S {
...
       "( "
-      "mU32Array = [ %" PRIu32 ", "
-      "%" PRIu32 ", "
-      "%" PRIu32 " ], "
+      "mU32Array = [ %s ], "
       "mF64 = %f"
       " )";
 
+    // Declare strings to hold any serializable toString() arguments
+    Fw::String mU32ArrayStr;
```
bocchino, line 168: Could we add all the struct members to the main string by concatenation? Then we wouldn't have to declare this temporary string.

compiler/tools/fpp-to-cpp/test/array/S3SerializableAc.ref.cpp
```diff
@@ -160,17 +160,33 @@ namespace S {
...
+
+    // Call toString for arrays and serializable types
+    for (FwSizeType i = 0; i < 3; i++) {
+      Fw::String mU32ArrayTmp;
+      mU32ArrayTmp.format("%" PRIu32 "", this->m_mU32Array[i]);
+
+      FwSizeType size = mU32ArrayTmp.length() + (i > 0 ? 2 : 0);
+      FwSizeType remaining = mU32ArrayStr.maxLength() - mU32ArrayStr.length();
```
bocchino, line 176: `FwSizeType` is unsigned, so we should avoid subtraction here, which can overflow. Let's refactor the arithmetic to use addition.

bocchino (commented):
Looks good! Just a couple of comments.

compiler/tools/fpp-to-cpp/test/array/S3SerializableAc.ref.cpp
Kronos3: We'll still need temporary strings to format each element (with primitive type) in this case. Currently it formats primitive element in one `format`. Is that ok?

bocchino: Good point. I think the code gen here is ok as is. The extra string allocation occurs only in the uncommon case of an array. If we concatenate everything then we allocate a string each time.

compiler/lib/src/main/scala/codegen/CppWriter/StructCppWriter.scala
```diff
@@ -402,8 +402,7 @@ case class StructCppWriter(
                       |${fillTmpString}
                       |
                       |FwSizeType size = ${n}Tmp.length() + (i > 0 ? 2 : 0);
-                      |FwSizeType remaining = ${n}Str.maxLength() - ${n}Str.length();
-                      |if (size < remaining) {
+                      |if (size + ${n}Str.length() < ${n}Str.maxLength()) {
```
bocchino, line 405: Looks good! I think `<` should be `<=`, right?

Kronos3: Updated!

bocchino (approved):
Looks great!

bocchino:
I confirmed that the check-cpp and FppTest tests pass.

PR #691, opened 2025-05-07
Interface Specification

### LINKED ISSUE #697: Add specification for port interfaces
Update the FPP spec to add port interface definitions and their uses in component definitions and in other port interface definitions.

Kronos3:
- Interfaces
~~Interface Templates~~ (~~merged into interfaces~~ We decided to not implement this in this round)
- Interface importing into components
~~- Interface Aliases~~

Closes #697.


Kronos3:
After discussion, I have merged interface templates into interfaces. Now interfaces are what we _were_ calling templates without any parameters.

compiler/lib/src/main/scala/analysis/Analyzers/InterfaceAnalyzer.scala
```diff
@@ -0,0 +1,20 @@
+package analysis.Analyzers
```
bocchino, line 1: I think this file should be removed from this PR.

docs/spec/Definitions/Interface-Alias-Definitions.adoc
```diff
@@ -0,0 +1,38 @@
+=== Interface Alias Definitions
```
bocchino, line 1: I think this file should be removed from this PR.

Kronos3:
Updated the wiki:
- Added interfaces to the analysis homepage
- Check semantics passes updates:
  - Enter symbols
  - Check uses
  - **Added new pass** [Check-Interface-Definitions](https://github.com/nasa/fpp/wiki/Check-Interface-Definitions) (before component definitions pass)
  - Check-Component-Definitions

bocchino (approved):
Looks good!

PR #696, opened 2025-05-13
Fix F' config include path

Brian-Campuzano:
This fixes the include path change that was made in https://github.com/nasa/fprime/pull/3563.

Without this fix I get numerous compilation errors when running `fpp-to-cpp/test/check-cpp`

```sh
compiling AArrayAc.cpp with default guards
In file included from /home/campuzan/projects/fprime/Fw/Types/BasicTypes.hpp:14,
                 from /home/campuzan/projects/fprime/Fw/FPrimeBasicTypes.h:24,
                 from /home/campuzan/projects/fprime/Fw/FPrimeBasicTypes.hpp:21,
                 from /home/campuzan/projects/fprime/Fw/Types/Assert.hpp:4,
                 from AArrayAc.cpp:7:
/home/campuzan/projects/fprime/Fw/Types/BasicTypes.h:41:10: fatal error: config/FPrimeNumericalConfig.h: No such file or directory
   41 | #include <config/FPrimeNumericalConfig.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

### DIFF
--- a/compiler/scripts/fprime-gcc
+++ b/compiler/scripts/fprime-gcc
@@ -2,7 +2,7 @@
 
 # ======================================================================
 # Compile F Prime source files
-# ---------------------------------------------------------------------- 
+# ----------------------------------------------------------------------
 # Setup:
 #
 # 1. Set FPRIME to point to the root of your F Prime working repo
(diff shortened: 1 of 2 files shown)


bocchino:
Thanks, good catch!

bocchino (approved):
Looks good!

PR #705, opened 2025-05-14
Fix issues with type aliasing and formatting

### LINKED ISSUE #704: Type aliases do not interact correctly with format specifiers
If the type associated with a format specifier is an alias type, the implementation should use the underlying type for analysis and code generation. Currently it uses the alias type. This causes the following problems:

1. An alias of an integer type is not recognized as an integer type, so it can't be used with an integer format specifier such as `x`.
2. When generating code for a format corresponding to an alias type, the code generator writes `%s`, which is not correct in general.

- [x] Fix analysis
- [x] Add unit tests for `fpp-check`
- [x] Fix code gen
- [x] Add unit tests for `fpp-to-cpp`

bocchino:
Closes #704.


compiler/lib/src/main/scala/codegen/CppWriter/FormatCppWriter.scala
```diff
@@ -9,38 +9,37 @@ import fpp.compiler.util._
...
-      case (TypeUtils.Signed, w) => s"%\" PRIi${w.toString} \""
+  def getDecimalFormat(pit: Type.PrimitiveInt): String =
+    (pit.signedness, pit.bitWidth) match {
+      case (Type.PrimitiveInt.Unsigned, w) => s"%\" PRIu${w.toString} \""
+      case (Type.PrimitiveInt.Signed, w) => s"%\" PRIi${w.toString} \""
     }
 
   /** Write an FPP format field as a C++ format string */
-  def writeField(f: Format.Field, tn: AstNode[Ast.TypeName]): String = {
+  def writeField(s: CppWriterState, f: Format.Field, tn: AstNode[Ast.TypeName]): String = {
```
jwest115, line 19: Do we need the type name if we are using the type directly from the type map? Maybe we could pass in the field and type instead of state, field, and type name? Just a thought but it also looks good the way it is here!

bocchino: If we did that, then we would have to look up the type every time we called this function. I don't have a strong opinion, but I think it's slightly better to pass in the raw data and look up the type in this function. In particular, now that we have introduced the notion of an "underlying type," looking up the type can be a source of bugs. I'm wondering what other bugs like this are still in the code.

bocchino:
I confirmed that the F Prime unit tests in fprime and fprime/FppTest tests pass with this change.

PR #709, opened 2025-05-14
Interface semantics

### LINKED ISSUE #699: Add semantics for port interfaces
- [ ] Update data structures and algorithms on the wiki
- [ ] Implement
- [ ] Test

Kronos3:
Implements the semantics of FPP Interfaces

- I pulled out common port maps/checking interface GenericPortInterface so that Interface and Component would not have duplicated code.
- I had to rework the C++ header computation a bit since interfaces are not a code-generated symbol like our other symbols. Rather they are compile-time only and refer to other symbols.

Closes #699


Kronos3:
@bocchino I've gone through https://github.com/nasa/fpp/wiki/Adding-New-Features-to-FPP#tools and updated tests accordingly. I'm not quite sure about the changes to the fpp-to-json outputs since there is a new field that is added to import.

Is there a JSON schema I need to update for the `fpp-to-json` output or is that just for the dictionary?

compiler/lib/src/main/scala/analysis/Analysis.scala
```diff
@@ -1,7 +1,9 @@
 package fpp.compiler.analysis
 
-import fpp.compiler.ast._
-import fpp.compiler.util._
+import fpp.compiler.ast.*
```
bocchino: What is the reason for replacing _ with * here?

compiler/lib/src/main/scala/analysis/Analysis.scala
```diff
@@ -59,6 +61,10 @@ case class Analysis(
   componentInstanceMap: Map[Symbol.ComponentInstance, ComponentInstance] = Map(),
   /** The component instance under construction */
   componentInstance: Option[ComponentInstance] = None,
+  /** The map from component symbols to components */
```
bocchino:
```suggestion
  /** The map from interface symbols to interfaces */
```

compiler/lib/src/main/scala/analysis/Analysis.scala
```diff
@@ -292,7 +318,7 @@ case class Analysis(
       ))
   }
 
-  /** Gets a nonnegative int value from an AST node */
+  /** Gets a non-negative int value from an AST node */
```
bocchino: Is this the result of an automated spell check? nonnegative is a standard word in math.

compiler/lib/src/main/scala/analysis/CheckSemantics/CheckUseDefCycles.scala
```diff
@@ -44,6 +44,16 @@ object CheckUseDefCycles extends UseAnalyzer {
     visitDefPost(a, symbol, node, super.defTopologyAnnotatedNode)
   }
 
+  override def defComponentAnnotatedNode(a: Analysis, node: Ast.Annotated[AstNode[Ast.DefComponent]]) = {
```
bocchino: Is this addition needed? How can a component definition be involved in a cycle?

compiler/lib/src/main/scala/codegen/CppWriter/CppWriterState.scala
```diff
@@ -132,42 +132,54 @@ case class CppWriterState(
...
+            getIncludePath(sym, ComputeCppFiles.FileNames.getStruct(name))
           )
-          case _: Symbol.Topology => Some(
-            ComputeCppFiles.FileNames.getTopology(name)
+          case _: Symbol.Topology => List(
+            getIncludePath(sym, ComputeCppFiles.FileNames.getTopology(name))
           )
-          case _ => None
+          case iface: Symbol.Interface =>
+            // Compute the symbols that this inteface depends on
```
bocchino: I don't understand this comment. Why do we need to compute used symbols here?

compiler/lib/src/main/scala/analysis/CheckSemantics/CheckUseDefCycles.scala
Kronos3: Ah right good point, I thought I'd need it to hit the interface inside but there's no way to loop back to the original component.

compiler/lib/src/main/scala/codegen/CppWriter/CppWriterState.scala
Kronos3: I believe so because the `UsedSymbols` pass only collects a shallow use of the interface not the ports embedded. We need to compute the deep port uses in the interface.

bocchino: OK, that makes sense. Could you use `UsedSymbols.resolveUses` for this? That function recursively resolves uses. If it works, it might be slightly cleaner.

Kronos3: Ah yeah good idea, works like a charm

bocchino:
> Is there a JSON schema I need to update for the fpp-to-json output or is that just for the dictionary?

There is no schema for the fpp-to-json output. Most of it is auto-generated by circe.

PR #710, opened 2025-05-14
Interface testing

### LINKED ISSUE #700: Implement and test code generation for port interfaces
Extend the component code generation to generate code for ports imported from interfaces.

- [ ] Implement
- [ ] Test

Kronos3:
Implements the tests for interfaces. Added interface tests where appropriate and converted the .fppi files in the component C++ tests into real interfaces. Since they are effectively the same, no changes were made to the final codegen-ed C++.

Closes #700


bocchino:
@Kronos3 Can you merge interface-semantics into this branch and fix up the unit tests? When I tried the merge, I saw some test failures in fpp-to-cpp. I think the failures are related to external parameters, which are a recent addition to fpp/main.

bocchino (approved):
Very nice! This not only tests the import mechanism, but cleans up the C++ tests.

bocchino:
It looks like 3 of the unit tests are failing. Please fix, and then we can merge!

Kronos3:
Looks like it's due to https://github.com/nasa/fpp/pull/709/commits/4fad4b0afdb9025717a27ce065921f0ddd2b23b2. There's a stack overflow since it infinitely recurses on the interface symbol. When I filter out that symbol I realize there is another issue with doing it this https://github.com/nasa/fpp/pull/709#discussion_r2125059655 way.

This computes a deep dependency on the interface triggering additional includes that are implicit:

```diff
--- PassiveCommandsComponentAc.ref.hpp        2025-06-06 09:42:30
+++ PassiveCommandsComponentAc.hpp      2025-06-06 10:28:28
@@ -8,25 +8,45 @@
 #define PassiveCommandsComponentAc_HPP
 
 #include "AArrayAc.hpp"
+#include "AliasAliasArrayAliasAc.hpp"
+#include "AliasArrayAliasAc.hpp"
+#include "AliasArrayAliasArrayAliasAc.hpp"
+#include "AliasBoolAliasAc.hpp"
+#include "AliasEnumAliasAc.hpp"
+#include "AliasPrim1AliasAc.hpp"
+#include "AliasPrim2AliasAc.hpp"
+#include "AliasStringAliasAc.hpp"
+#include "AliasStructAliasAc.hpp"
 #include "AliasTypedPortAc.hpp"
 #include "AliasTypedReturnPortAc.hpp"
 #include "AliasTypedReturnStringPortAc.hpp"
+#include "AnotherAliasStructAliasAc.hpp"
+#include "ArrayAliasArrayArrayAc.hpp"
 #include "EEnumAc.hpp"
+#include "Fw/Cmd/CmdArgBuffer.hpp"
 #include "Fw/Cmd/CmdPortAc.hpp"
 #include "Fw/Cmd/CmdRegPortAc.hpp"
+#include "Fw/Cmd/CmdResponseEnumAc.hpp"
 #include "Fw/Cmd/CmdResponsePortAc.hpp"
 #include "Fw/Cmd/CmdString.hpp"
 #include "Fw/Comp/ActiveComponentBase.hpp"
 #include "Fw/FPrimeBasicTypes.hpp"
+#include "Fw/Log/LogBuffer.hpp"
 #include "Fw/Log/LogPortAc.hpp"
+#include "Fw/Log/LogSeverityEnumAc.hpp"
 #if FW_ENABLE_TEXT_LOGGING == 1
 #include "Fw/Log/LogTextPortAc.hpp"
 #endif
+#include "Fw/Log/TextLogString.hpp"
 #include "Fw/Port/InputSerializePort.hpp"
 #include "Fw/Port/OutputSerializePort.hpp"
+#include "Fw/Prm/ParamBuffer.hpp"
+#include "Fw/Prm/ParamValidEnumAc.hpp"
 #include "Fw/Prm/PrmGetPortAc.hpp"
 #include "Fw/Prm/PrmSetPortAc.hpp"
+#include "Fw/Time/Time.hpp"
 #include "Fw/Time/TimePortAc.hpp"
+#include "Fw/Tlm/TlmBuffer.hpp"
 #include "Fw/Tlm/TlmPortAc.hpp"
 #include "NoArgsAliasStringReturnPortAc.hpp"
 #include "NoArgsPortAc.hpp"
@@ -34,8 +54,13 @@
 #include "NoArgsStringReturnPortAc.hpp"
 #include "Os/Mutex.hpp"
 #include "SSerializableAc.hpp"
+#include "StructWithAliasSerializableAc.hpp"
 #include "TypedPortAc.hpp"
 #include "TypedReturnPortAc.hpp"
+#include "config/FwChanIdTypeAliasAc.hpp"
+#include "config/FwEventIdTypeAliasAc.hpp"
+#include "config/FwOpcodeTypeAliasAc.hpp"
+#include "config/FwPrmIdTypeAliasAc.hpp"
```

Might be better to revert back to:

```scala
val Right(a) = UsedSymbols.defInterfaceAnnotatedNode(
  this.a,
  this.a.interfaceMap(iface).aNode,
)
a.usedSymbolSet.flatMap(getIncludeFiles).toList
```

bocchino:
> Might be better to revert back to:

That makes sense! Go ahead and revert.

compiler/lib/src/main/scala/codegen/CppWriter/CppWriterState.scala
```diff
@@ -162,10 +162,13 @@ case class CppWriterState(
             getIncludePath(sym, ComputeCppFiles.FileNames.getTopology(name))
           )
           case iface: Symbol.Interface =>
-            // Recursively resolve uses to pick up the header dependencies
+            // Resolve the uses to pick up the header dependencies
             // implied by imported interfaces
-            UsedSymbols.resolveUses(this.a, Set(iface))
-              .flatMap(getIncludeFiles).toList
+            val Right(a) = UsedSymbols.defInterfaceAnnotatedNode(
```
bocchino, line 167: I think I understand the way this works now. You are using the semantic interface map, which is already resolved to store all the interfaces transitively imported by this interface. So the used symbols of those interfaces should pick up all the recursive include dependencies. Do we have a test covering the recursive case? That is, component C imports interface I which imports interface I'? If not, can we add such a test?

Kronos3: Good idea! I split up `TypedPorts` into a recursive set of port groups to test this case.

bocchino: Great! Thanks. CI is passing, so I will go head and merge it.

PR #716, opened 2025-05-15
Array toString() with loop

### LINKED ISSUE #712: toString for arrays uses too much stack space
This is the same issue that we fixed in #690, but for array types. We should not allocate an `Fw::String` for every array element. This probably works on most systems, but it is unnecessary, and it fails on small systems.

Kronos3:
Closes #712 

See https://github.com/nasa/fprime/pull/3626 for FPPTest updates


bocchino (approved):
Looks good! I ran

* `fpp-check` in this branch against `nasa/fprime/devel`

*  FppTest in `Kronos3/fpp-reformat-arrays` against this branch

Both passed.

PR #720, opened 2025-05-20
Turn off tracing

### LINKED ISSUE #714: Remove generation of trace files from CI and release process
The CI process and the release process generate trace files and attempt to use them when running the native build. However, as far as I can tell, these trace files have no effect on the native build. The native build seems to use only the trace files that are checked into the FPP repo.

Can we please remove the generation of trace files from CI and the release process? I think manual management of these files is working fine, and removing the trace generation step would speed up the process. Especially for CI, this process is becoming a developer bottleneck, as the native builds take a long time, and they run on every merge to main or to a PR to main.

LeStarch:
Turns off tracing with the `trace:` flag.

Closes #714

### DIFF
--- a/.github/workflows/build-native.yml
+++ b/.github/workflows/build-native.yml
@@ -18,4 +18,5 @@ jobs:
       output-directory: ./compiler/bin
       meta-package: fpp
       fast-hack: true
+      trace: false
     secrets: inherit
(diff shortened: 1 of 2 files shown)


bocchino (approved):
Looks great, thanks! This will speed up the PR and release process.

Issue #719, opened 2025-05-19
Add DP types to dictionary required types


jwest115:
Include additional types (like those defined in DP header) to implied type use check:
- FwPacketDescriptorType
- FwDpIdType
- FwDpPriorityType
- Identify the constants that need to be there, require they are there, then dump them to the dictionary
    - Container user data size
    - Dp state serialized size
    - Time serialized size

PR #729, opened 2025-06-03
Require DP constants and enums in dictionary

jwest115:
Require that the following types are defined and output them to the dictionary:

**Aliases**
- FwDpIdType
- FwDpPriorityType
- FwSizeStoreType
- FwTimeBaseStoreType
- FwTimeContextStoreType

**Enums**
- Fw.DpState
- Fw.DpCfg.ProcType

**Constants**
- Fw.DpCfg.CONTAINER_USER_DATA_SIZE

Closes https://github.com/nasa/fpp/issues/719.

Spec updates are done in https://github.com/nasa/fprime/pull/3692.

### DIFF
--- a/compiler/lib/src/main/scala/analysis/Analysis.scala
+++ b/compiler/lib/src/main/scala/analysis/Analysis.scala
@@ -73,8 +73,8 @@ case class Analysis(
   tlmPacketSet: Option[TlmPacketSet] = None,
   /** Whether a dictionary is needed in code generation */
   dictionaryNeeded: Boolean = false,
-  /** The set of alias type symbols used by the dictionary */
-  dictionaryAliasSymbolSet: Set[Symbol] = Set(),
+  /** The set of type symbols used by the dictionary */
+  dictionaryTypeSymbolSet: Set[Symbol] = Set(),
   /** The set of constant symbols used by the dictionary */
   dictionaryConstantSymbolSet: Set[Symbol] = Set(),
   /** The set of enum symbols used by the dictionary */
--- a/compiler/lib/src/main/scala/analysis/Analysis.scala
+++ b/compiler/lib/src/main/scala/analysis/Analysis.scala
@@ -73,8 +73,12 @@ case class Analysis(
   tlmPacketSet: Option[TlmPacketSet] = None,
   /** Whether a dictionary is needed in code generation */
   dictionaryNeeded: Boolean = false,
-  /** The set of type symbols used by the dictionary */
-  dictionaryTypeSymbolSet: Set[Symbol] = Set()
+  /** The set of integer symbols used by the dictionary */
+  dictionaryIntegerSymbolSet: Set[Symbol] = Set(),
+  /** The set of constant symbols used by the dictionary */
+  dictionaryConstantSymbolSet: Set[Symbol] = Set(),
+  /** The set of enum symbols used by the dictionary */
+  dictionaryEnumSymbolSet: Set[Symbol] = Set()
 ) {
 
   /** Gets the qualified name of a symbol */
--- a/compiler/lib/src/main/scala/analysis/Analysis.scala
+++ b/compiler/lib/src/main/scala/analysis/Analysis.scala
@@ -73,8 +73,8 @@ case class Analysis(
   tlmPacketSet: Option[TlmPacketSet] = None,
   /** Whether a dictionary is needed in code generation */
   dictionaryNeeded: Boolean = false,
-  /** The set of integer symbols used by the dictionary */
-  dictionaryIntegerSymbolSet: Set[Symbol] = Set(),
+  /** The set of alias type symbols used by the dictionary */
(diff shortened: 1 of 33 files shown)


compiler/lib/src/main/scala/analysis/Analysis.scala
```diff
@@ -73,8 +73,12 @@ case class Analysis(
   tlmPacketSet: Option[TlmPacketSet] = None,
   /** Whether a dictionary is needed in code generation */
   dictionaryNeeded: Boolean = false,
-  /** The set of type symbols used by the dictionary */
-  dictionaryTypeSymbolSet: Set[Symbol] = Set()
+  /** The set of alias type symbols used by the dictionary */
+  dictionaryAliasSymbolSet: Set[Symbol] = Set(),
```
bocchino: I think we should call this TypeSymbolSet instead of AliasSymbolSet. The type definitions happen to be aliases, but other type definitions are possible, and we may want to add them later.

compiler/lib/src/main/scala/analysis/CheckSemantics/CheckUses.scala
```diff
@@ -120,25 +120,64 @@ object CheckUses extends UseAnalyzer {
   override def defTopologyAnnotatedNode(a: Analysis, node: Ast.Annotated[AstNode[Ast.DefTopology]]) = {
     a.dictionaryNeeded match {
       case true => {
-        val impliedTypeUses = List(
-          "FwChanIdType", 
-          "FwEventIdType", 
-          "FwOpcodeType", 
-          "FwPacketDescriptorType", 
-          "FwTlmPacketizeIdType"
+        val impliedAliasTypeUses = List(
```
bocchino: I think we should call this impliedTypeUses.

compiler/lib/src/main/scala/analysis/CheckSemantics/ConstructDictionaryMap.scala
```diff
@@ -24,17 +32,8 @@ object ConstructDictionaryMap
...
-        then Right(a)
-        else Left(
-          SemanticError.InvalidType(
-            loc,
-            s"this F Prime framework type must be an alias of an integer type"
-          )
-        )
-      })
+      a <- checkIntegerType(a.dictionaryAliasSymbolSet, "this F Prime framework type must be an alias of an integer type")
+      a <- checkIntegerType(a.dictionaryConstantSymbolSet, "this F Prime framework type must be a constant with a value of integer type")
```
bocchino:
```suggestion
      a <- checkIntegerType(a.dictionaryConstantSymbolSet, "this F Prime constant must have an integer type")
```

jwest115:
Going to update the user guide with info on the types that are required by the dictionary.

docs/users-guide/Analyzing-and-Translating-Models.adoc
```diff
@@ -1478,9 +1481,35 @@ types:
...
+* `Fw.DpState`
+* `Fw.DpCfg.ProcType`
+
+Constants:
+
+* `Fw.DpCfg.CONTAINER_USER_DATA_SIZE`
+
+The alias types and constants required by the dictionary must all be integer types.
+
+Type information for constant dictionary entries is determined by
```
bocchino: I think this is too much information for the FPP User's Guide. It's an implementation detail about how integer types are represented in the JSON dictionary spec. Can we move this information to the JSON dictionary spec?

jwest115: Done!

compiler/lib/src/main/scala/analysis/CheckSemantics/CheckUses.scala
```diff
@@ -121,24 +121,63 @@ object CheckUses extends UseAnalyzer {
...
-        val mapping = a.nestedScope.get (NameGroup.Type) _
-        for {
-          a <- Result.foldLeft (impliedTypeUses) (a) ((a, t) => {
+        def checkImpliedUses(
+          impliedUses: List[List[String]],
+          id: AstNode.Id,
+          typeName: String,
+          ng: NameGroup
+        ): Result.Result[Set[Symbol]] = {
+          Result.foldLeft (impliedUses) (Set[Symbol]()) ((s, identList) => {
```
bocchino, line 150: I have one more change to suggest here. Instead of using `Set[Symbol]` to represent these concepts, can we use `Set[Symbol.Type]`, `Set[Symbol.Enum]`, and `Set[Symbol.Constant]`? That would be more precise. And can we could throw an error if, for example, the symbol `Fw.DpState` is defined, but it's not an enum symbol? If we don't do this, I'm concerned we won't have enough error checking. For example, if `Fw.DpState` were a struct type definition or even a component definition, this check would pass.

bocchino: Superseded by #740.

jwest115:
Created new PR for this change https://github.com/nasa/fpp/pull/742

PR #733, opened 2025-06-05
Rework struct toString()

### LINKED ISSUE #722: toString() for structs uses too much stack space
This is the same issue as #712, but for structs (not arrays in structs, but plain structs). Example input:

```
    struct MyComplicatedStruct {
        entryA : MyEnum
        entryB : MyEnum
        entryC : MyEnum
        entryD : MyEnum
        entryE : MyEnum
        entryF : MyEnum
        entryG : MyEnum
        entryH : MyEnum
        entryI : MyEnum
        entryJ : MyEnum
    }
```

Example output:

```
#if FW_SERIALIZABLE_TO_STRING

  void MyComplicatedStruct ::
    toString(Fw::StringBase& sb) const
  {
    static const char* formatString =
      "( "
      "entryA = %s, "
      "entryB = %s, "
      "entryC = %s, "
      "entryD = %s, "
      "entryE = %s, "
      "entryF = %s, "
      "entryG = %s, "
      "entryH = %s, "
      "entryI = %s, "
      "entryJ = %s"
      " )";

    // Declare strings to hold any serializable toString() arguments
    Fw::String entryAStr;
    Fw::String entryBStr;
    Fw::String entryCStr;
    Fw::String entryDStr;
    Fw::String entryEStr;
    Fw::String entryFStr;
    Fw::String entryGStr;
    Fw::String entryHStr;
    Fw::String entryIStr;
    Fw::String entryJStr;

    // Call toString for arrays and serializable types
    this->m_entry00.toString(entryAStr);
    this->m_entry01.toString(entryBStr);
    this->m_entry02.toString(entryCStr);
    this->m_entry03.toString(entryDStr);
    this->m_entry04.toString(entryEStr);
    this->m_entry05.toString(entryFStr);
    this->m_entry06.toString(entryGStr);
    this->m_entry07.toString(entryHStr);
    this->m_entry08.toString(entryIStr);
    this->m_entry09.toString(entryJStr);

    sb.format(
      formatString,
      entryAStr.toChar(),
      entryBStr.toChar(),
      entryCStr.toChar(),
      entryDStr.toChar(),
      entryEStr.toChar(),
      entryFStr.toChar(),
      entryGStr.toChar(),
      entryHStr.toChar(),
      entryIStr.toChar(),
      entryJStr.toChar()
    );
  }

#endif
```

This results in the following error in our codebase:

```
[build] ld.lld: error: [...]MyComplicatedStructSerializableAc.cpp:234:0: stack frame size (3016) exceeds limit (1800) in function '[...]'
```

It should not take 3 KiB of stack to convert a 10-element struct to a string.

Kronos3:
This code cleans up the `toString()` on structs which as a side benefit reduces the stack memory required to format the struct into a string (Closes #722).

**Before**
Format each individual member into it's own string on the stack and then use a big old `Fw::String.format()` call with a predefined format string to glob all the members into a single string.

**Now**
Move through each member in the string and append the formatted member to the string. There is optionally a single `Fw::String tmp` which members will use to format into if needed before appending to the output.

- [x] Test with FppTest
- [x] Test with `check-cpp`


compiler/tools/fpp-to-cpp/test/array/S3SerializableAc.ref.cpp
```diff
@@ -158,36 +158,27 @@ namespace S {
...
+      sb += tmp;
     }
-
-    sb.format(
-      formatString,
-      mU32ArrayStr.toChar(),
-      this->m_mF64
-    );
+    sb += "] ";
+    sb += ", ";
```
Kronos3: @bocchino I've noticed a bug in my own code through inspection that _should_ have been caught by FppTest but wasn't.

Line `174` should be `" ]"` _not_ `"] "` (notice the space position). I dug deeper and it turns out it's because FppTest is truncating the relevant part of the string and never checks it.

In fact, the test string in FppTest is also buggy! https://github.com/nasa/fprime/blob/426ba869774b8d5dd79735902486fa3efdb45098/FppTest/struct/NonPrimitiveStructTest.cpp#L341-L342

The test code would add an additional space between `]` and `)`.

We should probably add a test case to FppTest that places array members earlier in the struct to avoid cut off and therefore test the spacing on array members.

Issue #702, opened 2025-05-13
Integrate and test port interfaces with F Prime


bocchino:
- [x] Add test cases to FppTest
- [x] Update `.fppi` files in FPrime core fpp components to hard interfaces
- [x] Make a PR to nasa/fprime/devel

PR #736, opened 2025-06-09
Fix interface locs

Kronos3:
I found this issue which working on https://github.com/nasa/fpp/issues/702. The generated locs file could not be parsed back into the model since I didn't update the parser. This was the only FPP change required to get FppTest passing


Kronos3:
@bocchino Should we add a test for this? I think it'll be some sort of cross tool check since we need to generate the locs and then parse it. It'll also always be caught by the Fprime core integration so it may not be 100% necessary.

bocchino:
Looks good, but it seems we never added a test for interfaces to fpp-depend. Can you add one here?

https://github.com/nasa/fpp/tree/main/compiler/tools/fpp-depend/test

bocchino (approved):
Looks good!

PR #737, opened 2025-06-10
Cast Fw types in C++ codegen

jwest115:
Related to https://github.com/nasa/fprime/pull/3714, in C++ codegen, cast opcodes, event IDs, channel IDs, etc. to their respective types (ie: FwOpcodeType, FwEventIdType, FwChanIdType, etc.). This helps avoid implicit type conversion warnings that occur when building.

Partially addresses https://github.com/nasa/fprime/issues/3529 (fprime updates are included in https://github.com/nasa/fprime/pull/3714).

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCommands.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCommands.scala
@@ -67,7 +67,7 @@ case class ComponentCommands (
           sortedCmds.map((_, cmd) =>
             lines(
               s"""|this->${portVariableName(cmdRegPort.get)}[0].invoke(
-                  |  this->getIdBase() + ${commandConstantName(cmd)}
+                  |  static_cast<FwOpcodeType>(this->getIdBase()) + ${commandConstantName(cmd)}
                   |);
                   |"""
             )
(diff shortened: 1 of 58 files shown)


bocchino (commented):
This PR looks good and correct, but I thought we were moving towards having a single ID type for all these IDs. If we add all these casts, won't we just need to delete them later? Does anything go wrong now if we omit these casts?

jwest115:
I ran without these changes and tried the case where FwChanIdType is smaller (U16) than the FwIdType (U32) and received the implicit conversion warnings. Currently in fprime, all of the id types are aliases of the FwIdType, maybe that is sufficient? However, if a user configures any of those id types to something different they will hit the warning.

jwest115:
@LeStarch do you think there is a use case for configuring an ID type to something smaller than the base ID type (ie: FwIdType)? Maybe casting the IDs is not a good idea since it would in a way "suppress" the implicit conversion warnings which might not alert users that there IDs are smaller than the base ID?

jwest115:
In [this](https://github.com/nasa/fprime/pull/3714) PR we introduced a new alias `FwIdType`. All of the IDs are aliases of that type, however, a user could configure them to be something different.

jwest115:
Closing since we decided to not add casts at this time.

PR #740, opened 2025-06-12
Revise handling of implied uses in topologies

### LINKED ISSUE #738: Don't bypass the UseAnalyzer when checking implied uses in topologies
In #686, we added checks for implied uses of type aliases in topologies in an ad-hoc way. In hindsight this approach is not correct, because it bypasses the UseAnalyzer. In particular, the dependency analyzer, which gets its use information from the UseAnalyzer, is not told about these implied uses.

Instead of doing this, I think we should do the following:

1. Represent the implied uses as AST nodes
2. When visiting a topology, visit those nodes in the ordinary way for uses

bocchino:
Closes #738.

### DIFF
--- a/compiler/lib/src/main/scala/ast/AstNode.scala
+++ b/compiler/lib/src/main/scala/ast/AstNode.scala
@@ -8,10 +8,16 @@ object AstNode {
   /** A node identifier */
   type Id = Int
 
+  /** Get a fresh identifier */
+  def getId: Id = {
+    val id0 = id
+    id = id + 1
+    id0
+  }
+
   /** Create a new node with a fresh identifier */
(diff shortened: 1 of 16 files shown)


compiler/lib/src/main/scala/ast/AstNode.scala
```diff
@@ -8,10 +8,16 @@ object AstNode {
   /** A node identifier */
   type Id = Int
 
+  /** Get a fresh identifier */
+  def getId: Id = {
+    val id0 = id
+    id = id + 1
+    id0
```
jwest115, line 15: Why is `id0` returned here instead of the new id (id + 1)?

jwest115 (approved):
I have one question but looks good! Thank you!

PR #742, opened 2025-06-18
Add DP implied uses to dictionary

jwest115:
Require that the following types are defined and output them to the dictionary:

**Aliases**
- FwDpIdType
- FwDpPriorityType
- FwSizeStoreType
- FwTimeBaseStoreType
- FwTimeContextStoreType

**Enums**
- Fw.DpState
- Fw.DpCfg.ProcType

**Constants**
- Fw.DpCfg.CONTAINER_USER_DATA_SIZE

Closes https://github.com/nasa/fpp/issues/719.


compiler/lib/src/main/scala/analysis/Semantics/ImpliedUse.scala
```diff
@@ -11,6 +11,15 @@ case class ImpliedUse(
   id: AstNode.Id
 ) {
 
+  def asExprNode: AstNode[Ast.Expr] = {
+    val Name.Qualified(qualifier, base) = name
+    val head :: tail = qualifier :+ base
```
bocchino:
```suggestion
    val head :: tail = name.toIdentList
```

compiler/tools/fpp-to-dict/test/top/arrayFwEventIdType.fpp
```diff
@@ -1,8 +1,47 @@
-type FwPacketDescriptorType = U32
-type FwOpcodeType = U32
 type FwChanIdType = U32
-type FwTlmPacketizeIdType = U16
+type FwDpIdType = U32
```
bocchino: There is some duplicated code in this and other tests that I think is not necessary. I think we could simplify the tests by (1) moving all the checks for invalid framework types to fpp-check and (2) using the same set of (correct) framework types for all fpp-to-json tests.

jwest115: Sounds good, I will make that change.

bocchino: In the context of fpp-check the tests for incorrect framework types and constants can be one-liners, e.g., `struct FwOpcodeType {}`.

jwest115: After looking into this more, I think we will want to keep some of the tests for checking whether a framework type is defined as part of the fpp-to-dict tests. Do you think that is something we should keep?

For example, if the `FwOpcodeType` is not defined and we are generating a dictionary, we'll want to test that we get the expected error.

bocchino: Yes, I agree. The tests for missing implied topology uses have to live in the fpp-to-dict tests, because we don't currently have a way to run fpp-check in dictionary generation mode. However, the tests for checking framework defs (if they exist) for correctness can go into fpp-check. Eventually we may want to add a dictionary-mode flag to fpp-check, but I think it's outside the scope of this PR.

jwest115: Agreed!

compiler/tools/fpp-to-dict/test/top/missingUserDataSizeConstant.fpp
```diff
@@ -0,0 +1,45 @@
+type FwChanIdType = U32
```
jwest115, line 1: Simplify this, take out annotations and simplify enum defs

bocchino (approved):
Looks good!

PR #748, opened 2025-06-18
Topology Port Export Spec

Kronos3:
My initial pass at a spec for exporting topology ports.

TODO:
- [ ] Update wiki

### DIFF
--- a/docs/spec/Definitions/Topology-Definitions.adoc
+++ b/docs/spec/Definitions/Topology-Definitions.adoc
@@ -31,6 +31,8 @@ instance specifier>>
 
 * A <<Specifiers_Topology-Import-Specifiers,topology import specifier>>
 
+* A <<Specifiers_Port-Export-Specifiers,topology port export specifier>>
+
 * An <<Specifiers_Include-Specifiers,include specifier>>
 
 ==== Semantics
@@ -138,7 +140,7 @@ of the topology to a set of packets.
 FPP automatically assigns port numbers as follows.
 
 *Check output port numbers:*
-At each output port _p_ of each component instance _I_, check that
+At each output port _p_ of each component instance (or topology) _I_, check that
 
 . The number of connections is in bounds for the
 size of _I_ `.` _p_.
@@ -157,7 +159,7 @@ a.p[0] -> c.p
 Assign matching numbers to matched pairs of ports.
 For each instance _I_ in the topology:
 
-. Let _C_ be the component type of _I_.
+. Let _C_ be the component type or a topology type of _I_.
 
 .  For each
 <<Specifiers_Port-Matching-Specifiers,port matching specifier>>
@@ -241,10 +243,14 @@ For purposes of port numbering, FPP orders connections as follows.
 A *connection endpoint* is _I_ `.` _p_ or _I_ `.` _p_ `[` _n_ `]`, where
 
 * _I_ refers to a
-<<Definitions_Component-Instance-Definitions,component instance>>.
+<<Definitions_Component-Instance-Definitions,component instance>> or
+an <<Specifiers_Topology-Import_Specifiers,imported>> <<Definitions_Topology-Definitions,topology definition>>.
 
-* _p_ is an identifier that names a port instance specified in
+* If _I_ is a component instance, _p_ is an identifier that names a port instance specified in
 the component definition associated with _I_.
+
+* If _I_ is topology definition, _p_ is an identifier that names an <<Specifiers_Port-Export-Specifiers,exported port>> inside
+_I_. _I_ must be <<Specifiers_Topology-Import_Specifiers,imported>> into the parent topology of the connection.
 
 * _n_ is an optional port number that is present if and only
 if it appears in the model source.
--- a/docs/spec/Specifiers/Connection-Graph-Specifiers.adoc
+++ b/docs/spec/Specifiers/Connection-Graph-Specifiers.adoc
@@ -92,12 +92,16 @@ unmatched.
 
 .. For each of the two port instance identifiers _I_ appearing in _C_:
 
-... The component instance named in _I_
+... If the parent definition of _I_ is a component instance, the component instance
 must be available in the enclosing topology,
 either through
 <<Specifiers_Component-Instance-Specifiers,direct specification>>
 or through
 <<Specifiers_Topology-Import-Specifiers,import>>.
+
+... If the parent definition of _I_ is a topology definition, the topology
+must be imported through a <<Specifiers_Topology-Import-Specifiers,topology import>> and
+the port instance _I_ must be exported through a <<Specifiers_Port-Export-Specifiers,port export>>.
 
 ... The optional expression _e_ following the identifier, if it is present,
 represents a port number, i.e., an index into an
--- a/docs/spec/Specifiers/Port-Export-Specifiers.adoc
+++ b/docs/spec/Specifiers/Port-Export-Specifiers.adoc
@@ -1,0 +1,74 @@
+=== Port Export Specifiers
+
+A port export specifier maps a <<Port-Instance-Specifiers,port instance>>
+out to a port instance for the parent <<Definitions_Topology-Definitions,topology definition>>.
+
+==== Syntax
+
+`port`
+<<Component-Instance-Member-Identifiers_Port-Instance-Identifiers,_port-instance-id_>>
+_[_ `[` <<Expressions,_expression_>> `]` _]_
+
+_[_
+`as`
+<<Lexical-Elements_Identifiers,_identifier_>>
+_]_
+
+==== Semantics
+
+. The port instance identifier references a <<Port-Instance-Specifiers,port instance>>
+inside a component instance named in the enclosing topology _T_.
+
+.. The optional expression _e_ following the port instance identifier, if it is present,
+represents a port number, i.e., an index into an
+array of port instances.
+The type of _e_ must be a
+<<Types_Internal-Types_Numeric-Types,numeric type>>, and
+_e_ must
+<<Evaluation,evaluate>> to a compile-time constant
+that becomes a non-negative integer _n_ when
+<<Evaluation_Type-Conversion,converted to>> type _Integer_.
+_n_ must lie in the range [0,2^31^) and must be in bounds for the
+array size of the port instance specifier named in _I_.
+
+. The optional identifier _I2_, if provided, will rename the port being exported.
+If not provided, the name of the port being exported will match the name of the port instance identifier.
+
+. The port referenced must be either a
+<<Specifiers_Port-Instance-Specifiers,output port instance>>
+or an
+<<Specifiers_Port-Instance-Specifiers,input port instance>>.
+
+.. Exporting an input or output port will expose a port instance of the same
+type to the enclosing topology _T_. Any connections made on this exposed port will
+internally be mapped to the port instance identifier _I_.
+
+==== Example
+
+[source,fpp]
+----
+topology T {
+    instance c1
+    instance c2
+
+    connections {
+        # [...] Internal connections
+    }
+
+    export c1.portOut
+    export c2.portIn as customPortName
+}
+
+topology Y {
+    import T
+
+    instance c3
+    connections {
+        T.portOut -> c3.portIn
+        c3.portOut -> T.customPortName
+    }
+}
+----
+
+See also the <<Definitions_Topology-Definitions_Examples,examples for topology
+definitions>>.


Kronos3:
Some open questions for @bocchino 

- Currently this spec states that you must `import` a topology to be able to connect to ports exported from the topology. Thinking ahead to sub-topologies, I'm not sure if this is right. I think that composing topologies should potentially be done another way:

**Without re-usable sub-topologies**
```
topology T1 {
 instance i

 export i.p[0] as schedIn
}

topology T2 {
  # Not sure of an exact syntax, but something to denote that we are bringing in a topology without doing a full white-box import
  topology T1   
}
```

**With re-usable subtopologies**
```
topology T1(constant baseId) {
  instance i1: Component base id baseId + 0
  instance i2: Component base id baseId + 0x20

  connections {
    # ... internal connections ...
  }

  export i1.port as externalPort1
  export i2.port as externalPort2
}

topology T2 {
  topology t1 = T1(0xA)
  topology t2 = T1(0xB)

  connections {
    t1.externalPort1 -> t2.externalPort2
  }
}
```

Kronos3:
Notes from our meeting today:

```
topology T1 {
  instance I2
  export I2.port as schedOut
}

topology T2 { 
  instance I1
  import T1

  connections {
     # flatten + connect
     I2.p1 -> I1.p2
  }

  connections {
     # connect + flatten
     T1.schedOut -> I1.p2
  }
}
```

Notes:
- Export in topology can _name_ a connection endpoint
- Endpoint can be either a port instance of a component instance or an exported port of a *directly* imported topology

TODO:
- Updated the algorithm for topology resolution to include exported ports of imported topologies in the set of endpoints

Kronos3:
Some breaking changes we identified
- Telemetry packets on private instances
  - We don't want to allow referring to a private instance's telemetry channels in a "parent" topology because that would defeat the purpose of private component names
  - We'd need some mechanism to define packets inside the sub-topology but that breaks the current paradigm of defining all the packets in the outermost topology.
  - **There doesn't seem to be a good resolution to this at the moment**
- Pattern graph resolution: How to connect to "private" `cmdDisp`
  - We do this connection _after_ all imported subtopologies are flattened.
  - To refer to a _private_ `cmdDisp` instance inside a sub-topology, you can just name the sub topology and then the connections will be made to a `cmdDisp` component inside of the refered topology
- ~~Qualfiers on command mnemonics~~ This one is actually fine

docs/spec/Interfaces.adoc
```diff
@@ -0,0 +1,91 @@
+== Interfaces
```
bocchino: Can we rename this section to "Port Interfaces" and consistently use the term "port interface"? I think it's OK to use "interface" as a shorthand for "port interface" from time to time, but I think we shouldn't do that too often.

docs/spec/Interfaces.adoc
```diff
@@ -0,0 +1,91 @@
...
+<<Specifiers_Interface-Import-Specifiers,interface import>>).
+
+==== Anonymous Topology Interfaces
+
+A <<Definitions_Topology-Definitions,topology definition>>
+will create a corresponding anonymous interface. This interface includes
+the set of all <<Specifiers_Topology-Port-Specifiers,topology port specifiers>>
+inside the topology definition.
+
+=== Instance Values
```
bocchino: Can we avoid using the term "value" here? I don't think there are any values that are computed or used at runtime. Instead, can we use the term "port interface instance" for an instance of a port interface? A port interface instance can be a component instance, a topology, or an interface parameter.

docs/spec/Instance-Member-Identifiers.adoc
```diff
@@ -1,14 +1,13 @@
-== Component Instance Member Identifiers
+== Instance Member Identifiers
```
bocchino, line 1: This section is tricky to rewrite because we need to generalize component instances (which can have ports or telemetry channels) into two kinds of things: (1) things that have ports (and may not be component instances) and (2) things that have telemetry channels (and must be component instances).

Can we switch from focusing on the instances to focusing on the selection? Maybe we can call the section "Port Instance and Telemetry Channel Selectors" and have (1) one section on Port Instance Selectors and (2) another section on Telemetry Channel Selectors. We could also make (1) and (2) into completely separate sections, not under the same subsection.

In part (1), we can say that the left part of the selection is a port interface instance. In part (2), we can say that the left part of the selection is a component instance.

Kronos3: We decided to leave this section as is but update the description sentence to be tighter knit to what each instance member identifier _is_. (See 3cc01e39e61a4d3666af2ad9d8eec8c58cb9eec3)

docs/spec/Specifiers/Connection-Graph-Specifiers.adoc
```diff
@@ -92,12 +92,8 @@ unmatched.
...
 .. For each of the two port instance identifiers _I_ appearing in _C_:
 
-... The component instance named in _I_
-must be available in the enclosing topology,
-either through
-<<Specifiers_Component-Instance-Specifiers,direct specification>>
-or through
-<<Specifiers_Topology-Import-Specifiers,import>>.
+... _I_ must name an <<Instance-Member-Identifiers_Port-Instance-Identifiers,
+port instance id>> available in the enclosing topology.
```
bocchino: The concept of availability is now defined in the section on Topology Definitions, right? If so, we should provide a link there. Otherwise it's not clear what "available" means here.

docs/spec/Specifiers/Connection-Graph-Specifiers.adoc
```diff
@@ -92,12 +92,8 @@ unmatched.
 
 .. For each of the two port instance identifiers _I_ appearing in _C_:
 
-... The component instance named in _I_
-must be available in the enclosing topology,
-either through
-<<Specifiers_Component-Instance-Specifiers,direct specification>>
-or through
-<<Specifiers_Topology-Import-Specifiers,import>>.
+... _I_ must name an <<Instance-Member-Identifiers_Port-Instance-Identifiers,
```
bocchino:
```suggestion
... _I_ must name a <<Instance-Member-Identifiers_Port-Instance-Identifiers,
```

docs/spec/Specifiers/Topology-Port-Specifiers.adoc
```diff
@@ -0,0 +1,73 @@
...
+==== Example
+
+[source,fpp]
+----
+topology T {
+    instance c1
+    instance c2
+
+    connections {
+        # [...] Internal connections
```
bocchino: This notation is not clear. Does [...] mean that some code is being elided? If so I would move the ... out of the comment. What does "internal connections" mean? Are they connections between c1 and c2 and so "internal" to T? If so could you say that, or add some connections between c1 and c2?

docs/spec/Port-Interfaces.adoc
```diff
@@ -0,0 +1,89 @@
...
+
+=== Named Port Interfaces
+
+A named port interface is a port interface associated
+with an <<Definitions_Port-Interface-Definitions,
+interface definition>>.
+
+=== Anonymous Port Interfaces
+
+There are definitions in FPP that will implicitly define a port interface
```
bocchino: "There are definitions" is vague. Can we name the definitions and provide links to the corresponding sections of the spec?

docs/spec/Port-Interfaces.adoc
```diff
@@ -0,0 +1,89 @@
...
+so that instances of these definitions may be bound to a port interface.
+These implicitly defined port interfaces are known as anonymous port interfaces
+as they do not have explicit names associated with them.
+
+==== Anonymous Component Port Interfaces
+
+A <<Definitions_Component-Definitions,component definition>>
+will create a corresponding anonymous port interface. This port interface
```
bocchino:
```suggestion
defines a corresponding anonymous port interface. This port interface
```

docs/spec/Port-Interfaces.adoc
```diff
@@ -0,0 +1,89 @@
...
+includes the set of all <<Specifiers_Port-Instance-Specifiers,port instances>>
+inside the component definition (either explicitely, or via an
+<<Specifiers_Interface-Import-Specifiers,interface import>>).
+
+==== Anonymous Topology Port Interfaces
+
+A <<Definitions_Topology-Definitions,topology definition>>
+will create a corresponding anonymous port interface. This port interface includes
```
bocchino:
```suggestion
defines a corresponding anonymous port interface. This port interface includes
```

docs/spec/Port-Interfaces.adoc
```diff
@@ -0,0 +1,89 @@
...
+the set of all <<Specifiers_Topology-Port-Specifiers,topology port specifiers>>
+inside the topology definition.
+
+=== Port Interface Instance
+
+The port interface concept is simply a set of port instances.
```
bocchino:
```suggestion
A port interface is a named set of port instances.
```

docs/spec/Port-Interfaces.adoc
```diff
@@ -0,0 +1,89 @@
...
+
+. <<Definitions_Topology-Definitions,Topologies>>
+
+.. Topology values are bound to their corresponding
+<<Port-Interfaces_Anonymous-Port-Interfaces_Anonymous-Topology-Port-Interfaces,
+anonymous topology port interface>>.
+
+=== Binding
+
+Port interface binding refers to converting a port interface instance to an instance
```
bocchino:
```suggestion
*Port interface binding* means converting a port interface instance to an instance
```

docs/spec/Specifiers/Topology-Port-Specifiers.adoc
```diff
@@ -0,0 +1,73 @@
+=== Topology Port Specifiers
+
+A topology port specifiers names a <<Specifiers_Port-Instance-Specifiers,
```
bocchino:
```suggestion
A *topology port instance specifier* names a <<Specifiers_Port-Instance-Specifiers,
```

docs/spec/Definitions/Topology-Definitions.adoc
```diff
@@ -31,10 +37,24 @@ instance specifier>>
...
 * An <<Specifiers_Include-Specifiers,include specifier>>
 
 ==== Semantics
 
+The identifier is the name of the topology.
+
+If _interface-list_ is provided, each element of the list must
+<<Scoping-of-Names_Resolution-of-Qualified-Identifiers,refer to>> an
+<<Definitions_Port-Interface-Definitions,interface definition _K_>>.
+And the set of
```
bocchino:
```suggestion
<<Definitions_Port-Interface-Definitions,interface definition _K_>>,
and the set of
```

docs/spec/Definitions/Topology-Definitions.adoc
```diff
@@ -31,10 +37,24 @@ instance specifier>>
...
+<<Specifiers_Topology-Port-Specifiers,ports>> in the topology
+definition must be a super-set of the ports in _K_.
```
bocchino:
```suggestion
definition must be a superset of the set of port instance specifiers represented by _K_.
```

docs/spec/Specifiers/Topology-Port-Specifiers.adoc
```diff
@@ -0,0 +1,73 @@
+=== Topology Port Specifiers
```
bocchino:
```suggestion
=== Topology Port Instance Specifiers
```

docs/spec/Specifiers/Topology-Port-Specifiers.adoc
```diff
@@ -0,0 +1,73 @@
...
+==== Syntax
+
+`port`
+<<Lexical-Elements_Identifiers,_identifier_>>
+`=`
+<<Instance-Member-Identifiers_Port-Instance-Identifiers,_port-instance-id_>>
+
+==== Semantics
+
+For each topology port specifier defined in an enclosing topology _T_:
```
bocchino:
```suggestion
For each topology port instance specifier defined in an enclosing topology _T_:
```

docs/spec/Definitions/Topology-Definitions.adoc
```diff
@@ -31,10 +37,24 @@ instance specifier>>
...
 
 ==== Semantics
 
+The identifier is the name of the topology.
+
+If _interface-list_ is provided, each element of the list must
+<<Scoping-of-Names_Resolution-of-Qualified-Identifiers,refer to>> an
+<<Definitions_Port-Interface-Definitions,interface definition _K_>>.
+And the set of
+<<Specifiers_Topology-Port-Specifiers,ports>> in the topology
```
bocchino:
```suggestion
<<Specifiers_Topology-Port-Specifiers,port instance specifiers>> represented by the topology
```

We need to make the new definitions for topologies correspond to the existing definitions for port interface definitions.

1. We need to say "port instance specifier" everywhere, not "port."
2. We need to specify explicitly that a topology represents a set of port instance specifiers, as we do for port interface definitions. I think that the topology port specifiers map unqualified names to port instance specifiers of components, and the component port specifiers form the set that is referred to here. We need to say this.
3. We need to specify that the topology port specifiers for a topology must have distinct names, as we do for port interface definitions.

I think items 2 and 3 require some additional specification in the section on Topology Definitions -- unless this specification exists somewhere in the draft and I missed it.

bocchino: I see that item 3 is specified in the section on Topology Port Specifiers. However I think it might be good to move it to the section on Topology Definitions. We take a similar approach with Port Instance Specifiers and Component Definitions. The Specifier section specifies the local behavior of the specifier, and the enclosing definition section specifies the behavior of the collection of specifiers in the definition.

docs/spec/Port-Interfaces.adoc
```diff
@@ -0,0 +1,89 @@
...
+... _P_ and _P'_ have the same port type
+
+... If _P_ is an `input` port, _P'_ must also be an `input` port.
+
+.... _P'_ must have the same calling behavior as _P_ (i.e. `async`,
+`guarded`, `sync`).
+
+... If _P_ is an `output` port, _P'_ must also be an `output` port.
+
+... _P_ and _P'_ must share the same _number_ of ports.
```
bocchino:
```suggestion
... _P_ and _P'_ must have the same _number_ of port instances in their port instance arrays.
```

docs/spec/Port-Interfaces.adoc
```diff
@@ -0,0 +1,89 @@
...
+
+=== Binding
+
+Port interface binding refers to converting a port interface instance to an instance
+from another port interface.
+
+For an interface instance _A_ associated with an anonymous interface _A'_,
+the following conditions must be met for _A_ to bind to an interface _I_:
+
+. For every <<Specifiers_Port-Instance-Specifiers,port instance>> _P_ in _I_:
```
bocchino:
```suggestion
. For every <<Specifiers_Port-Instance-Specifiers,port instance specifier>> _P_ in _I_:
```

docs/spec/Port-Interfaces.adoc
```diff
@@ -0,0 +1,89 @@
...
+
+.. There exists a port instance _P'_ in _A'_ that has the same name of _P_.
```
bocchino:
```suggestion
.. There exists a port instance specifier _P'_ in _A'_ that has the same name as the name of _P_.
```

docs/spec/Port-Interfaces.adoc
```diff
@@ -0,0 +1,89 @@
...
+
+.. If _P_ is a <<Specifiers_Port-Instance-Specifiers_Semantics_General-Port-Instances,
+general port instance>>, _P'_ must also be a general port instance:
```
bocchino:
```suggestion
.. If _P_ specifies a <<Specifiers_Port-Instance-Specifiers_Semantics_General-Port-Instances,
general port instance>>, then _P'_ must also specify a general port instance:
```

docs/spec/Definitions/Topology-Definitions.adoc
bocchino: Here we leave the notion of equality of port instance specifiers implicit in the notion of set containment. However, in the section on interface binding, we have a very careful notion of equality of port instance specifiers as members of interfaces. Can we reuse that definition here? If not, how are they different?

docs/spec/Definitions/Topology-Definitions.adoc
bocchino: > I think that the topology port specifiers map unqualified names to port instance specifiers of components, and the component port specifiers form the set that is referred to here. 

Or is the set defined by substituting the names in the domain of the map into the port instance specifiers in the range of the map? In other words, if a topology T maps p to c.q, where q is `async input port q: P`, the set is `{ async input port p: P }`, right? This needs to be defined.

docs/spec/Definitions/Topology-Definitions.adoc
bocchino: I think we need to specify how to interpret a topology as an interface, and then say that the topology must be able to bind to each of the interfaces listed in the interface list.

Issue #741, opened 2025-06-13
Omission in base id and identifier range checking


bocchino:
This model should produce an error, but it passes fpp-check:
```
module Fw {
  port Time
  port Tlm
}

passive component C1 {

}

passive component C2 {
  time get port timeGetOut
  telemetry port tlmOut
  telemetry T: U32
}

instance c1: C1 base id 0x100
instance c2: C2 base id 0x100
```

Swapping the order of the instances c1 and c2 produces the expected error.

PR #749, opened 2025-06-24
Fix base id and identifier range checking

jwest115:
Fixed check for base id conflicts, updated tests
Closes https://github.com/nasa/fpp/issues/741.


bocchino (approved):
Looks good! I revised the error message and added a couple of tests.

PR #751, opened 2025-06-26
Topology ports syntax

Kronos3:


compiler/tools/fpp-format/test/include.ref.txt
```diff
@@ -237,8 +237,23 @@ module DefinitionsAndSpecifiers {
...
   } @< Topology definition
 
+  @ Topology definition with one implements
+  topology T implements I {
+
+
+  } @< Topology definition with one implements
+
+  @ Topology definition with two implements
+  topology T implements I, I {
```
bocchino: Can we render this as
```
topology T implements \
  I
  I
{

}
```
?

It's more consistent with the way we render other lists.

compiler/lib/src/main/scala/ast/Ast.scala
```diff
@@ -784,6 +786,12 @@ object Ast {
     omitted: List[AstNode[TlmChannelIdentifier]]
   )
 
+  /** Topology port specifier */
+  final case class SpecTopPort(
+    name: Ident,
+    underylingPort: AstNode[QualIdent],
```
bocchino:
```suggestion
    underlyingPort: AstNode[QualIdent],
```

PR #753, opened 2025-06-26
Add state machine Tester friend

m-aleem:
In support of nasa-fprime#3446 Similar to components, add a default `<X>Tester` to state machine classes, in order to support white-box testing


bocchino (changes requested):
This is a good start, but I think a bit more design is required. It looks like this approach uses the unqualified name of the state machine implementation class to create a friend for the state machine base class, and there's no friend for the derived implementation class. Instead, I suggest that we do the following:
1. Use two friends, one for the base class and one for the derived implementation class. The base class friend should be inserted here, and the derived class friend should be inserted in the code that instantiates a state machine in a component. See here for an example: https://github.com/nasa/fpp/blob/ccdfc4eb2e1838c3f3864d57078e844a4b663786/compiler/tools/fpp-to-cpp/test/component/base/SmInitialActiveComponentAc.ref.hpp#L63-L64.
2. Use the fully-qualified names for the friends, as the code generator does for the class names in the example above. In the example above, the names of the friends could be `FppTest::SmInitial::BasicStateMachineBaseTester` for the base class and `FppTest_SmInitial_BasicTester` for the derived class. It's a little more verbose, but it is more structured and less likely to create name collisions. Also, it's consistent with the existing code gen strategy.

bocchino:
> So in the example above we could use (1) FppTest::SmInitial::BasicStateMachineBaseTester for the friend of the state machine base class 

Or maybe it's better to use `FppTest::SmInitial::BasicTester` as the friend class name (similarly to what is currently done, but with qualification), so that the state machine tester can access the base class. I suspect that may play better with what is done in FppTest, and that's why you used the derived class name as the base class friend.

bocchino:
So after all that, my suggestion is to do the following:
1. Keep the derived-class name as the friend in the state machine base class, but add qualification.
2. Add the component tester and tester base as friends of the state machine implementation in the component generated code.

m-aleem:
I think I have addressed the requested updates, so using the example above now in file `SmInitialActiveComponentAc.ref.hpp` [classes derived from FppTest::SmInitial::BasicStateMachineBase](https://github.com/nasa/fpp/pull/753/files#diff-b86498e7eb508fdeb656d00df2b13febf94286168711b0118dd2f6623779402c) will have as friends SmInitialActiveTesterBase & SmInitialActiveTester (the same/existing friends as of `SmInitialActiveComponentBase`)

Let me know if this addresses the requested changes, happy to iterate further if needed.

PR #756, opened 2025-07-03
CppWriter corresponding to making TimeBase an FPP enum

m-aleem:
Needed for #3783 / [#3844](https://github.com/nasa/fprime/pull/3844). We moved the TimeBase enum to `FpConfig.fpp`, this is the correspondent update to the autocoder to pre-prend the use of TimeBase::TB_NONE, since it’s now a FPP-defined enum.


m-aleem:
@bocchino Using this FPP branch locally w/ devel-branch F-Prime builds & UTs pass, so I think this can be merged without having to do special coordination with F-Prime.

bocchino (approved):
Looks good! I confirmed that everything passes in check-cpp.

PR #764, opened 2025-07-29
Clang 20.1 UT Errors

m-aleem:


m-aleem:
@bocchino @thomas-bc Following up on the tagup discussion relating to this PR:

I tried to change the autocoder to do `const char* severityString = "SEVERITY ERROR";` and then remove assignment in the `default` case. Interestingly this gives the same error (seems like a false positive?):

```
Value stored to 'severityString' during its initialization is never read
```

So I think we have a couple options to keep the static analyzer happy:
1) `const char* severityString;` as currently proposed and keep the switch case as-is
2) `const char* severityString = nullptr;` and keep the switch case as-is
3) `const char* severityString = "UNKNOWN";  //NOLINT(clang-analyzer-deadcode.DeadStores)` and keep the switch case as-is
4) `const char* severityString = "SEVERITY ERROR";  //NOLINT(clang-analyzer-deadcode.DeadStores)` and update the `default` switch case to be empty/just `break`

Preferences?

bocchino (commented):
Looks good, but could we still initialize the `severityString` variable while avoiding the compiler warning? What happens if we initialize the variable to `""`? Does the compiler still warn?

m-aleem:
So to initialize it without warning I think we'd need to use the //NOLINT (and might as well keep the UNKNOWN in that case, I think)

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/TestWriter/ComponentHistory.scala
```diff
@@ -476,7 +476,7 @@ case class ComponentHistory(
       ),
       CppDoc.Type("void"),
       lines(
-        """|const char* severityString = "UNKNOWN";
+        """|const char* severityString;
```
bocchino:
```suggestion
        """|const char* severityString = {};
```
I think this should fix it. I verified it by installing clang-tidy 20.1.8, manually modifying the generated code, and running the check.

bocchino: `= nullptr` should also work (option 2 listed above). I think `= {}` does the same thing and is a bit shorter.

m-aleem: Yes, that's the same as option 2 (all options 1-4 provided were good/no warnings), so I'll update the PR.

bocchino: Sounds good.

bocchino (approved):
Looks good!

PR #765, opened 2025-07-30
Use dynamic size for array and struct records

### LINKED ISSUE #759: [Data products] Use dynamic string size to compute record size
When computing the size of a data product record, currently the size of a string record is the static serialized size (the max length of the string plus the size of the stored size) when it should be the dynamic serialized size (the actual length of the string plus the size of the stored size). This issue also occurs when strings are members of arrays or structs, or when arrays with string or struct members are members of arrays of structs, etc.

- In the CPP autocoder, rename the writeSerializedSizeExpr function to writeStaticSerializedSizeExpr
- In the CPP autocoder, add a new function called writeDynamicSerializedSizeExpr
- In the generated code for structs and arrays, add a function `getSerializedSize` that gets the dynamic serialized size of the struct or array, similarly to what we already have in `Fw::StringBase`.
- Use the dynamic size when generating the code that computes the contribution of a string, array, or struct to the size of a data product record.
- Test with check-cpp and fpp test

jwest115:
Use dynamic string size to compute record size for array and struct records.

Closes #759.

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ArrayCppWriter.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ArrayCppWriter.scala
@@ -437,7 +437,7 @@ case class ArrayCppWriter (
     ))
     val serializedSize = eltType.getUnderlyingType match {
       case ts: (Type.String | Type.Array | Type.Struct) => {
-        List(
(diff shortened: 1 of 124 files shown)


bocchino (approved):
Looks good!

PR #766, opened 2025-07-31
Deprecate serialize/deserialize in favor of [de]serialize[From/To]

Kronos3:
- [x] cpp-check
- [x] FppTest
- [x] Updated docs that reference `deserialize`/`serialize`


Kronos3:
I retested check-cpp and FppTest with the latest main backmerge since it changed our .ref.cpp files. Everything is still passing.

PR #771, opened 2025-08-04
Remove `fpp-to-xml` from docs

### LINKED ISSUE #732: Remove XML from tools and docs
Now that XML is gone from nasa/fprime/devel, we can do this.

- [ ] Remove `fpp-to-xml`
- [ ] Remove `fpp-from-xml`
- [ ] Revise User's Guide

bocchino:
Closes #732.


jwest115 (approved):
Looks good!

PR #772, opened 2025-08-04
Use dynamic serialized size for array records of structs and arrays

jwest115:


bocchino:
@jwest115 It looks like there are conflicts stemming from the recent changes to the serialization interface. Can you fix the conflicts and run CI?

bocchino (approved):
Looks good!

PR #782, opened 2025-08-19
Revise error checking for dot expressions

### LINKED ISSUE #781: Invalid dot expression causes a crash
Running this code through `fpp-check` produces a crash:
```
constant a = { x = 1 }.x
```
Instead it should produce an error.

bocchino:
Properly handle the case where the prefix expression is not a use of a definition. See &sect; 10.4 of the FPP Language Specification, v3.0.0.

Closes #781.


Kronos3 (approved):
Looks good!

bocchino:
Thanks! I think we'll have to revert this change when we add struct member access, but it's good to get this fix in and document it.

zimri-leisher:
Interesting, I just realized I had a similar issue in Fpy.

PR #783, opened 2025-08-19
Struct Member Selection Semantics

### LINKED ISSUE #769: Struct member semantics
- [x] Update analysis to support struct field selection in value expressions
- [x] Add some tests to validate the analysis

Kronos3:
This PR reworks the use analysis of dot expressions to allow selecting members of expressions that have struct/anon struct types. This change is a little tricky since the previous implementation assumed that every complete dot expression was a constantUse of some sort. This is no longer the case. Now _part_ of a dot expression can be a constantUse and another might be a member selection. Also we are allowing member selection on struct literals so some dot expressions might not have _any_ constantUses.

Closes #769

### DIFF
--- a/docs/fpp-spec.html
+++ b/docs/fpp-spec.html
@@ -5,7 +5,7 @@
 <meta http-equiv="X-UA-Compatible" content="IE=edge">
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
 <meta name="generator" content="Asciidoctor 2.0.23">
-<title>The F Prime Prime (FPP) Language Specification, v3.0.0</title>
+<title>The F Prime Prime (FPP) Language Specification, Unreleased, after v3.0.0</title>
 <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Open+Sans:300,300italic,400,400italic,600,600italic%7CNoto+Serif:400,400italic,700,700italic%7CDroid+Sans+Mono:400,700">
 <style>
 /*! Asciidoctor default stylesheet | MIT License | https://asciidoctor.org */
@@ -437,7 +437,7 @@ body.book #toc,body.book #preamble,body.book h1.sect0,body.book .sect1>h2{page-b
 </head>
 <body class="article toc2 toc-left">
 <div id="header">
-<h1>The F Prime Prime (FPP) Language Specification, v3.0.0</h1>
+<h1>The F Prime Prime (FPP) Language Specification, Unreleased, after v3.0.0</h1>
 <div id="toc" class="toc2">
 <div id="toctitle">Table of Contents</div>
 <ul class="sectlevel1">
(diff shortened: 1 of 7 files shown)


Kronos3:
@bocchino I made the change we discussed, I created another analyzer called `BasicUseAnalyzer` which effectively works the same way as before this change, it assumes a dotExpr is a `constantUse` iff its a qual-ident.

Then the true `UseAnalyzer` is simply an extension to the `BasicUseAnalyzer` which limits `constantUse` to the part of the qualified identifier that actually is the `constantUse`. In `fpp-depend`, only the basic analyzer is used since we don't run `CheckUses` in that tool.

compiler/lib/src/main/scala/analysis/CheckSemantics/CheckUses.scala
```diff
@@ -133,7 +148,7 @@ object CheckUses extends UseAnalyzer {
         for {
           _ <- {
             Result.annotateResult(
-              constantUse(a, itu.asExprNode, itu.name),
```
bocchino, line 136: This looks like it introduces a regression. In the old code, `constantUse` would fail if the definition referred to did not exist. Now it looks like that check is gone, because the `constantUse` analysis here is more permissive. I wonder if the `BasicUseAnalyzer` should have two kinds of `constantUse` functions: one that works the way it worked before, where we are expecting a constant use and we fail if it's not there; and one that doesn't a priori know whether a dot expression is a constant use, so allows it not to be. The second one could try the first one and check for value selection if it fails. Here it seems we should be using the first one.

bocchino: After discussion, we'll (1) call `constantUse` here and (2) check that the qualified identifier really is a constant use, i.e., it is represented in the use-def map.

Kronos3: Turns out this was a little more involved than we initially thought. Implied uses, unlike normal AstNodes will recursively use the same node id on all sub expressions. This means that if one of the sub-expressions qualifying an implied use _does_ reference a constant def, the entire expression will act like a constant def.

I added a special method `ImpliedUse.asUniqueExprNode` which generates an ast node with unique node ids to avoid this issue. I added a test to trigger the warning about an implied use not being a constant definition (and just being a member of a constant).

compiler/lib/src/main/scala/analysis/CheckSemantics/CheckUses.scala
```diff
@@ -121,24 +136,54 @@ object CheckUses extends UseAnalyzer {
...
-            s"when constructing a dictionary, the type ${itu.name} must be defined"
-          )
-        } yield ()
-      })
       _ <- Result.foldLeft (impliedConstantUses) (()) ((_, itu) => {
+        val impliedUse = itu.asUniqueExprNode
+
         for {
-          _ <- {
+          a <- {
```
bocchino, line 143: I think this `a` should be an underscore. We are not using the result.

compiler/lib/src/main/scala/analysis/CheckSemantics/CheckUses.scala
bocchino: Sounds good.

compiler/lib/src/main/scala/analysis/CheckSemantics/CheckUses.scala
```diff
@@ -121,24 +136,54 @@ object CheckUses extends UseAnalyzer {
...
             Result.annotateResult(
-              constantUse(a, itu.asExprNode, itu.name),
+              constantUse(a, impliedUse, itu.name),
               s"when constructing a dictionary, the constant ${itu.name} must be defined"
             )
           }
+
+          // Check to make sure this implied use is actually a constant
+          // ...rather than a member of a constant.
+          a <- {
```
bocchino: I think this `a` should be an underscore.

bocchino (approved):
Looks good, just one comment.

bocchino:
Before we merge this PR, we need to extract the specification for this behavior from #748, and merge it. I just realized that the specification for this behavior is not in main, and it's not in this PR. So things are a little out of order. Let's fix it.

compiler/lib/src/main/scala/analysis/CheckSemantics/CheckUses.scala
Kronos3: We use this in the next step, the second one is not used though.

PR #792, opened 2025-08-21
Array Subscript Expr Syntax 

### LINKED ISSUE #789: Array Member Expression Syntax
- [x] Implement array subscript expression syntax
- [x] Add tests

Kronos3:
Closes #789

### DIFF
--- a/compiler/lib/src/main/scala/ast/Ast.scala
+++ b/compiler/lib/src/main/scala/ast/Ast.scala
@@ -285,6 +285,7 @@ object Ast {
   /** Expression */
   sealed trait Expr
   final case class ExprArray(elts: List[AstNode[Expr]]) extends Expr
+  final case class ExprArraySubscript(e: AstNode[Expr], i: AstNode[Expr]) extends Expr
   final case class ExprBinop(e1: AstNode[Expr], op: Binop, e2: AstNode[Expr]) extends Expr
   final case class ExprDot(e: AstNode[Expr], id: AstNode[Ident]) extends Expr
   final case class ExprIdent(value: Ident) extends Expr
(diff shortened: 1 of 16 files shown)


compiler/lib/src/main/scala/ast/Ast.scala
```diff
@@ -285,6 +285,7 @@ object Ast {
   /** Expression */
   sealed trait Expr
   final case class ExprArray(elts: List[AstNode[Expr]]) extends Expr
```
bocchino, line 287: Can we rename the fields of `ExprArray` to `e1` and `e2`? It's more consistent with the other expression forms.

Kronos3: I'm not sure I agree, the only other Expr that uses `e1`/`e2` is BinOp where I think it makes sense. I think its more clear that `e` and `i` do. It falls inline with how `ExprDot` is named.

bocchino: `ExprDot` has an expression `e` and an identifier `id`. Here we have an expression `e` and an expression `i`. It's nonstandard to use the letter `i` to refer to an expression. I expect `i` to be an instance or identifier or something that starts with i.

`e1` and `e2` seem clear and seem consistent with all the other expression forms, including BinOp. We name things after what they are (expression, identifier, value) and introduce numbers when we have more than one of them. This is a standard approach in language definition. Is the concern that `e2` doesn't specifically call out that `e2` is an index? I think that's clear enough from the form of the syntax.

Kronos3: Ok that's reasonable, I'll make the change

PR #791, opened 2025-08-21
Allow array types and values of arbitrary size

bocchino:
Closes #694.

Procedure:

1. Review this PR.
2. Create an alpha release off this branch.
3. Create a PR to F Prime that points to the alpha release.
4. Review and merge the F Prime PR.
5. Merge this PR.

PR #795, opened 2025-08-22
Array subscript semantics

### LINKED ISSUE #787: Array Member Expression Semantics
- [x] Implement array member semantics
- [x] Add tests

Kronos3:
Closes #787


compiler/lib/src/main/scala/analysis/CheckSemantics/CheckExprTypes.scala
```diff
@@ -112,6 +112,35 @@ object CheckExprTypes extends UseAnalyzer {
...
+          case _ => Left(SemanticError.InvalidType(
+            Locations.get(e.e.id),
+            "expected an array type"
+          ))
+        }
+      }
+
+      a <- super.exprNode(a, e.i)
+      _ <- {
+        if a.typeMap(e.i.id).getUnderlyingType.isInt
```
bocchino: For consistency, I think we should use `convertNodeToNumeric` here. Look at how it's done for the size expression of an array type definition.

compiler/lib/src/main/scala/analysis/CheckSemantics/EvalConstantExprs.scala
```diff
@@ -88,6 +88,46 @@ object EvalConstantExprs extends UseAnalyzer {
...
+        a.valueMap(e.i.id) match {
+          case Value.PrimitiveInt(value, _) => Right(value)
+          case Value.Integer(value) => Right(value)
+          case _ => throw InternalError("type of index should be an integer type")
+        }
+      }
+
+      // Check if the index is in bounds
+      _ <- {
+        if index < 0
```
bocchino, line 114: Let's use `Analysis.getBoundedArraySize` here.

Kronos3: I don't think we can because of two reasons:
1. getBoundedArraySize was removed as part of #791 (we can still use `getUnboundedArraySize`)
2. Size bounds != index bounds. Size bounds check for `size >= 1` which won't work with index bounds since `0` is a valid index.

bocchino: That makes sense.

compiler/lib/src/main/scala/analysis/CheckSemantics/CheckExprTypes.scala
```diff
@@ -112,6 +112,26 @@ object CheckExprTypes extends UseAnalyzer {
...
+    for {
+      a <- super.exprNode(a, e.e1)
+
+      eltType <- {
+        a.typeMap(e.e1.id).getUnderlyingType match {
+          case Type.AnonArray(_, eltType) => Right(eltType)
+          case Type.Array(_, anonArr, _, _) => Right(anonArr.eltType)
+          case _ => Left(SemanticError.InvalidType(
+            Locations.get(e.e1.id),
+            "expected an array type"
```
bocchino: Can we report the non-array type that we saw here?

compiler/lib/src/main/scala/analysis/CheckSemantics/CheckExprTypes.scala
```diff
@@ -112,6 +112,26 @@ object CheckExprTypes extends UseAnalyzer {
...
+    for {
+      a <- super.exprNode(a, e.e1)
+
+      eltType <- {
+        a.typeMap(e.e1.id).getUnderlyingType match {
+          case Type.AnonArray(_, eltType) => Right(eltType)
+          case Type.Array(_, anonArr, _, _) => Right(anonArr.eltType)
+          case t => Left(SemanticError.InvalidType(
+            Locations.get(e.e1.id),
+            s"expected an array type, got '${t.toString()}'"
```
bocchino:
```suggestion
            s"${t.toString()} is not an array type"
```

compiler/lib/src/main/scala/analysis/CheckSemantics/EvalConstantExprs.scala
```diff
@@ -88,6 +88,46 @@ object EvalConstantExprs extends UseAnalyzer {
...
+        then Left(SemanticError.InvalidIntValue(
+          Locations.get(e.e2.id),
+          index,
+          "value may not be negative"
+        ))
+        else if index >= elements.length
+        then Left(SemanticError.InvalidIntValue(
+          Locations.get(e.e2.id),
+          index,
+          s"index value out of array bounds"
```
bocchino: This error message could be slightly more helpful, e.g., `index value must be in range [0, 9]`.

bocchino: I noticed this after approving the PR.

bocchino: Or better: `index value 10 is not in the range [0, 9]`.

bocchino: I guess the index value is already reported by `InvalidIntValue`. So only the range needs to be reported in the extra message.

PR #797, opened 2025-09-03
Tighten up constants-cpp include 

### LINKED ISSUE #796: Tighten up constants-cpp include
In https://github.com/nasa/fprime/issues/3288 we are moving C++/C constants to the FPP model. One of our includes in FPP causes an include loop so we need to tighten up the include to not cause an include loop.

Kronos3:
Closes #796


compiler/lib/src/main/scala/codegen/CppWriter/ConstantCppWriter.scala
```diff
@@ -15,7 +15,7 @@ object ConstantCppWriter extends CppWriterUtils {
       case constantMembers => 
         val fileName = ComputeCppFiles.FileNames.getConstants
         val hppHeaderLines = {
-          val headers = List("Fw/FPrimeBasicTypes.hpp")
+          val headers = List("Fw/Types/BasicTypes.h")
```
bocchino: As discussed, we should include the `.hpp` version of this file.

PR #799, opened 2025-09-04
Unify all FPP tools

### LINKED ISSUE #798: Unify FPP into a single tool
To speed up our CI build times/PIP package size we can unify all the FPP tools into a single binary and use git-style subcommands to select each tool. This will mean only a single native build.

Kronos3:
This PR creates a single FPP entrypoint that wraps our tools. It passes a single argument to a `scopt` parser to allow parsing the FPP command. The top level entrypoint select the tool to run and then passes off the rest of the arguments to that tools entrypoint.

The main tool also includes a nice `--help` message to guide users (I copied the `--help` style of Docker and git):
```
$ fpp --help
fpp v3.0.1a1-1-g72d368fdd
Usage: fpp [options] COMMAND

FPP Commands:

check       performs semantic checking of FPP models
depend      computes dependencies for FPP source files
filenames   writes out the names of C++ files generated from FPP source files
format      parses FPP source files and writes out formatted source files
from-xml    parses older F Prime XML files and converts them to FPP files
locate-defs parses FPP source files and reports the locations of symbol definitions
locate-uses parses FPP source files and reports the locations of symbols used in the files
syntax      parses FPP source files into an abstract syntax tree (AST) and optionally writes out the AST
to-cpp      parses FPP models, performs semantic checking on them, and writes out C++ files
to-dict     writes JSON dictionaries corresponding to F Prime topologies
to-json     parses an FPP model, performs semantic checking on it, and writes out the model in JSON
to-layout   writes layout text files for connection graphs within F Prime topologies

  -h, --help     print this message and exit
  -v, --version  print version and exit
  COMMAND

Run 'fpp COMMAND --help' for more information on a command.

```

Closes #798 

- [x] Get CI to pass here
- [x] Make sure we are publishing all the tools we need to work with F Prime, e.g., wrappers like `fpp-check` that call `fpp check`.
- [x] Make an alpha release against this branch.
- [x] Create a PR to F Prime with any changes to install this version of FPP. Get CI to pass there and merge the PR. https://github.com/nasa/fprime/pull/4287
- [x] Merge this PR.

### DIFF
--- a/compiler/lib/src/main/resources/META-INF/native-image/reflect-config.json
+++ b/compiler/lib/src/main/resources/META-INF/native-image/reflect-config.json
@@ -9,9 +9,6 @@
   "name":"[Lio.circe.Encoder;"
 },
 {
-  "name":"[Ljava.io.File;"
-},
-{
   "name":"[Ljava.lang.String;"
 },
 {


Kronos3:
Ah good catch, `to-xml` is actually included I just forgot to put it in the help message since I was getting the descriptions from https://github.com/nasa/fpp/wiki/Tools where its missing. I'll add it in.

bocchino:
I'm marking this PR as fprime-fpp-tandem, since this represents a significant change to how we publish the FPP tools to F Prime. Before merging this PR, let's make sure we have worked out the details of how F Prime will use this version of the tools.

Kronos3:
I did a combo of both, I opened a PR that removes XML entirely against this PR.

bocchino:
@Kronos3 It looks like the FPP unit tests are passing, but the native image build is failing. At this point I think we can hand it over to @LeStarch to fix the native image build. He said he would help us with this.

pyproject.toml
```diff
@@ -0,0 +1,62 @@
...
+fpp-filenames = "fprime_fpp.__main__:main"
+fpp-format = "fprime_fpp.__main__:main"
+fpp-from-xml = "fprime_fpp.__main__:main"
+fpp-locate-defs = "fprime_fpp.__main__:main"
+fpp-locate-uses = "fprime_fpp.__main__:main"
+fpp-syntax = "fprime_fpp.__main__:main"
+fpp-to-cpp = "fprime_fpp.__main__:main"
+fpp-to-dict = "fprime_fpp.__main__:main"
+fpp-to-json = "fprime_fpp.__main__:main"
+fpp-to-layout = "fprime_fpp.__main__:main"
```
Kronos3, line 56: @LeStarch Do you think this will add a significant overhead on the F Prime build since it'll spin up a Python interpreter to wrap the `fpp` executable?

LeStarch: No, it should not.  This is exactly how the old native images worked....just less hidden.  They too had python wrappers to call binary files.  This overhead hasn't been noticed so far.   Here is why it must be done:

1. Cannot modify `bin` in python packaging directly. PIP prevents users from putting things in `bin` easily.
2. Cross-Platform support and F Prime standards.  F Prime uses python as opposed to bash for our tooling (e.g. a shell wrapper).

Kronos3: Ah understood, I didn't realize you needed wrappers in the previous build. Thanks!

PR #800, opened 2025-09-04
Revise spec for value conversion

### LINKED ISSUE #794: Type conversion rules for string
I was trying to find out where in the FPP spec it details what exactly happens when a string of size X is converted to string size Y where X > Y. Obviously I assume it truncates, but I figured it would be good to check. I went to the [type conversion section](https://nasa.github.io/fpp/fpp-spec.html#Type-Checking_Type-Conversion) and saw that "Any [string type](https://nasa.github.io/fpp/fpp-spec.html#Types_String-Types) may be converted to any other string type."
This doesn't quite mention that truncation happens. Then I looked at the first sentence of section 18.13, which says: "We say that a type T1 may be converted to another type T2 if every [value](https://nasa.github.io/fpp/fpp-spec.html#Values) represented by type T1 can be [converted](https://nasa.github.io/fpp/fpp-spec.html#Evaluation_Type-Conversion) into a value of type T2."
So I looked at the definition for [value conversion](https://nasa.github.io/fpp/fpp-spec.html#Evaluation_Type-Conversion) but it didn't mention strings either.

Where are string type conversions defined?

bocchino:
* Clarify how string values are converted to different string types
* Fix an omission in the value conversion rules for arrays

Closes #794.

### DIFF
--- a/docs/spec/Types.adoc
+++ b/docs/spec/Types.adoc
@@ -42,6 +42,12 @@ where _n_ is an integer value in the range [1,2^31-1].
 There is one string type `string` and one string type `string size` _n_
 for each legal value of _n_.
 
+The value _n_, if present, is called the *size* associated with the string 
+type.
+The size provides a maximum string length that the code generator can use
+for allocating memory and serializing data.
+If no explicit size is present, then the code generator uses a default size.
+
 === Array Types
 
 An *array type* is a type associated with an
--- a/docs/fpp-spec.html
+++ b/docs/fpp-spec.html
@@ -9888,6 +9888,13 @@ where <em>n</em> is an integer value in the range [1,2^31-1].
 There is one string type <code>string</code> and one string type <code>string size</code> <em>n</em>
 for each legal value of <em>n</em>.</p>
 </div>
+<div class="paragraph">
+<p>The value <em>n</em>, if present, is called the <strong>size</strong> associated with the string
+type.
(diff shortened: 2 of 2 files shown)


docs/spec/Types.adoc
```diff
@@ -39,7 +39,7 @@ The *string types* correspond to the
 <<Type-Names_String-Type-Names,string type names>>.
 A string type is written `string` or `string size` _n_,
 where _n_ is an integer value in the range [1,2^31-1].
-There is one string type `string` and one string type `string` _n_
+There is one string type `string` and one string type `string size` _n_
```
Kronos3, line 42: Maybe not related to this change but the meaning of the string size seems underspecified. For example what happens when the size is not provided. Also, where is the size used by the model.

bocchino: All this is handled by the code generator. I can put in some more explanation to that effect.

bocchino: Note that in general, code generation behavior is not specified in the FPP spec. We just specify the semantics of the model. In the model, the semantics of string types is that they can have an optional size, and that's about it. The code generator uses the size, if present, to allocate buffers.

bocchino: OK, I added some more explanation.

Kronos3 (approved):
Looks good, thanks for updating!

PR #802, opened 2025-09-05
Fix bug in ComponentTestUtils

bocchino:
Generate test history code in the case that there is telemetry and there are no other output ports. Closes https://github.com/nasa/fprime/issues/4120.

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/TestWriter/ComponentTestUtils.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/TestWriter/ComponentTestUtils.scala
@@ -28,7 +28,6 @@ abstract class ComponentTestUtils(
 
   val hasHistories: Boolean =
     hasCommands ||
-    hasCommands ||
     hasDataProducts ||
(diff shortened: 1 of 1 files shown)


compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/TestWriter/ComponentTestUtils.scala
```diff
@@ -27,12 +27,13 @@ abstract class ComponentTestUtils(
   val hasTypedOutputPorts = typedOutputPorts.nonEmpty
 
   val hasHistories: Boolean =
-    hasTypedOutputPorts ||
     hasCommands ||
-    hasParameters ||
-    hasEvents ||
     hasCommands ||
-    hasDataProducts
```
Kronos3: `hasCommands` is listed twice.

bocchino: Good catch, fixed.

Kronos3 (approved):

PR #801, opened 2025-09-04
Purge FPP-to-XML

Kronos3:
This PR is just meant to split off from #799 so that its easier to understand whats getting deleted. The tests won't pass since they are not working on #799 yet but we can merge regardless.

Closes #770

Issue #770, opened 2025-08-04
Remove `fpp-to-xml` from FPP tools


bocchino:
- [x] Remove `fpp-to-xml` from CI and tool publication
- [x] Remove `fpp-to-xml` from the code base

LeStarch:
@bocchino not for v3 FPP/v4 FPrime, right?

bocchino:
Right.

bocchino:
This task is not complete -- the first check box is unchecked. But that check box will be handled by #798, so I think it's OK to close this issue.

PR #811, opened 2025-09-15
Remove FPP-To-XML

Kronos3:
This PR is similar to #801 but it creates a dummy binary for `to-xml` so that our current build is happy.

Related to #770 

@bocchino I ran this against FppCheck to make sure the build still works with this dummy tool and it's still working.

- [x] Create FPP alpha release
- [x] Create PR against nasa/fprime https://github.com/nasa/fprime/pull/4161
- [x] Merge PR against nasa/fprime
- [x] Merge this PR

### DIFF
--- a/compiler/tools/fpp-to-xml/src/main/scala/fpp-to-xml.scala
+++ b/compiler/tools/fpp-to-xml/src/main/scala/fpp-to-xml.scala
@@ -1,8 +1,36 @@
 package fpp.compiler.tools
 
+import fpp.compiler.analysis._
+import fpp.compiler.ast._
+import fpp.compiler.codegen._
+import fpp.compiler.syntax._
+import fpp.compiler.transform._
+import fpp.compiler.util._
+import scopt.OParser
+
 object FPPToXml {
 
+  case class Options()
+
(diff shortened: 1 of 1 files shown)


compiler/tools/fpp-to-xml/src/main/scala/fpp-to-xml.scala
```diff
@@ -1,108 +1,10 @@
...
-        .text("default string size"),
-      arg[String]("file ...")
-        .unbounded()
-        .optional()
-        .action((f, c) => c.copy(files = File.fromString(f) :: c.files))
-        .text("files to translate"),
-    )
+  def main(args: Array[String]) = {
+    println("fpp-to-xml is deprecated")
+    System.exit(0)
```
bocchino:
```suggestion
    println("fpp-to-xml is not supported")
```

bocchino:
Looks good! I'll approve at the end of the tandem process.

I added the help and version info back into the stub `fpp-to-xml`. That info is needed to support the tools I use to pull over non-released versions of FPP to F Prime.

bocchino (approved):
Looks good!

PR #820, opened 2025-09-18
Event throttle time semantics

### LINKED ISSUE #806: Time limited event throttling semantics
Check that the specified duration is valid.

Kronos3:
Closes #806


compiler/lib/src/main/scala/analysis/Semantics/Event.scala
```diff
@@ -35,14 +45,59 @@ object Event {
...
       }
+      def getEveryIntervalValue(every: AstNode[Ast.Expr]) = {
+        val loc = Locations.get(every.id)
+        val Value.AnonStruct(intervalValue) = Analysis.convertValueToType(a.valueMap(every.id), Type.AnonStruct(
+          Map(
+            ("seconds", Type.U32),
+            ("useconds", Type.U32),
+          )
+        ))
+        def getNonZeroMember(member: String, maxValue: Option[Int] = None) = {
```
bocchino: I think this function should not be called `getNonZeroMember`. Zero is allowed. Maybe just `getMember`?

compiler/lib/src/main/scala/analysis/Semantics/Event.scala
```diff
@@ -35,14 +45,59 @@ object Event {
...
+            intervalValue.get(member).get,
+            Type.U32
+          )
+
+          if (v < 0) {
+            Left(SemanticError.InvalidIntValue(
+              loc, v, s"$member may not be negative"
+            ))
+          } else {
+            maxValue match {
```
bocchino: I think we should always have a max value here. For the seconds value, it has to fit in a U32.

compiler/lib/src/main/scala/analysis/Semantics/Event.scala
```diff
@@ -35,14 +45,59 @@ object Event {
...
+              case Some(m) => if v >= m then
+                Left(SemanticError.InvalidIntValue(
+                  loc, v, s"$member must be smaller than $m"
+                )) else Right(v.intValue)
+              case None => Right(v.intValue)
+            }
+          }
+        }
+        for {
+          seconds <- getNonZeroMember("seconds")
```
bocchino: The seconds value must fit in a U32.

bocchino (changes requested):
Looks good. Please see the comments. Also, let's add tests for `fpp-depend` and `fpp-locate-uses`.

compiler/lib/src/main/scala/analysis/Semantics/Event.scala
bocchino: Actually the spec says it has to fit in an I32, and the implementation reflects that. But maybe we should make it a U32, since that is what F Prime uses? If so we need to update the spec and the implementation. In Scala we could use a Long for this, or a BigInt.

bocchino: I just updated the spec.

compiler/lib/src/main/scala/analysis/Semantics/Event.scala
```diff
@@ -35,14 +45,56 @@ object Event {
...
+            ))
+          } else {
+            if v >= maxValue then
+            Left(SemanticError.InvalidIntValue(
+              loc, v, s"$member must be smaller than $maxValue"
+            )) else Right(v.longValue)
+          }
+        }
+        for {
+          seconds <- getMember("seconds", 4294967295L)
```
bocchino: Can we add a unit test covering this upper bound?

bocchino: Also, can we move this integer literal to a constant defined in `util`? Maybe we could add a `Uint` object (we already have a `Chars` object) and define `Uint.max`.

bocchino: I meant `UInt` instead of `Uint` -- auto-correct "corrected" it.

bocchino (approved):
Looks good! I made a few small changes to the error reporting.

PR #821, opened 2025-09-18
Event Throttle Time Codegen

### LINKED ISSUE #807: Time limited event throttling codegen
- [x] Update CPP Writer to output two counters or time limited event throttling, one of count (already exists) and one for time.

Kronos3:
Closes #807

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentEvents.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentEvents.scala
@@ -349,11 +349,19 @@ case class ComponentEvents (
           eventThrottleResetName(event),
           Nil,
           CppDoc.Type("void"),
-          lines(
-            s"""|// Reset throttle counter
-                |this->${eventThrottleCounterName(event.getName)} = 0;
-                |"""
-          )
+          List(
+            lines(
+              s"""|// Reset throttle counter
+                  |this->${eventThrottleCounterName(event.getName)} = 0;
+                  |"""
+            ),
+            if event.throttle.get.every.isDefined
+            then lines(
+              s"""|// Reset throttle timeout
+                  |this->${eventThrottleTimeName(event.getName)}.set(0, 0);
+                  |"""
+            ) else Nil
+          ).flatten
         )
       )
     )
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentEvents.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentEvents.scala
@@ -60,12 +60,23 @@ case class ComponentEvents (
     addAccessTagAndComment(
       "private",
       "Counter values for event throttling",
-      throttledEvents.map((_, event) =>
-        linesClassMember(
-          Line.blank :: lines(
-            s"""|//! Throttle for ${event.getName}
-                |std::atomic<FwIndexType> ${eventThrottleCounterName(event.getName)};
-                |"""
+      List.concat(
+        throttledEvents.map((_, event) =>
+          linesClassMember(
+            Line.blank :: lines(
+              s"""|//! Throttle for ${event.getName}
+                  |std::atomic<FwIndexType> ${eventThrottleCounterName(event.getName)};
+                  |"""
+            )
+          )
+        ),
+        intervalThrottledEvents.map((_, event) =>
+          linesClassMember(
+            Line.blank :: lines(
+              s"""|//! Throttle time ${event.getName}
+                  |Fw::Time ${eventThrottleTimeName(event.getName)};
+                  |"""
+            )
           )
         )
       ),
@@ -246,9 +257,10 @@ case class ComponentEvents (
     )
     def writeBody(id: Event.Id, event: Event) = intersperseBlankLines(
       List(
+        // Hard throttle counter can be checked immediately
+        // We don't need to get time
         event.throttle match {
-          case None => Nil
-          case Some(_) => lines(
+          case Some(Event.Throttle(_, None)) => lines(
             s"""|// Check throttle value
                 |if (this->${eventThrottleCounterName(event.getName)} >= ${eventThrottleConstantName(event.getName)}) {
                 |  return;
@@ -258,6 +270,7 @@ case class ComponentEvents (
                 |}
                 |"""
           )
+          case _ => Nil
         },
         lines(
           s"""|// Get the time
@@ -271,6 +284,27 @@ case class ComponentEvents (
               |_id = this->getIdBase() + ${eventIdConstantName(event.getName)};
               |"""
         ),
+        // Time based throttle timeout needs above time
+        event.throttle match {
+          case Some(Event.Throttle(_, Some(Event.TimeInterval(seconds, useconds)))) => lines(
+            s"""|// Check throttle value & throttle timeout
+                |if (this->${eventThrottleCounterName(event.getName)} >= ${eventThrottleConstantName(event.getName)}) {
+                |  if (Fw::TimeInterval(this->${eventThrottleTimeName(event.getName)}, _logTime) >= Fw::TimeInterval($seconds, $useconds)) {
+                |    // Reset the throttle count & timeout
+                |    this->${eventThrottleTimeName(event.getName)} = _logTime;
+                |    this->${eventThrottleCounterName(event.getName)} = 0;
+                |  }
+                |  else {
+                |    return;
+                |  }
+                |}
+                |else {
+                |  (void) this->${eventThrottleCounterName(event.getName)}.fetch_add(1);
+                |}
+                |"""
+          )
+          case _ => Nil
+        },
         writeLogBody(id, event),
         writeTextLogBody(event)
       )
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriter.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriter.scala
@@ -562,6 +562,9 @@ case class ComponentCppWriter (
               throttledEvents.map((_, event) => line(
                 s"this->${eventThrottleCounterName(event.getName)} = 0;"
               )),
+              intervalThrottledEvents.map((_, event) => line(
+                s"this->${eventThrottleTimeName(event.getName)}.set(0, 0);"
+              )),
               sortedParams.flatMap((_, param) => guardedList(!param.isExternal) (
                 lines(s"this->${paramValidityFlagName(param.getName)} = Fw::ParamValid::UNINIT;")
               ))
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriterUtils.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriterUtils.scala
@@ -202,6 +202,14 @@ abstract class ComponentCppWriterUtils(
     }
   )
 
+  /** List of throttled events */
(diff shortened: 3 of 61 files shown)


compiler/lib/src/main/scala/analysis/Semantics/Event.scala
```diff
@@ -70,17 +70,17 @@ object Event {
...
-          }
         } yield TimeInterval(seconds, useconds.toInt)
       }
       def checkEventThrottle(throttle: AstNode[Ast.EventThrottle]) = {
         for {
           count <- a.getNonnegativeIntValue(throttle.data.count.id)
+          _ <- {
+            if count > 0 then Right(())
+            else Left(SemanticError.InvalidIntValue(
+              loc, 0, s"event throttle count must be greater than zero"
```
bocchino:
```suggestion
              loc, count, s"event throttle count must be greater than zero"
```

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentEvents.scala
```diff
@@ -271,6 +284,26 @@ case class ComponentEvents (
               |_id = this->getIdBase() + ${eventIdConstantName(event.getName)};
               |"""
         ),
+        // Time based throttle timeout needs above time
+        event.throttle match {
+          case Some(Event.Throttle(_, Some(Event.TimeInterval(seconds, useconds)))) => lines(
+            s"""|// Check throttle value & throttle timeout
+                |if (this->${eventThrottleCounterName(event.getName)} == 0 ||
```
bocchino: This code is not concurrency safe. The original code was doing a fetch and increment. We need to figure out how to add this logic to the fetch and increment.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentEvents.scala
```diff
@@ -60,12 +60,23 @@ case class ComponentEvents (
...
+                  |std::atomic<FwIndexType> ${eventThrottleCounterName(event.getName)};
+                  |"""
+            )
+          )
+        ),
+        intervalThrottledEvents.map((_, event) =>
+          linesClassMember(
+            Line.blank :: lines(
+              s"""|//! Throttle time for ${event.getName}
+                  |Fw::Time ${eventThrottleTimeName(event.getName)};
```
bocchino, line 80: Fw::Time should be atomic if this variable is accessed concurrently.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentEvents.scala
bocchino: I think we can do an atomic `fetch_add` and check the return value, which is the value just before the increment. I think the existing logic for checking the throttle count also needs to be corrected. Instead of checking and then incrementing, which is not guaranteed to be atomic, we should atomically increment and then check the return value.

bocchino: > I think the existing logic for checking the throttle count also needs to be corrected.

The existing logic is
```
1  if (throttleCount < maxThrottleCount) {
2    atomically increment throttle count
3    emit event
4  }
```
That's not correct because any number of threads could read the same throttle count value in line 1. Instead I think it should be
```
1  if (throttleCount < maxThrottleCount) {
2    atomically increment throttleCount and set prevThrottleCount to the value before the increment
3    if (prevThrottleCount < maxThrottleCount) {
4      emit event
5    }
6  }
```
That way any number of threads can read the throttle count in line 1, but only one of them will win the race in line 2 and proceed to line 4.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentEvents.scala
```diff
@@ -271,6 +284,34 @@ case class ComponentEvents (
               |_id = this->getIdBase() + ${eventIdConstantName(event.getName)};
               |"""
         ),
+        // Time based throttle timeout needs above time
+        event.throttle match {
+          case Some(Event.Throttle(_, Some(Event.TimeInterval(seconds, useconds)))) => lines(
+            s"""|// Check throttle value & throttle timeout
+                |FwIndexType count = this->${eventThrottleCounterName(event.getName)}++;
+                |U64 throttle_time_us = this->${eventThrottleTimeName(event.getName)}.load();
```
bocchino: As discussed, let's use a pair of U32 values instead of a U64 value.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentEvents.scala
```diff
@@ -271,6 +284,34 @@ case class ComponentEvents (
...
+                |       Fw::Time(static_cast<U32>(throttle_time_us / 1000000), static_cast<U32>(throttle_time_us % 1000000)),
+                |       _logTime
+                |    ) >= Fw::TimeInterval($seconds, $useconds))
+                |  ) {
+                |  // Reset the throttle count & timeout
+                |  this->${eventThrottleTimeName(event.getName)} = (
+                |    (static_cast<U64>(_logTime.getSeconds()) * 1000000) +
+                |    static_cast<U64>(_logTime.getUSeconds())
+                |  );
+                |  this->${eventThrottleCounterName(event.getName)} = 1;
```
bocchino: This assignment can cause a lost update. All the updating in the zero to one case has to be done in the atomic fetch and add instruction.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentEvents.scala
```diff
@@ -271,6 +284,34 @@ case class ComponentEvents (
...
+                |}
+                |else if (count >= ${eventThrottleConstantName(event.getName)}) {
+                |  // Throttle this event
+                |  this->${eventThrottleCounterName(event.getName)}--;
```
bocchino: All the updating has to be done in the atomic fetch and add instruction.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentEvents.scala
bocchino: Here is a suggested algorithm:
```
atomic throttleCount = 0
atomic startTime = 0

emitThrottledEvent() {
  currentTime = getCurrentTime()
  if (throttleCount < maxThrottleCount) {
    incrementThrottleCount(currentTime)
  }
  else if (currentTime - startTime >= throttleInterval) {
    throttleCount = 0
    incrementThrottleCount(currentTime)
  }
}

incrementThrottleCount(currentTime) {
  prevThrottleCount = throttleCount++
  if (prevThrottleCount == 0) {
    startTime = currentTime
  }
  if (prevThrottleCount < maxThrottleCount) {
    emitEvent()
  } 
}
```

bocchino: It seems to me that the code gen in the case where there's no throttle interval can be the same, except that we omit the `else if` clause in the first function.

bocchino: This algorithm assumes that the code for `incrementThrottleCount` is guaranteed to run in a shorter amount of time than the throttle interval. This seems like a reasonable assumption.

bocchino: I think this algorithm still has a race condition. We may need to use `exchange` to store the zero value, and check the previous value. Otherwise two threads can read the timeout condition and race to write the zero.

Kronos3: @bocchino I'm not sure what do check the `exchange`d value with though. I can see there still is a race condition but I don't see how `exchange` will solve it since both threads will still race to write a `0`.

Kronos3: Ok I think I figured out a way to avoid the race using `compare_exchange` https://en.cppreference.com/w/cpp/atomic/atomic/compare_exchange.html (https://github.com/nasa/fpp/pull/821/commits/4dfc833c5cb34aacc47e866df534ff12be443dc8)

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentEvents.scala
bocchino: Here is another try. `std::atomic` provides compare and swap via `compare_exchange_weak` and `compare_exchange_strong`.
```
atomic throttleCount = 0
atomic startTime = 0

emitThrottledEvent() {
  currentTime = getCurrentTime()
  localStartTime = startTime
  if (throttleCount < maxThrottleCount) {
    incrementThrottleCount(currentTime)
  }
  else if (currentTime - localStartTime >= throttleInterval) {
    if (startTime.compareAndSwap(localStartTime, currentTime)) {
      throttleCount = 0
    }
    incrementThrottleCount(currentTime)
  }
}

incrementThrottleCount(currentTime) {
  prevThrottleCount = throttleCount++
  if (prevThrottleCount < maxThrottleCount) {
    emitEvent()
  } 
}
```

bocchino: It assumes that if the CAS succeeds, another thread won't successfully perform the CAS before the throttle count is set to zero.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentEvents.scala
bocchino: Yes, I think we need `compare_exchange`. See my comment below.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentEvents.scala
bocchino: Another alternative is to use locks. It would be much easier to implement that way. We would have to agree that taking a lock in an event call is OK.

Kronos3:
The locks method seems very straight forward and easy to read. The CAS loop method looks like it keeps retrying until it's able to emit the event once things stabilize.

The main performance bottleneck from locks comes from the need to context switch if taking the lock fails. I have a feeling we won't really see much of an issue because I think in the average case components emit events the same thread. Maybe we can start with the lock method and if it needs tweaking or performance boosts we could look at the CAS lockless method.

> Also note that on some architectures, C++ will wrap the EventData in a mutex instead of a CPU atomic operation meaning you never avoid the lock anyway.

compiler/lib/src/main/scala/analysis/Semantics/Event.scala
```diff
@@ -70,17 +70,17 @@ object Event {
...
-            ))
-          }
         } yield TimeInterval(seconds, useconds.toInt)
       }
       def checkEventThrottle(throttle: AstNode[Ast.EventThrottle]) = {
         for {
           count <- a.getNonnegativeIntValue(throttle.data.count.id)
+          _ <- {
+            if count > 0 then Right(())
+            else Left(SemanticError.InvalidIntValue(
```
bocchino: I changed this to an InvalidEvent error before. The message "invalid value 0. Throttle count must be greater than zero" seems redundant.

Kronos3: Fixed!

PR #830, opened 2025-10-02
Fix interface resolution & use def-cycles

### LINKED ISSUE #829: Importing interfaces into interfaces does not work as expected
This code causes `fpp-check` to crash:
```
interface I1 {

  import I2

}

interface I2 {

}
```
If I reorder the definitions of I1 and I2 in the file, there is no crash. I suspect that when the analysis resolves I1, it is crashing because I2 is not resolved yet. I think the fix is to resolve I2 at that point. It's similar to how we do topology definitions. Also, this is what the spec says to do.

A related issue is that cycle detection seems broken for interface imports. If I run this through `fpp-check`, I get a crash:
```
interface I {

  import I

}
```
I would expect an error that reports an illegal def-use cycle.

Kronos3:
Added three missing tests:
- Use-def cycle checks for interfaces
- Out of order resolution (added to `ok.fpp`)
- Duplicate symbol import

Updated the wiki here https://github.com/nasa/fpp/wiki/Check-Interface-Definitions/4b8dbc9ad330fb4787e356badb86f6889dd36b1b

Fixes #829

### DIFF
--- a/compiler/lib/src/main/scala/analysis/CheckSemantics/CheckInterfaceDefs.scala
+++ b/compiler/lib/src/main/scala/analysis/CheckSemantics/CheckInterfaceDefs.scala
@@ -14,14 +14,29 @@ object CheckInterfaceDefs
     a: Analysis,
     aNode: Ast.Annotated[AstNode[Ast.DefInterface]]
   ) = {
-    val a1 = a.copy(interface = Some(Interface(aNode)))
-    for {
-      a <- super.defInterfaceAnnotatedNode(a1, aNode)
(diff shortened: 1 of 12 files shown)


bocchino:
Looks good! I'll run this version on the F Prime code that I am developing and confirm that it fixes the issue.

bocchino:
Thanks for the quick turnaround!

bocchino (approved):
Looks good! I confirmed that it fixes the issue in the F Prime code that I am developing.

PR #834, opened 2025-10-06
Revise code gen for on-change telemetry

### LINKED ISSUE #833: Autocode for boolean update on change tlm channels fails due to 0 literal for boolean
Image

This should be changed to generate `true` or `false` instead of `1` or `0`.

When changed, please implement ticket  nasa/fprime#4266

bocchino:
* Revise code gen. Use C++11 `= {}` initialization to avoid the issue of whether to write `= 0` or `= false` or anything else.
* Add a unit test to cover Boolean on change telemetry
* Update F Prime framework harness for check-cpp
* Incidental fixes
    * Replace NULL with nullptr
    * Replace flatten with concat

Closes #833.


Kronos3 (approved):
Looks good to me!

PR #836, opened 2025-10-07
Revert "Revise code gen for on-change telemetry"

bocchino:
Reverts nasa/fpp#834

I'm getting a failure running `FppTest`.


bocchino:
I'll make a more targeted change to fix the regression.

PR #837, opened 2025-10-07
Fix regression in on-change telemetry

bocchino:
Fix regression introduced by #834.

I still need to investigate why this code change caused a regression. For now, let's ensure that `fpp/main` is working.


bocchino:
Possibly we could update the `Fw::StringBase(const char*)` constructor to tolerate a null pointer.

PR #845, opened 2025-10-13
Revise code gen for array init

### LINKED ISSUE #793: Promote array initializer_list size check to static_assert with C++14
In #791 we are forced to conform to C++11 in which `std::initializer_list::size` is a `const` rather than `constexpr`. In C++14 this is promoted to `constexpr` which will allow us to make a static_assert

https://github.com/nasa/fpp/blob/ac6df7653b0971d8580c26be9691bc5d1590ee5e/compiler/lib/src/main/scala/codegen/CppWriter/ArrayCppWriter.scala#L350-L359

Depends on https://github.com/nasa/fprime/issues/4031

bocchino:
* Upgrade `fprime-gcc` to C++14 for `check-cpp` tests
* Revise comment before assertion. We will continue to use a dynamic assertion, as stated here: https://github.com/nasa/fpp/issues/793#issuecomment-3398614984.

Closes #793.

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ArrayCppWriter.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ArrayCppWriter.scala
@@ -347,8 +347,7 @@ case class ArrayCppWriter (
           ),
         ),
         CppDoc.Type(s"$name&"),
-        lines("""|// Since we are required to use C++11, this has to be a runtime check
-                 |// In C++14, it can be a static check
+        lines("""|// Check that the initializer has the expected size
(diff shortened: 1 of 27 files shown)


compiler/lib/src/main/scala/codegen/CppWriter/ArrayCppWriter.scala
```diff
@@ -347,8 +347,7 @@ case class ArrayCppWriter (
           ),
         ),
         CppDoc.Type(s"$name&"),
-        lines("""|// Since we are required to use C++11, this has to be a runtime check
-                 |// In C++14, it can be a static check
```
Kronos3, line 351: Ah right this makes sense since it's not a template or inline function so it can't give it a compile-time value.

Kronos3 (approved):
Looks good!

PR #849, opened 2025-10-16
Fix dictionary encoding of event throttles

Kronos3:
Event throttle used to be dumped to the dictionary as the throttle count, this updates the encoding to an object that includes the reset interval (`every`). @jwest115 Can you take a look to see if this will need a GDS update?

### DIFF
--- a/compiler/tools/fpp-to-dict/test/top/multipleTops.fpp
+++ b/compiler/tools/fpp-to-dict/test/top/multipleTops.fpp
@@ -133,8 +133,6 @@ module Module1 {
       format "Arg1 is {f}"
 
     @ Event with throttle
-    @ Multiple lines of annotation
-    @ And not used on purpose
     event Event4(
       arg1: F64 @< description of arg1 formal param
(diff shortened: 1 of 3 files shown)


compiler/tools/fpp-to-dict/test/top/multipleTops.fpp
```diff
@@ -132,6 +132,27 @@ module Module1 {
       severity activity high \
       format "Arg1 is {f}"
 
+    @ Event with throttle
+    @ Multiple lines of annotation
+    @ And not used on purpose
```
bocchino: Do we need these annotation lines? It looks like they were copy-pasted from an existing event.

bocchino: @jwest115 Can you comment on these lines?

jwest115: I agree, looks like copy and paste and not needed.

Issue #762, opened 2025-07-23
Add a `dictionary` qualifier to type and constant definitions


bocchino:
This will provide a general way for users to say that they want particular types, constants, and enums exported to the dictionary. For example, one could say
```
dictionary constant SpacecraftId = 0xABCD1234
```

Any FPP model that needs to interact with a GDS feature requiring a SpacecraftId constant (e.g., CCDSD packets) can include such a definition. This way we don't have to maintain a list of "optional dictionary symbols" in the FPP analyzer, and we don't have to update the analyzer every time someone needs a new symbol in the dictionary.

If a definition is marked `dictionary`, then it is always reported as a dependency, so we don't fall into this trap again: #738.

Core F Prime dictionary types, such as `FwCommandOpcodeType`, will still be implied uses according to the existing mechanism (and eventually we should do full type checking for these implied uses). The feature described here is for definitions that may or may not be part of the configuration of any F Prime deployment.

- [x] Update spec
- [x] Update syntax
- [x] Update wiki
- [x] Dictionary definition semantics
- [x] Add/update tests
- [x] Update user guide

thomas-bc:
@thomas-bc 

needed for https://github.com/nasa/fprime/issues/3664

PR #853, opened 2025-10-22
Dictionary specifier syntax, semantics, and tests

jwest115:
Syntax and semantics for dictionary specifier related to https://github.com/nasa/fpp/issues/762
- Added `dictionary` specifier which can be used when defining types/constants (if `dictionary` is used, the type/constant will be output to the dictionary)
- Added `dictionary` location specifier for type and constant locations
- Added new semantic check `CheckDictionaryDefs` for dictionary definitions
- Updated dependency analysis to include dictionary location specifier files in the analysis dependency file set
- Updated analysis to contain a dictionary symbol set (for storing dictionary definition symbols, used by the dictionary)
- Updated codegen
- Updated/added test cases

User guide updates are included in separate PR, see https://github.com/nasa/fpp/pull/854

### DIFF
--- a/compiler/lib/src/main/scala/analysis/UsedSymbols.scala
+++ b/compiler/lib/src/main/scala/analysis/UsedSymbols.scala
@@ -3,8 +3,32 @@ package fpp.compiler.analysis
 import fpp.compiler.ast.*
 import fpp.compiler.util.*
 
-/** Compute used symbols */
+/**
+ *  Compute used symbols
+ *
+ *  There are two forms of resolution:
+ *
+ *  1. Shallow resolution (don't follow uses from uses). This is used to
+ *     generate header files. You get this by calling a visitor method
+ *     of UsedSymbols.
+ *
+ *  2. Deep resolution (follow uses from uses). This is used to generate
+ *     dictionary symbols. You get this by calling UsedSymbols.resolveUses.
+ */
 object UsedSymbols extends UseAnalyzer {
+
+  // When resolving uses, don't visit the default value of an enum definition.
+  // The default value is an enum constant, which is ignored (for shallow resolution)
+  // or converted to a use of this enum (for deep resolution).
+  // So it adds nothing to the resolution.
+  override def defEnumAnnotatedNode(a: Analysis, node: Ast.Annotated[AstNode[Ast.DefEnum]]) = {
+    val (_, node1, _) = node
+    val data = node1.data
+    for {
+      a <- opt(typeNameNode)(a, data.typeName)
+      a <- visitList(a, data.constants, defEnumConstantAnnotatedNode)
+    } yield a
+  }
 
   override def componentUse(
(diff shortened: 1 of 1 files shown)


compiler/lib/src/main/scala/analysis/CheckSemantics/CheckSpecLocs.scala
```diff
@@ -18,19 +18,44 @@ object CheckSpecLocs
...
+  }
+
+  override def defAliasTypeAnnotatedNode(a: Analysis, aNode: Ast.Annotated[AstNode[Ast.DefAliasType]]) = {
+    val (_, node, _) = aNode
+    val name = node.data.name
+    val qualName = a.getQualifiedName(Symbol.AliasType(aNode))
+    val id = aNode._2.id
+    for {
+      a <- checkDictionarySpecLoc(a, Ast.SpecLoc.Type, qualName, id, node.data.isDictionaryDef)
+    } yield a
```
jwest115: Alias types weren't initially covered in this pass. Do we need to run the location specifier check for aliases? Currently, I only included the dictionary location specifier check.

compiler/lib/src/main/scala/codegen/XmlFppWriter/ArrayXmlFppWriter.scala
```diff
@@ -59,7 +59,8 @@ object ArrayXmlFppWriter extends LineUtils {
           AstNode.create(Ast.ExprLiteralInt(xmlSize.text)),
           AstNode.create(eltType),
           fppDefaultsOpt,
-          fppFormatOpt.map(AstNode.create(_))
+          fppFormatOpt.map(AstNode.create(_)),
+          false
```
jwest115, line 63: For XML related codegen I provide array, enum, struct, etc. AST with a default value of "false" for whether the definition is a dictionary definition. I think this should be fine?

compiler/lib/src/main/scala/analysis/CheckSemantics/CheckSpecLocs.scala
jwest115: Add in check for alias type location specifiers

compiler/lib/src/main/scala/codegen/XmlFppWriter/ArrayXmlFppWriter.scala
bocchino: Yes, this sounds good!

compiler/lib/src/main/scala/analysis/CheckSemantics/CheckSpecLocs.scala
bocchino: This is done!

bocchino (approved):
Looks good! I revised the logic for resolving uses of enum constant symbols (this was tricky) and added some tests.

PR #855, opened 2025-10-22
Topology ports semantics

### LINKED ISSUE #745: Export ports in topology semantics
- [ ] Update data structures and algorithms on the wiki
- [ ] Implement
- [ ] Test

Kronos3:
This PR implements topology port semantics according to the spec from #748

### Interesting Changes

- Topology uses and component instance uses were merged into a single `interfaceInstanceUse`
- The component's `portMap` + `specialPortMap` were moved out into a separate `PortInterface` construct so that it could be reused inside topology
- A new port instance case class was added `PortInstance.Topology` for wrapping ports with a new name to impement topology ports.
- SpecLoc.Instance (was for component instance) and SpecLoc.Topology was merged into SpecLoc.Instance (interface instance). (This required a spec update).

Closes #745

### DIFF
--- a/docs/fpp-spec.html
+++ b/docs/fpp-spec.html
@@ -12412,9 +12412,14 @@ definition</a> or
 <div class="sect2">
 <h3 id="Ports_Sub-Interfaces">22.4. Sub-Interfaces</h3>
 <div class="paragraph">
-<p>An interface <em>I</em> is a <strong>sub-interface</strong> of an interface <em>I'</em> if
+<p>A port interface <em>I</em> is a <strong>sub-interface</strong> of a port interface <em>I'</em> if
 the set of port instances represented by <em>I</em> is a subset of the set of
-port instances represented by <em>I'</em>.</p>
+port instances represented by <em>I'</em>.
+For purposes of this comparison, two port instances <em>i</em> and <em>i'</em>
+are considered
+to be identical if they are defined in different places in the model
+but are otherwise identical.
+In this case we say that <em>i</em> and <em>i'</em> have the same <strong>signature</strong>.</p>
 </div>
 </div>
 </div>
@@ -12528,7 +12533,7 @@ equivalent.</p>
 </div>
 <div id="footer">
 <div id="footer-text">
-Last updated 2026-01-28 22:03:08 -0800
+Last updated 2026-02-19 17:19:51 -0800
 </div>
 </div>
 <script src="code-prettify/run_prettify.js"></script>
--- a/docs/fpp-users-guide.html
+++ b/docs/fpp-users-guide.html
@@ -16305,7 +16305,7 @@ serialized according to its
 </div>
 <div id="footer">
 <div id="footer-text">
-Last updated 2026-01-28 22:45:16 -0800
+Last updated 2026-02-19 17:20:19 -0800
 </div>
 </div>
 <script src="code-prettify/run_prettify.js"></script>
--- a/docs/index.html
+++ b/docs/index.html
@@ -4,7 +4,7 @@
 <meta charset="UTF-8">
 <meta http-equiv="X-UA-Compatible" content="IE=edge">
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
-<meta name="generator" content="Asciidoctor 2.0.23">
+<meta name="generator" content="Asciidoctor 2.0.20">
 <title>F Prime Prime (FPP)</title>
 <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Open+Sans:300,300italic,400,400italic,600,600italic%7CNoto+Serif:400,400italic,700,700italic%7CDroid+Sans+Mono:400,700">
 <style>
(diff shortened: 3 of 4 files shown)


compiler/lib/src/main/scala/analysis/Semantics/Symbol.scala
```diff
@@ -72,4 +72,6 @@ object Symbol {
     override def getUnqualifiedName = node._2.data.name
   }
 
+  type InterfaceInstance = Symbol.ComponentInstance | Symbol.Topology
```
bocchino: As discussed, for consistency, let's make `InterfaceInstance` a sealed trait if possible.

compiler/tools/fpp-check/test/connection_direct/invalid_port_instance.ref.txt
```diff
@@ -2,4 +2,4 @@ fpp-check
 [ local path prefix ]/compiler/tools/fpp-check/test/connection_direct/invalid_port_instance.fpp:6.7
     c.out -> c.in
       ^
-error: out is not a port instance of component C
+error: out is not a port instance of interface c
```
bocchino: 1. Should this error message say `interface instance c`?
2. Can we provide a note here saying that c has type C? Otherwise there is a small regression in the quality of the error output.

compiler/tools/fpp-check/test/connection_direct/undef_instance.ref.txt
```diff
@@ -3,4 +3,4 @@ fpp-check
     c.out -> c.in
     ^
 error: symbol c is not defined
-note: looking for a component instance here
+note: looking for a port interface instance here
```
bocchino: Can this error message say "looking for a component instance, topology, or interface parameter here"? I think that would be clearer.

compiler/tools/fpp-check/test/port_matching/p2_not_port_instance.ref.txt
```diff
@@ -2,4 +2,4 @@ fpp-check
 [ local path prefix ]/compiler/tools/fpp-check/test/port_matching/p2_not_port_instance.fpp:6.17
   match p1 with p2
                 ^
-error: p2 is not a port instance of component C
+error: p2 is not a port instance of interface C
```
bocchino: This change to the error message doesn't look right. A matching specifier occurs in a component definition, not an interface definition.

compiler/tools/fpp-locate-uses/test/stdin.ref.txt
```diff
@@ -38,8 +40,6 @@ locate state machine C1.S at "defs.fpp"
 locate state machine M.C1.S at "defs.fpp"
 locate state machine M.S at "defs.fpp"
 locate state machine S at "defs.fpp"
-locate topology M.T1 at "defs.fpp"
```
bocchino, line 43: It looks like we have removed location specifiers for topologies. Why have we done that? We still have topology symbols, right? So I think we should still have separate location specifiers for topologies and component instances (even though they are both interface instances). I think it's clearer that way.

compiler/lib/src/main/scala/analysis/Semantics/PortInstance.scala
```diff
@@ -43,6 +43,18 @@ sealed trait PortInstance {
     loc: Location // The location whether the connection is requested
   ): Result.Result[Unit] = Right(())
 
+  override def equals(obj: Any): Boolean =
```
bocchino: Why do we need to override equals here? What happens if we use the default equals?

compiler/tools/fpp-locate-uses/test/stdin.ref.txt
bocchino: Maybe the issue is that we have put component instances and topologies into the same name group, and the location specifier is tracking the name group?

bocchino: I think the way it is now is correct, based on the spec.

compiler/tools/fpp-check/test/connection_direct/invalid_port_instance.ref.txt
bocchino: In general for this kind of error, `c` has to be a topology `T`, a component instance with component type `C`, or an interface parameter with interface `I`, right? Could the error message say "... is not a port instance of topology `T`/component `C`/interface `I`"? That would be clearer and would preserve what we had before.

compiler/lib/src/main/scala/analysis/Semantics/PortInstance.scala
Kronos3: This is required because we need a slightly more permissive `==` compared to default. This is so that we can compute port interfaces as sets of port instances. The default rule will also check the `node.id` which is not relevant for port instance but just for error reporting.

bocchino: That makes sense, thanks!

bocchino: I think what we are doing here is defining an equivalence relation that says that two port instance values are the same if they represent the same port instance in the context of a single interface. However, I'm a bit uncomfortable with the way the code is written -- it imposes a global equality relation to encode a relation that only makes sense locally. That could cause confusion in another context where we need or expect the true value equality relationship, and it's been overridden. Let's discuss offline.

bocchino: Addressed in commit 8eb58c89aeb3b2425f6143bca364b6c683d4136a.

Issue #840, opened 2025-10-10
Implement code generation for direct port calls


bocchino:
Implement the code generation for https://github.com/nasa/fprime/issues/3986.

* Add a compilation switch `FW_DIRECT_PORT_CALLS`.
* When the switch is not defined, use the standard F Prime ports. This approach can support unit testing.
* When the switch is defined, use direct port calls. Move the implementation of port invocation from the component module to the topology module, and use the topology model to generate a switch statement that directly calls the receiver functions in the target components.

This issue includes the FPP unit tests.

PR #856, opened 2025-10-28
Revise component code gen to support direct port calls

bocchino:
This is the first PR in support of #840 (direct port calls). To make the review process easier, I plan to break #840 into several PRs.

* Refactor the C++ code generation for component dictionary elements (commands, events, telemetry, parameters).
* Revise the code generation to remove direct references to port objects and replace them with calls to functions that can be re-implemented without port objects. As part of this change, make the port invocation functions for data products `private` instead of `protected` (they are for the generated code only; users should not call them directly).
* Add guards based on the preprocessor variable `FW_DIRECT_PORT_CALLS`. These guards surround the code that will be unused or reimplemented when direct port calls are enabled.

## Suggestions for Review

* As part of the refactoring, I have restructured the Scala code to break functions up into smaller pieces and to clean up the code. The resulting code should be easier to understand and maintain. However, a few of the Scala files have a large number of changes. I recommend reviewing those files as a whole, rather than attempting to review the diff in the PR. The unit tests will make sure that the output is correct.
* I recommend reviewing the PR diff of the unit test reference output, to confirm that the changes to the generated code are correct.

## Testing

This branch passes `./cpp-check` and `FppTest` against this commit of nasa/fprime: https://github.com/bocchino/fprime/commit/5ff2bd6acc7b09d15cdcd0842dbc8318311ec24d.

## Work To Go

To close out #840, I need to add the code generation on the topology side that will fill in the missing code when direct port calls are enabled.

### DIFF
--- a/compiler/README.adoc
+++ b/compiler/README.adoc
@@ -173,6 +173,9 @@ Occasionally when developing FPP, you may need to re-run the tracing
 agent to capture new runtime behavior.
 To do that, use the following procedure:
 
+. Ensure that `native-image` is installed and up-to-date,
+as described above.
+
 . Run `./install-trace` to install JVM versions of the FPP tools with
 tracing enabled in the `bin` directory.
 
@@ -189,7 +192,14 @@ is an errant `.lock` file left behind by the tracing agent.
 Run `find . -name .lock` in this directory to find the file;
 then delete it and retry.
 
-. Commit the updated trace files to the repository.
+. Set the environment variable `FPRIME` to point to the root of
+a local copy of the
+https://github.com/nasa/fprime[F Prime repository] or
+a fork of the repository.
+
+. Run `./trace-fprime` to trace F Prime projects that run in CI.
+
+. Commit the updated trace files to this repository.
 
 . Run `./install` to reinstall the standard version
 of the tools.
--- a/compiler/lib/src/main/resources/META-INF/native-image/reflect-config.json
+++ b/compiler/lib/src/main/resources/META-INF/native-image/reflect-config.json
@@ -9,9 +9,6 @@
   "name":"[Lio.circe.Encoder;"
 },
 {
-  "name":"[Ljava.io.File;"
-},
-{
   "name":"[Ljava.lang.String;"
 },
 {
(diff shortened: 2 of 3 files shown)


compiler/tools/fpp-to-cpp/test/component/base/ActiveAsyncProductPortsOnlyComponentAc.ref.cpp
```diff
@@ -367,54 +385,6 @@ void ActiveAsyncProductPortsOnlyComponentBase ::
   // Default: no-op
 }
 
-// ----------------------------------------------------------------------
-// Invocation functions for special output ports
-// ----------------------------------------------------------------------
-
-void ActiveAsyncProductPortsOnlyComponentBase ::
```
LeStarch, line 374: Where did these go?

compiler/tools/fpp-to-cpp/test/component/base/ActiveAsyncProductsComponentAc.ref.cpp
```diff
@@ -4439,6 +4495,78 @@ void ActiveAsyncProductsComponentBase ::
   );
 }
 
+#if !FW_DIRECT_PORT_CALLS
```
LeStarch, line 4498: This file has them moved here, but the other file just deleted them.

compiler/tools/fpp-to-cpp/test/component/base/ActiveEventsComponentAc.ref.cpp
```diff
@@ -2963,16 +3061,14 @@ void ActiveEventsComponentBase ::
 {
   // Get the time
   Fw::Time _logTime;
-  if (this->m_timeGetOut_OutputPort[0].isConnected()) {
-    this->m_timeGetOut_OutputPort[0].invoke(_logTime);
+  if (this->isConnected_timeGetOut_OutputPort(0)) {
+    this->timeGetOut_out(0, _logTime);
   }
 
-  FwEventIdType _id = static_cast<FwEventIdType>(0);
```
LeStarch, line 2970: Interesting

compiler/tools/fpp-to-cpp/test/component/base/QueuedTestComponentAc.ref.hpp
```diff
@@ -2946,6 +2950,90 @@ class QueuedTestComponentBase :
...
     );
 
+  private:
+
+    // ----------------------------------------------------------------------
+    // Invocation functions for special output ports
+    // ----------------------------------------------------------------------
+
+    //! Invoke output port cmdRegOut
+    void cmdRegOut_out(
```
LeStarch, line 2960: Perhaps I missed it earlier....but aren't these to be gated by `#if !FW_DIRECT_PORTS` or is this only in the C++.  Logically it is only the C++.

compiler/tools/fpp-to-cpp/test/component/base/ActiveAsyncProductPortsOnlyComponentAc.ref.cpp
LeStarch: Based on the patterns I have seen, this should have moved into a `#if !FW_DIRECT_PORT` block, but I just see the deletion.

compiler/tools/fpp-to-cpp/test/component/base/QueuedTestComponentAc.ref.hpp
LeStarch: Nevermind

LeStarch (commented):
Ok, I have looked through the code samples and Scala up to my limited ability.  The only real thing amiss is that a pair of functions that I expected to see moved, was removed entirely (see comment).

Were these erroneous to begin with, or should they have moved?

As a side note, this PR was really hard to review given the number and size of changes.

compiler/tools/fpp-to-cpp/test/component/base/ActiveAsyncProductPortsOnlyComponentAc.ref.cpp
bocchino: That component has data product ports but no data products. In the new code gen, the data product invoke functions are private, and they are present only if there are data products.

bocchino:
> As a side note, this PR was really hard to review given the number and size of changes.

Noted, thanks for the thorough review! I will try to make the next one smaller.

PR #860, opened 2025-11-03
Dictionary definitions

### LINKED ISSUE #762: Add a `dictionary` qualifier to type and constant definitions

bocchino:
Add dictionary definitions. Closes #762.

### DIFF
--- a/compiler/tools/fpp-to-dict/test/top/DictionaryDefsTopologyDictionary.ref.json
+++ b/compiler/tools/fpp-to-dict/test/top/DictionaryDefsTopologyDictionary.ref.json
@@ -11,6 +11,22 @@
   },
   "typeDefinitions" : [
     {
+      "kind" : "array",
+      "qualifiedName" : "A",
+      "size" : 3,
+      "elementType" : {
+        "name" : "U32",
+        "kind" : "integer",
+        "size" : 32,
+        "signed" : false
+      },
+      "default" : [
+        0,
+        0,
+        0
+      ]
+    },
+    {
       "kind" : "alias",
       "qualifiedName" : "FwTlmPacketizeIdType",
       "type" : {
@@ -56,20 +72,6 @@
     },
     {
       "kind" : "alias",
-      "qualifiedName" : "f32Alias",
-      "type" : {
-        "name" : "F32",
-        "kind" : "float",
-        "size" : 32
-      },
-      "underlyingType" : {
-        "name" : "F32",
-        "kind" : "float",
-        "size" : 32
-      }
-    },
-    {
-      "kind" : "alias",
       "qualifiedName" : "BaseIdType",
       "type" : {
         "name" : "U32",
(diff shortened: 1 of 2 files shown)


compiler/tools/fpp-check/test/constant/dictionary_error.ref.txt
```diff
@@ -0,0 +1,5 @@
+fpp-check
+[ local path prefix ]/compiler/tools/fpp-check/test/constant/dictionary_error.fpp:1.1
+dictionary constant c = { x = 0, y = 1 }
+^
+error: dictionary constant must have a numeric, Boolean, string, or enum type
```
Kronos3, line 5: This is pretty minor but why is `Boolean` capital?

compiler/tools/fpp-to-dict/test/top/QualifiedCompInstTopologyDictionary.ref.json
Kronos3, line 1: These dictionary changes seem sort of large, is this just due to ordering not being maintained now that json ending was changed? Seems odd since only the analysis encoder was touched.

Kronos3: Ah nevermind I see the Fw type aliases were changed in the model

docs/users-guide/Dictionary-Definitions.adoc
```diff
@@ -0,0 +1,70 @@
...
+For example, a downlink configuration parameter may be
+shared by the FSW implementation and the GDS and may be otherwise unused
+in the FPP model.
+
+In this case you can mark a type or constant definition as a *dictionary 
+definition*.
+A dictionary definition tells the FPP analyzer two things:
+
+. The definition should be included in the model, even if it
+isn't used anywhere in the model.
```
Kronos3: Can you elaborate on what this means practically? Does it affect anything from the user's standpoint? The bullet below seems to be the major delta here.

compiler/tools/fpp-check/test/constant/dictionary_error.ref.txt
bocchino: Because it's named after George Boole. https://en.wikipedia.org/wiki/Boolean_data_type

I guess sometimes Boolean is capitalized and sometimes not. Looks like Wikipedia capitalizes it.

abelian usually isn't capitalized even though it's named after Abel. https://en.wikipedia.org/wiki/Abelian_group

docs/users-guide/Dictionary-Definitions.adoc
bocchino: I don't think it has much practical effect on the user. It is broader than what we need (dictionary defs are included in models even when there's no topology or dictionary) but this is for simplicity, so we don't have to implement and test the logic for deciding whether to include them. In practice there will be a few extra dependencies on types and constants that are part of the configuration of the system and should rarely change anyway. Do you have a suggestion for something else to say here?

Kronos3: I think then its more of spec-related detail. Maybe just keep it simple and collapse these two bullets into a cursory explanation to keep the users guide more focused.

bocchino: OK, I made this revision.

PR #865, opened 2025-11-05
Use C++ zero args constructor when possible

### LINKED ISSUE #851: Ensure that construction of default values is stack safe
FPP should ensure that constructing a default value using the C++ zero-arg constructor is stack safe when the value is not constructed on the stack. For example, suppose we have this model:
```
struct S { x: U32 } default { x = 42 }
struct T { s: S }
array A [3] T
```
Then the constructor `S()` should look like this:
```
S() : m_x() { this->m_x = 42; }
```
And the constructor `T()` should look like this:
```
T() : m_s() {}
```
And in the class for `A`, the member array of 3 `T` objects should be initialized `= {}`, and the constructor `A()` should do nothing.

Currently we are not quite doing this. Sometimes, during initialization, we create temporary objects that can overflow the stack if the arrays and structs get large.

Thanks to @LeStarch for pointing out this issue.

Kronos3:
This PR reworks the default constructor codegen for array and structs. It utilizes the C++ zero args constructor on members who's constructor is sufficient to initialize the struct/array member to the proper value.

- [x] check-cpp
- [x] FppTestProject

Closes #851

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/StructCppWriter.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/StructCppWriter.scala
@@ -768,7 +768,7 @@ case class StructCppWriter(
               case _ => Nil
             }
           },
-          lines(s"this->m_$memberName[i] = ${memberValue};")
+          lines(s"this->m_$memberName[i] = $memberValue;")
         )
       )
   }


bocchino (approved):
Looks good! I made one change to the array code gen. It was still using the stack for a default array of strings. That's fixed now.

PR #867, opened 2025-11-09
Improve error message for undefined symbols

### LINKED ISSUE #778: Type mismatch in FPP results in "undefined symbol" error
| | |
|:---|:---|
|**_F´ Version_**| v4.0.0a1 |
|**_Affected Component_**| fpp |
---
## Problem Description

When compiling broken code like the following:

```
    enum ExampleEnum {
        # ...
    }
    passive component ExampleComponent {
        # ...
        sync input port examplePort: ExampleEnum
    }
```

FPP gives this error:

```
[build] fpp-to-cpp
[build] /workspaces/[...]/ExampleComponent.fpp:56.37
[build]         sync input port examplePort: ExampleEnum
[build]                                     ^
[build] error: undefined symbol ExampleEnum
```

This error message is misleading because the symbol _is_ defined... it's just the totally wrong type. An enum was specified instead of a port, and that's what's wrong, not a typo in the name.

## Expected Behavior

The error message should identify that the problem is a type mismatch.

I didn't realize that it was a type mismatch when I first saw this error, because I assumed that the "undefined symbol" error was caused by https://github.com/nasa/fprime/issues/2999. It wasn't until I did a full rebuild and it still failed that I realized I'd forgotten to define the port I wanted.

## Context / Environment

```
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-4.18.0-553.58.1.el8_10.x86_64-x86_64-with-glibc2.28
Python version: 3.9.20
CMake version: 3.24.2
Pip version: 25.1.1
Pip packages:
    fprime-tools==4.0.0a1
    fprime-gds==3.6.2a2.dev14+gabd041b
    fprime-fpp-*==3.0.0a10
Project submodules:
    https://github.com/nasa/fprime.git @ v4.0.0a1-1-g8e64db70e
    https://github.com/fprime-community/fprime-baremetal @ ca90b58
```

Kronos3:
This copies the same error message style Rust uses for their undefined symbols:
Screenshot 2025-11-09 at 9 01 51 AM

Closes #778

### DIFF
--- a/compiler/tools/fpp-check/test/component/array_undef_constant.ref.txt
+++ b/compiler/tools/fpp-check/test/component/array_undef_constant.ref.txt
@@ -2,5 +2,5 @@ fpp-check
 [ local path prefix ]/compiler/tools/fpp-check/test/component/array_undef_constant.fpp:2.29
   array A = [3] U32 default a
                             ^
-error: cannot find symbol `a`
+error: symbol a is not defined
 note: looking for a constant here
--- a/compiler/tools/fpp-check/test/component/array_undef_type.ref.txt
+++ b/compiler/tools/fpp-check/test/component/array_undef_type.ref.txt
@@ -2,5 +2,5 @@ fpp-check
 [ local path prefix ]/compiler/tools/fpp-check/test/component/array_undef_type.fpp:2.17
   array A = [3] T
                 ^
-error: cannot find symbol `T`
+error: symbol T is not defined
 note: looking for a type here
--- a/compiler/tools/fpp-check/test/component/enum_undef_constant.ref.txt
+++ b/compiler/tools/fpp-check/test/component/enum_undef_constant.ref.txt
@@ -2,5 +2,5 @@ fpp-check
 [ local path prefix ]/compiler/tools/fpp-check/test/component/enum_undef_constant.fpp:2.16
   enum E { X = a }
                ^
-error: cannot find symbol `a`
+error: symbol a is not defined
 note: looking for a constant here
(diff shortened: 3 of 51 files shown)


compiler/lib/src/main/scala/util/Error.scala
```diff
@@ -337,8 +336,8 @@ sealed trait Error {
         System.err.println("for this component instance:")
         System.err.println(instanceLoc)
       case SemanticError.TypeMismatch(loc, msg) => Error.print (Some(loc)) (msg)
-      case SemanticError.UndefinedSymbol(name, loc) =>
-        Error.print (Some(loc)) (s"undefined symbol ${name}")
+      case SemanticError.UndefinedSymbol(name, symbolKind, loc) =>
+        Error.print (Some(loc)) (s"cannot find $symbolKind `${name}` in this scope")
```
bocchino: This is good, but I think we need to revise the error message slightly. Can we say
```
cannot find symbol $symbolKind
```
And add a note saying
```
note: looking for a $symbolKind here
```
?

Here is my reasoning:
1. The phrase "in this scope" doesn't add any useful information to the error message
2. We can't say "cannot find $symbolKind $name" in general, because that can produce error messages like
    ```
    async port p: Fw.FooBar
                  ^
    error: cannot find port Fw
    ```
    But here `Fw` is a module symbol (in the Port namespace) and not a port symbol. This is especially important for implied dependencies, e.g., when a command specifier implicitly depends on a port `Fw.Cmd`. We shouldn't say `error: can't find port Fw` here.

Kronos3: Ah good call, I'll fix that.

PR #872, opened 2025-11-11
Revise installation procedure

bocchino:
This PR revises the installation procedure for machines that don't have Java 11 installed as the default java. The old installation procedure was not correct; it depended on having `$JAVA_HOME/bin` first in the `PATH`, but it did not say that. The new procedure does not require any modification to `PATH`.

### DIFF
--- a/compiler/install
+++ b/compiler/install
@@ -10,7 +10,7 @@
 # To pass flags to sbt, set the environment variable
 # FPP_SBT_FLAGS
 #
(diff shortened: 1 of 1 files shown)


Kronos3 (approved):
This looks good to me. I left both `FPP_GRAALVM_JAVA_HOME` and `GRAALVM_JAVA_HOME` set work backwards with branches not in sync.

PR #875, opened 2025-11-12
Revise dependency analysis for dictionary defs

### LINKED ISSUE #873: Some `dictionary` constants do not appear in the dictionary
## How-To Reproduce

1. checkout latest `fprime@devel`
2. Mark some of the constants in `config/FpConstants.fpp` with `dictionary`
3. Build `Ref/`
4. The `FpConstants.fpp` constants marked `dictionary` do not appear in the Ref dictionary

## Issue

@jwest115 has tracked down the issue to be due to the fact that `FpConstants.fpp` is not a dependency of `Ref`. This is because these constants are only used in the C++ layer. 
From a user perspective this is unpredictable and does not seem to match the [specification](https://nasa.github.io/fpp/fpp-spec.html#Definitions_Dictionary-Definitions) which does not call this as a caveat of the specifier.

EDIT: the bug is likely in the integration of this in the build system. Not in the tool itself. Maybe should move this ticket over to core `fprime`

bocchino:
Closes #873.

Wiki is updated here: https://github.com/nasa/fpp/wiki/Computing-Dependencies.

@thomas-bc @jwest115 Can you confirm that this fixes #873?

### DIFF
--- a/compiler/lib/src/main/scala/analysis/ComputeDependencies/AddDependencies.scala
+++ b/compiler/lib/src/main/scala/analysis/ComputeDependencies/AddDependencies.scala
@@ -51,7 +51,6 @@ object AddDependencies extends BasicUseAnalyzer {
     analyzeUse(a, Ast.SpecLoc.Type, use)
 
   private def analyzeUse(a: Analysis, kind: Ast.SpecLoc.Kind, use: Name.Qualified): Result = {
-
     def computeNameList: List[Name.Qualified] = {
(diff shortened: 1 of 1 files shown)


jwest115 (approved):
I confirmed this addresses the issue reported in https://github.com/nasa/fpp/issues/873 by doing the following:
- Checked out latest fprime@devel
- Marked a constant in config/FpConstants.fpp with dictionary
- Built Ref/
- Verified the constant is now included in the dictionary

Thanks for fixing this!

bocchino:
Thanks, @jwest115!

PR #881, opened 2025-11-18
Topology code for direct port calls with typed ports

bocchino:
This PR adds code generation for direct calls with typed ports.
* Implement code generation
* Add unit tests

Code generation for serial ports is future work.

Further to #840.


bocchino:
A couple of issues to fix, uncovered during integration:

- [ ] The generated code for ports uses `#if FW_DIRECT_PORT_CALLS` before it includes the config header. This works only if `FW_DIRECT_PORT_CALLS` is defined on the command line. We need to support setting it in the config header.
- [ ] The generated code for topologies uses `#ifdef FW_DIRECT_PORT_CALLS` instead of `#if FW_DIRECT_PORT_CALLS`.

bocchino:
Further issues uncovered during integration:

- [ ] The topology code gen is not generating the functions `CComponentBase::p_isConnected` and `CComponentBase::p_out` in the case that (1) component `C` has an output port `p` and (2) an instance of `C` is part of the topology and (3) there is no connection from `c.p` for any instance `c` of `C`. This causes a linker error. We should generate code that runs `ASSERT(0)` in this case, since an unconnected port should not be invoked.
- [ ] The topology code gen does not correctly handle connections to the special input port `cmd recv`. The `handlerBase` calling protocol appears to be different than for general ports. Likely the same issue exists for `product recv`.

PR #888, opened 2025-11-22
Merge release/v3.1.0a12 into release/v3.1.0

bocchino:
This branch fixes a few issues that came up when integrating https://github.com/nasa/fprime/pull/4473.
1. On RHEL, the compiler was not happy about calling default constructors with `{}`; I switched it to `()`.
2. The F Prime tracing was not enabling fpp-to-json, so when fpp-to-json was enabled in CI, there was an explosion. I fixed it.
3. The F Prime tracing was spending time doing builds, but we know those work, so this seemed like wasted effort. As I understand it, all the fpp-to-json work is done in the generate step. So I revised the trace-prime script to do generate only. @LeStarch, please review.

I will make release v3.1.0a12 off of this branch and test it in F Prime CI before merging this PR.


LeStarch (approved):
I am not up-to-speed on the nuance between `()` and `{}` but I confirm this, and the script are the changes.  Both look good to me.

PR #901, opened 2026-01-15
Remove calls to updateHash() from generated code

bocchino:
Tandem with https://github.com/nasa/fprime/pull/4622.


bocchino:
Yes, that sounds reasonable. I think we should write something into those bytes and not leave them uninitialized when building the DP header.

Does the value have to be configurable? Or can it just be specified as a constant in DpContainer? That would be easier to implement and maintain.

thomas-bc (approved):

PR #906, opened 2026-01-26
Type and constant definitions in state machines

### LINKED ISSUE #617: Allow type and constant definitions inside state machines
For example, one could define a type _T_ inside a state machine and use it in the signals of _T_.

- [x] Spec
- [x] Syntax
- [x] Semantics
- [x] Code gen
- [x] FppTest
- [x] User's Guide

bocchino:
Closes #617.

### DIFF
--- a/compiler/lib/src/main/resources/META-INF/native-image/reflect-config.json
+++ b/compiler/lib/src/main/resources/META-INF/native-image/reflect-config.json
@@ -259,6 +259,26 @@
   "fields":[{"name":"0bitmap$1206"}]
 },
 {
+  "name":"fpp.compiler.codegen.AnalysisJsonEncoder$$anon$1315",
+  "fields":[{"name":"0bitmap$1205"}]
+},
+{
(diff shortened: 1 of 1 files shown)


compiler/tools/fpp-to-cpp/test/array/SM_AArrayAc.ref.hpp
```diff
@@ -0,0 +1,171 @@
+// ======================================================================
+// \title  SM_AArrayAc.hpp
+// \author Generated by fpp-to-cpp
+// \brief  hpp file for SM_A array
+// ======================================================================
+
+#ifndef SM_SM_AArrayAc_HPP
+#define SM_SM_AArrayAc_HPP
```
Kronos3: Why do these if guard double qualify `SM`?

This may not be related to this PR (also pretty minor)...

bocchino: Good catch! I fixed it.

Issue #903, opened 2026-01-21
Make default string size configurable in the model


bocchino:
The default string size (i.e., the size used for a type `string` with no size specifier) is unspecified in the model and is hard-set to 80 in the code generator. In theory this size is configurable, but the configuration is brittle and is not used. We should specify the default string size in the model and make it configurable, as follows:

1. Add a constant `FwDefaultStringSize` as a framework definition.
2. Add this constant to the list of constants required by dictionary generation, as specified here: https://fprime.jpl.nasa.gov/latest/docs/reference/fpp-json-dict/#dictionaries.
3. When generating C++ code or JSON code for string types with default sizes, use the symbolic constant `FwDefaultStringSize` as the size.
4. Eliminate the `-s` option in `fpp-to-cpp`, which was a workaround for not having this feature.

This feature is required for https://github.com/nasa/fpp/issues/725. Until we have it, the size of the type `string` is not specified in the model.

To integrate this change with F Prime, we'll have to add a configuration constant. However, an alpha release of FPP should not be required, since adding this constant to F Prime does not require a change to FPP.

bocchino:
@jwest115 I updated the main comment: let's use `FW_FIXED_LENGTH_STRING_SIZE` -- a constant we already have in #908 -- as the default string size in the model. That way we don't have to add another constant.

PR #909, opened 2026-01-28
Make default string size configurable in the model, use in cpp and dictionary codegen

jwest115:
- Updated dictionary spec to require `FW_FIXED_LENGTH_STRING_SIZE` (PR: https://github.com/nasa/fprime/pull/4667)
- Updated cpp and dictionary codegen to use `FW_FIXED_LENGTH_STRING_SIZE` for string size
   - Note: Since this constant is not required by cpp codegen, I kept 80 as the default if the constant is not in the model.
- Removed `-s` option from `fpp-to-cpp` and `fpp-to-dict`
- Updated user guide

Closes https://github.com/nasa/fpp/issues/903.

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentInternalStateMachines.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentInternalStateMachines.scala
@@ -643,7 +643,7 @@ case class ComponentInternalStateMachines(
                   s.a.typeMap(tn.id).getUnderlyingType match {
                     case t: Type.String => (
                       ts + Type.String(None),
-                      max(maxStringSize, getStringSize(s, t))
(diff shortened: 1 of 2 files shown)


compiler/lib/src/main/scala/codegen/DictionaryJsonWriter/DictionaryJsonEncoderState.scala
```diff
@@ -29,17 +27,24 @@ case class DictionaryJsonEncoderState(
     }
   }
 
+  def getFwDefaultStringSize: BigInt = {
```
jwest115: This could probably be in FrameworkDefinitions and return an optional BigInt

bocchino:
This is great work! I'm closing this PR in favor of #913, which is on a branch in the repo. This work is exposing bugs in FppTest, so we need an F Prime tandem development.

PR #913, opened 2026-01-29
Revise modeling and code generation for default string size

### LINKED ISSUE #903: Make default string size configurable in the model

bocchino:
Supersedes #909. I needed to make a new PR with a branch in the repo, because this work is exposing bugs in FPP Test, so this is a tandem development with F Prime.

Tandem development is on this branch: https://github.com/bocchino/fprime/tree/fpp-issue-903-default-string-size.

Closes #903.

### DIFF
--- a/compiler/lib/src/main/resources/META-INF/native-image/reflect-config.json
+++ b/compiler/lib/src/main/resources/META-INF/native-image/reflect-config.json
@@ -259,26 +259,6 @@
   "fields":[{"name":"0bitmap$1206"}]
 },
 {
-  "name":"fpp.compiler.codegen.AnalysisJsonEncoder$$anon$1315",
-  "fields":[{"name":"0bitmap$1205"}]
-},
-{
-  "name":"fpp.compiler.codegen.AnalysisJsonEncoder$$anon$1316",
-  "fields":[{"name":"0bitmap$1201"}]
-},
-{
-  "name":"fpp.compiler.codegen.AnalysisJsonEncoder$$anon$1317",
-  "fields":[{"name":"0bitmap$1202"}]
-},
-{
-  "name":"fpp.compiler.codegen.AnalysisJsonEncoder$$anon$1318",
-  "fields":[{"name":"0bitmap$1203"}]
-},
-{
-  "name":"fpp.compiler.codegen.AnalysisJsonEncoder$$anon$1319",
-  "fields":[{"name":"0bitmap$1204"}]
-},
-{
   "name":"fpp.compiler.codegen.AnalysisJsonEncoder$$anon$1321",
   "fields":[{"name":"0bitmap$1211"}]
 },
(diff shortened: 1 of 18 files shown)


compiler/lib/src/main/scala/codegen/CppWriter/ArrayCppWriter.scala
```diff
@@ -454,18 +454,14 @@ case class ArrayCppWriter (
                  |tmp.format(\"%s\", tmp.toChar());"""
 
     val formatLoop = indexIterator(lines(
-      s"""|Fw::String tmp;
+      s"""|// Array data
```
bocchino, line 457: @Kronos3 I decided to simplify the code generation here. Testing the old code was difficult, and IMO it is over-optimized. If you are hitting this case (trying to write a large array of strings into a much smaller string) then your system is mistuned anyway -- so I don't think we should try to optimize it.

compiler/lib/src/main/scala/codegen/DictionaryJsonWriter/DictionaryJsonEncoderState.scala
```diff
@@ -29,17 +27,24 @@ case class DictionaryJsonEncoderState(
...
+      }
+      case None => throw InternalError("FW_FIXED_LENGTH_STRING_SIZE constant not defined")
+    }
+  }
+
 }
 
 case object DictionaryJsonEncoderState {
 
-  /** The default string size */
```
bocchino, line 36: @jwest115 I completely removed the default default string size. The code generation now depends only on the symbolic constant `FW_FIXED_LENGTH_STRING_SIZE`. That is what we said in #903.

compiler/lib/src/main/scala/codegen/CppWriter/ConstantCppWriter.scala
```diff
@@ -129,7 +129,7 @@ object ConstantCppWriter extends CppWriterUtils {
       val hppLines = {
         val defLine = line(s"$name = $value")
         List(
-          line(s"enum FppConstant_$name {"),
```
bocchino, line 132: I removed the enum type here. It was causing the compiler to complain about comparing enum values with different types.

bocchino: Update: I had to add static casts to remove the warnings on gcc. So removing the enum type isn't necessary. However, I also think that the enum type doesn't really add anything there, so we can remove it.

PR #920, opened 2026-02-11
Fix small errors in spec

kevinthegreat1:
I believe `port interface specifier` in 7.10.1 means `port interface location specifier`. In 15.6., there is no `D`, only `Q`.


bocchino (approved):
Good catches, thanks!

Issue #725, opened 2025-05-27
Add sizeof to FPP


jwest115:
It would be nice if we had a way to use the size of a type in FPP. For instance, in FPrime [config](https://github.com/nasa/fprime/blob/378a7a3e6fccfe665fcb819a7325be1a66b410d0/default/config/FpySequencerCfg.fpp#L11C18-L11C48), there is a constant `MAX_LOCAL_VARIABLE_BUFFER_SIZE` that is equal to FW_COM_BUFFER_MAX_SIZE - the size of FwChanIdType - the size of FwPacketDescriptorType.

`constant MAX_LOCAL_VARIABLE_BUFFER_SIZE = 512 - 4 - 4`

It would be useful if we had the ability to use the size of those types rather than "hardcoding" the sizes in the model.

PR #921, opened 2026-02-19
String type name implied uses

jwest115:
Adds the following implies uses for string type names:
- `FwSizeStoreType` (always required when using strings)
- `FW_FIXED_LENGTH_STRING_SIZE` (only required when strings do not have a size specified)

Updated spec and user guide examples to include implied uses
Updated and regenerated tests

Needed for #725.

### DIFF
--- a/docs/fpp-users-guide.html
+++ b/docs/fpp-users-guide.html
@@ -7720,8 +7720,7 @@ Here is an example:</p>
 </div>
 <div class="listingblock">
 <div class="content">
-<pre class="prettyprint highlight"><code data-lang="fpp"># The following definitions are required when using string types
-type FwSizeStoreType = U16
+<pre class="prettyprint highlight"><code data-lang="fpp">type FwSizeStoreType = U16
 constant FW_FIXED_LENGTH_STRING_SIZE = 256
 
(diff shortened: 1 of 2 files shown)


compiler/lib/src/main/scala/analysis/CheckSemantics/CheckUses.scala
```diff
@@ -192,6 +192,35 @@ object CheckUses extends BasicUseAnalyzer {
...
+    val impliedTypeUses = a.getImpliedUses(ImpliedUse.Kind.Type, node.id).toList
+    val impliedConstantUses = a.getImpliedUses(ImpliedUse.Kind.Constant, node.id).toList
+    for {
+      a <- Result.foldLeft (impliedConstantUses) (a) ((a, iu) => {
+        val exprNode = iu.asExprNode
+        for {
+          a <- Result.annotateResult(
+            constantUse(a, exprNode, iu.name),
+            s"string type names require that the constant ${iu.name} is defined"
+          )
```
jwest115: Match on the implied use to see if its the default string size and tailor the message to that i.e. "string type names without a size require that the constant x is defined"

PR #923, opened 2026-02-19
Add State enum for state machines

### LINKED ISSUE #615: Auto-generate an FPP enumeration for the state of an FPP state machine
The state of an FPP state machine _S_ is currently stored as a C++-only enum. This makes it difficult to report the current state in events and telemetry. As part of the analysis for _S_ we could auto-generate an enum type _S_State_ representing the state of _S_.

bocchino:
This PR adds the State enum for state machines and revises the C++ code gen to match. This way we can use the state enum in FPP events and telemetry. Closes #615.

This PR also does the following:
* Replace classes with objects, to remove parentheses from primitive types in the AST. This is a code simplification.
* Refactor AST state transformer to eliminate duplicated code.
* Move include resolution out of CheckSemantics and into each tool that calls CheckSemantics. This approach is more uniform, and it makes more sense now another syntax pass precedes analysis (AddStateEnums).
* Fix an issue where type defs inside state machines inside components were not named properly.
* Fix an issue in enum code gen where enum constant `T` could shadow enum `T`.

Tandem development for this PR is here: https://github.com/bocchino/fprime/tree/fpp-issue-615-sm-enum.


compiler/lib/src/main/scala/transform/AddStateEnums.scala
```diff
@@ -0,0 +1,131 @@
...
+    in: State,
+    aNode: Ast.Annotated[AstNode[Ast.DefStateMachine]],
+    members: List[Ast.StateMachineMember]
+  ) = {
+    val (pre, node, post) = aNode
+    val name = node.data.name
+    val id = node.id
+    for (stateEnum <- getStateEnum(Symbol.StateMachine(aNode)))
+    yield {
+      val enumNode = AstNode.create(stateEnum, id)
```
Kronos3: I think you should create a new ID for this node instead of re-using the one for the state machine. This will need to be done for templates in the future anyway.

compiler/tools/fpp-to-cpp/test/state-machine/state/BasicGuardTestArrayStateMachineAc.ref.hpp
```diff
@@ -34,19 +35,12 @@ namespace FppTest {
...
         //! The state type
-        enum class State : FwEnumStoreType {
-          //! The uninitialized state
-          __FPRIME_AC_UNINITIALIZED,
-          //! State S
-          S,
-          //! State T
-          T,
-        };
+        using State = FppTest::SmState::BasicGuardTestArray_State;
```
Kronos3, line 38: Wow this codegen diff is remarkably clean!

Kronos3 (commented):
This looks great. I have one minor comment

compiler/lib/src/main/scala/transform/AddStateEnums.scala
bocchino: Done!

PR #930, opened 2026-02-25
Allow zero-sized strings

### LINKED ISSUE #878: Zero Sized Strings
In the spec https://nasa.github.io/fpp/fpp-spec.html#Type-Names_String-Type-Names it says that string sizes must be in the range `[0, 2^31)`, the wiki agrees with this.

The implementation doesn't allow `0`:
https://github.com/nasa/fpp/blob/35ac2c90e923ffdec6b8ed88f4fcbdf87c1de142/compiler/lib/src/main/scala/analysis/CheckSemantics/FinalizeTypeDefs.scala#L222-L226

Not really sure what the implications of `string size 0` are but presumably it could be useful if the `string size configuredConstant` use case is used and you want to disable a field? We should test to make sure compilation works with `string size 0` without any warnings.

CalMorris:
Fixes #878

## Summary

The FPP spec and the published website describe `string size` _n_ as valid in the range `[0, 2^31)`, but the compiler was rejecting `string size 0` with an "invalid string size" error.

This is useful when a configured constant drives the string size and a value of 0 is used to disable a field.

## Changes

- Relax the validation in `FinalizeTypeDefs.scala` from `size > 0` to `size >= 0`
- Update the spec (`Types.adoc`) to document the valid range as `[0,2^31-1]`

## Testing

Ran the existing `fpp-check` type test suite — all 7 tests pass, including `string_size_negative` (rejects -1) and `string_size_too_large` (rejects `0x80000000`).

### DIFF
--- a/compiler/lib/src/main/scala/analysis/CheckSemantics/FinalizeTypeDefs.scala
+++ b/compiler/lib/src/main/scala/analysis/CheckSemantics/FinalizeTypeDefs.scala
@@ -219,7 +219,7 @@ object FinalizeTypeDefs
             a.valueMap(id),
             Type.Integer
           )
-          if (size > 0 && size <= Int.MaxValue) Right(t)
+          if (size >= 0 && size <= Int.MaxValue) Right(t)
           else {
             val loc = Locations.get(id)
             Left(SemanticError.InvalidStringSize(loc, size))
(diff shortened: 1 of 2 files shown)


bocchino:
I'm closing this PR, as it is stale and no work has occurred in a while. We can open it later if necessary.

PR #931, opened 2026-02-26
Fix dictionary JSON struct default values

### LINKED ISSUE #925: Incorrect default value JSON dictionary entry for member arrays of arrays
I believe I have found an issue with the JSON dictionary generated by this FPP:
```
        struct ExampleStruct {
            arrayMemberU32: [2] U32Array,
        }
```

Where U32Array is:
```
array U32Array = [5] U32
```

That generates this JSON:
```
    {
      "kind" : "struct",
      "qualifiedName" : "ExampleStruct",
      "members" : {
        "ex" : {
          "type" : {
            "name" : "U32Array",
            "kind" : "qualifiedIdentifier"
          },
          "index" : 0,
          "size" : 2
        }
      },
      "default" : {
        "ex" : [
          0,
          0,
          0,
          0,
          0
        ]
      }
    },
```

Notice that the default value does not match the actual structure of the member. Instead, I expect something like:

```
      "default" : {
        "ex" : [[
          0,
          0,
          0,
          0,
          0
        ], [
          0,
          0,
          0,
          0,
          0
        ]]
      }
```

jwest115:
- Takes into account struct member sizes when generating struct value dictionary JSON
- Clean up some redundant code for array JSON
- Updated dictionary tests to exercise struct member size case

Closes #925

### DIFF
--- a/docs/fpp-spec.html
+++ b/docs/fpp-spec.html
@@ -3665,23 +3665,32 @@ of the specified type.  Each
 identifier appearing in the struct type member sequence must be distinct.</p>
 </div>
 <div class="paragraph">
-<p>The optional expression <code>[</code> <em>e</em> <code>]</code> for a struct member specifies the
-<strong>size</strong> of the member, i.e., the number of elements stored in the member.
+<p>For each member <em>m</em>,</p>
+</div>
+<div class="ulist">
+<ul>
+<li>
+<p>The type name specifies the type \$T_m\$ of the member.</p>
+</li>
+<li>
+<p>The optional expression <code>[</code> <em>e</em> <code>]</code> specifies the
+<strong>size</strong> of <em>m</em>, i.e., the number of elements stored in <em>m</em>.
 If present, <em>e</em> must have a
 <a href="#Types_Internal-Types_Numeric-Types">numeric type</a>, and it must
-evaluate to a value greater than zero after conversion to
+evaluate to a value \$n &gt; 0\$ after conversion to
 <a href="#Types_Internal-Types_Integer"><em>Integer</em></a>.
-If <em>e</em> is not present, then default size is one.
-If a member <em>m</em> has a size specified, then <em>m</em>
-is translated to an array in the implementation language.</p>
-</div>
-<div class="paragraph">
-<p>The optional format specifier for a struct member specifies a
+If <em>e</em> is not present, then <em>m</em> represents a single value of type \$T_m\$.
+Otherwise <em>m</em> represents an array of <em>n</em> values of type \$T_m\$.</p>
+</li>
+<li>
+<p>The optional format specifier specifies a
 <a href="#Format-Strings">format string</a>.
 The format is applied when displaying the struct member.
 There is one argument to the format string, which is the member value.
 If the size of the member is greater than one, then the
 translator applies the format to each element.</p>
+</li>
+</ul>
 </div>
 <div class="paragraph">
 <p>The expression following the keyword <code>default</code> is optional.
@@ -12647,7 +12656,7 @@ equivalent.</p>
 </div>
 <div id="footer">
 <div id="footer-text">
-Last updated 2026-03-09 11:27:02 -0700
+Last updated 2026-03-09 13:30:08 -0700
 </div>
 </div>
 <script src="code-prettify/run_prettify.js"></script>
--- a/docs/spec/Definitions/Struct-Definitions.adoc
+++ b/docs/spec/Definitions/Struct-Definitions.adoc
@@ -34,17 +34,20 @@ The definition associates the name _N_ with a
 of the specified type.  Each
 identifier appearing in the struct type member sequence must be distinct.
 
-The optional expression `[` _e_ `]` for a struct member specifies the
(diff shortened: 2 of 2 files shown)


bocchino:
This looks good, but now that I see the change, I don't think we want to expand the default values for struct array members. I added a comment to #925. I think we should just write out the default value of the struct, as the value is represented in FPP. If there's some reason that won't work in the GDS, we should try to address it on the GDS side.

compiler/lib/src/main/scala/codegen/DictionaryJsonWriter/DictionaryJsonEncoder.scala
```diff
@@ -197,7 +202,7 @@ case class DictionaryJsonEncoder(
                 val qualifiedName = dictionaryState.a.getQualifiedName(Symbol.Enum(t.node)).toString
                 s"${qualifiedName}.${v._1}".asJson // FQN of the enum constant
             }
-            case Value.Struct(Value.AnonStruct(members), t) => members.map((key, value) => (key.toString -> valueAsJson(value))).asJson
```
bocchino, line 200: As noted, the original code looks more correct here. The new code seems to be a special case to implement a difference between FPP semantics and GDS semantics. I'd rather keep those the same, if possible.

compiler/lib/src/main/scala/codegen/DictionaryJsonWriter/DictionaryJsonEncoder.scala
```diff
@@ -169,17 +169,22 @@ case class DictionaryJsonEncoder(
...
-        val arrayRes = for(e <- elements) yield {
-            val res = e match {
-                // Case where array is N-dimensional
-                case Value.Array(a, t) => arrayElementsAsJson(a.elements)
-                case _ => valueAsJson(e)
-            }
-
-            res.asJson
-        }
+        val arrayRes = for(e <- elements) yield valueAsJson(e)
```
bocchino, line 172: This looks like a code improvement (eliminating a special case). Can we keep this change and preserve the existing behavior?

compiler/lib/src/main/scala/codegen/DictionaryJsonWriter/DictionaryJsonEncoder.scala
```diff
@@ -169,17 +169,22 @@ case class DictionaryJsonEncoder(
...
         arrayRes.asJson
     }
+
```
bocchino, line 175: As discussed in issue #925, after further discussion, I think we should revert this change. The current behavior is correct according to the way that FPP represents struct values with member arrays. However, @jwest115 can you make sure that the current F Prime JSON spec is consistent with the FPP model and the generated JSON here? This is a tricky case.

bocchino: Now I'm not so sure if we should revert. I think the root cause of the issue is that the dictionary JSON is underspecified here. I opened an issue on this: https://github.com/nasa/fprime/issues/4787. There are at least two ways this could work:

1. When a struct member of type _T_ has a size _n_, the value stored in the member must be a JSON array with _n_ elements, each of type _T_. When there is no size, the value must be a single value of type _T_.

2. Regardless of whether the struct member has a size, the value stored in the member can be a single value or an array. If there is no size and an array of elements, the array size must be one. If there is a size and a single element, the single element is stored in every position in the array.

Probably we want option (1), but the JSON spec does not specify what behavior is intended. We should tighten up the JSON spec first, and then make the `fpp-to-json` output conform to it.

compiler/lib/src/main/scala/codegen/DictionaryJsonWriter/DictionaryJsonEncoder.scala
bocchino: After discussion, we decided to keep the code gen change here.

compiler/lib/src/main/scala/codegen/DictionaryJsonWriter/DictionaryJsonEncoder.scala
bocchino: We decided to go with option (1).

compiler/lib/src/main/scala/codegen/DictionaryJsonWriter/DictionaryJsonEncoder.scala
```diff
@@ -280,7 +285,7 @@ case class DictionaryJsonEncoder(
                         "members" -> membersFormatted.toMap.asJson,
                     )
                     val optionalValues = Map(
-                        "default" -> default, 
+                        "default" ->  default.map(structValueAsJson),
```
bocchino: Why is this special case (struct value as default of struct type) called out here? It seems to me the general-case code for writing a struct value should work here and in all other cases, e.g., a struct as a member of a struct or array.

bocchino: I removed the special-case code.

bocchino (approved):
Looks good!

PR #933, opened 2026-02-27
Check framework constant values

### LINKED ISSUE #926: Add numeric checks for framework constants
For example, `FW_FIXED_LENGTH_STRING_SIZE` should be in range for a valid string size. Right now we're just checking that it's an integer.

- [x] Update spec
- [x] Update wiki
- [x] Implement and test semantics

bocchino:
Also fix some transposed text in the spec regarding types and type options.

Closes #926.


Kronos3:
Oops I misread the timestamp on the latest merge, this looks good!

Issue #896, opened 2025-12-11
FPP support for `deinit`


bocchino:
The framework has introduced a `deinit` function for active and queued components. The FPP team needs to understand (1) what these deinit functions are for; (2) how they are intended to be used; and (3) what modeling and code generation are required to support them.

There is a specific suggestion for unit test support here: #895. More support may be required.

thomas-bc:
Discussion:
- need deinit phase in the topology user guide/codegen 
- need deinit call in the UT codegen

PR #938, opened 2026-03-03
Add deinit call to tester and topology

LeStarch:
This adds the `deinit` calls to testers and topology.  Fixes: https://github.com/nasa/fpp/issues/896

### DIFF
--- a/.github/workflows/native-build.yml
+++ b/.github/workflows/native-build.yml
@@ -185,16 +185,20 @@ jobs:
         with:
           name: build-${{ matrix.tag }}
           path: ${{ inputs.output-directory }}
+      - name: "Download JARs"
+        uses: actions/download-artifact@v4
+        with:
+          name: build-jar
+          path:  ${{ inputs.output-directory }}
       - name: "Run Builder"
         run: |
           pip install build
+          cp -vr ${{ inputs.output-directory }}/*.jar python/fprime_fpp/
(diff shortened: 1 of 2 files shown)


LeStarch:
Fixes:
1. isQueue check should be removed, F Prime passives should have a deinit call
2. Add phasing test
3. this->component.deinit() in tester
4. Check C++ Checks should pass

bocchino:
This is also tandem:

> isQueue check should be removed, F Prime passives should have a deinit call

compiler/lib/src/main/scala/codegen/CppWriter/TopologyCppWriter/TopologyCppWriterUtils.scala
```diff
@@ -28,6 +28,10 @@ abstract class TopologyCppWriterUtils(
   def isActive(ci: ComponentInstance): Boolean =
     ci.component.aNode._2.data.kind == Ast.ComponentKind.Active
 
+  def hasQueue(ci: ComponentInstance): Boolean =
```
bocchino: Do we need this function any more?

bocchino: I removed the function -- it is unused.

PR #943, opened 2026-03-05
Refactor fpp-to-cpp tests

bocchino:
This PR does the following in `tools/fpp-to-cpp/test`:
* When framework definitions are needed, use the symbols provided in `test/fprime` instead of using ad-hoc definitions. This makes the code gen more consistent, because all framework dependencies use the same header files.
* Clean up the structure of `top` to move individual tests into subdirectories. This makes the test code more modular and scalable.
* Add more `top` tests to cover events, telemetry, data products, and typed ports. These are needed for the upcoming work on direct port calls.

### DIFF
--- a/compiler/tools/fpp-to-cpp/test/top/run.sh
+++ b/compiler/tools/fpp-to-cpp/test/top/run.sh
@@ -4,7 +4,9 @@ basic()
   run_test "-p $PWD,$FPRIME_DIR -i $FPRIME_DEPS,../phases.fpp,components.fpp -n basic.names.txt" topology && \
     diff -u basic.names.txt basic.names.ref.txt && \
     diff_cpp BasicTopology
+  status=$?
   cd $TOP_DIR
+  return $status
 }
 
 commands()
@@ -12,7 +14,9 @@ commands()
   cd $TOP_DIR/commands
   run_test "-p $PWD,$FPRIME_DIR -i $FPRIME_DEPS,../phases.fpp,components.fpp" topology && \
     diff_cpp CommandsTopology
+  status=$?
   cd $TOP_DIR
+  return $status
 }
 
 dp()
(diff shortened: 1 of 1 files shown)


compiler/tools/fpp-to-cpp/test/top/tlm_packets/NoInstances_P1TlmPacketsAc.ref.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "/Users/bocchino/JPL/Tools/fpp/compiler/tools/fpp-to-cpp/test/fprime/config/ApidEnumAc.hpp"
+#include "/Users/bocchino/JPL/Tools/fpp/compiler/tools/fpp-to-cpp/test/fprime/config/FwChanIdTypeAliasAc.hpp"
+#include "/Users/bocchino/JPL/Tools/fpp/compiler/tools/fpp-to-cpp/test/fprime/config/FwEventIdTypeAliasAc.hpp"
+#include "/Users/bocchino/JPL/Tools/fpp/compiler/tools/fpp-to-cpp/test/fprime/config/FwIdTypeAliasAc.hpp"
+#include "/Users/bocchino/JPL/Tools/fpp/compiler/tools/fpp-to-cpp/test/fprime/config/FwOpcodeTypeAliasAc.hpp"
+#include "/Users/bocchino/JPL/Tools/fpp/compiler/tools/fpp-to-cpp/test/fprime/config/FwPacketDescriptorTypeAliasAc.hpp"
+#include "/Users/bocchino/JPL/Tools/fpp/compiler/tools/fpp-to-cpp/test/fprime/config/FwSizeStoreTypeAliasAc.hpp"
+#include "/Users/bocchino/JPL/Tools/fpp/compiler/tools/fpp-to-cpp/test/fprime/config/FwTimeBaseStoreTypeAliasAc.hpp"
+#include "/Users/bocchino/JPL/Tools/fpp/compiler/tools/fpp-to-cpp/test/fprime/config/FwTimeContextStoreTypeAliasAc.hpp"
+#include "/Users/bocchino/JPL/Tools/fpp/compiler/tools/fpp-to-cpp/test/fprime/config/TimeBaseEnumAc.hpp"
```
jwest115: Should these headers be shortened to `#include config/XAc.hpp` rather than the full path?

bocchino: Fixed!

PR #951, opened 2026-03-12
Fix CI NodeJS deprecation

thomas-bc:
Upgrades the GitHub Actions versions that we're using


bocchino:
This looks better, but I still see four warnings: https://github.com/nasa/fpp/actions/runs/22980687807.

thomas-bc:
Ok @bocchino I am not seeing the warnings anymore - please confirm and we should be good to go.

PR #954, opened 2026-03-15
Add String concatenation and String Interpolation

### LINKED ISSUE #893: String concatenation and interpolation
It would be useful to be able to concatenate strings and interpolate string constants into other strings. This would help with configuring subtopologies. For example, one could provide a configuration constant
```
fooConfig="defaultFooConfig"
```
and then provide a subtopology code snippet like
```
"configureFoo(" + fooConfig + ")"
```
or
```
"configureFoo($fooConfig)"
```

abh80:
Added:
## **String concatenation**
- Adds Binop.Add(...) support to `String` type.
```fpp
constant c = "a" + "b" + "c" // gives "abc"
const a = "hello " + c  + "!" // gives "hello abc!"
```
## String interpolation
- Desugars interpolated strings into series of string concatenation, using Binop.Add implementation from above
```
constant c = "a" + "b" + "c" // gives "abc"
const a = "hello $c!" // gives "hello abc!"
```

Closes #893

### DIFF
--- a/compiler/lib/src/main/scala/syntax/Lexer.scala
+++ b/compiler/lib/src/main/scala/syntax/Lexer.scala
@@ -765,8 +765,9 @@ object Lexer {
             fetchStringLitSingle()
           case _ =>
             error("invalid usage of '$', expected identifier. If you intended to use $ as a symbol use character escape '\\$...'")
+            putChar('$')
             nextChar()
-            fetchToken()
+            fetchStringLitSingle()
         }
       case _ =>
         putChar(ch)


bocchino:
Thanks for this pull request! It is a good start.

To contribute a new feature to FPP, you will need to use this process: https://github.com/nasa/fpp/wiki/Adding-New-Features-to-FPP. If this sounds OK, we will make a feature branch, and you can make PRs to that.

The first thing we will need to do is to agree on the spec. The approach to string concatenation seems good. For interpolation, I think we will need to define an interpolation syntax that is parsed and interpreted during analysis, similar to the way that string interpolation works in Scala or Python. For instance, we probably want to allow this:
```
constant s1 = "def"
constant s2 = "abc${s1}ghi"
```

It might make sense to add concatenation separately, since it seems simpler.

bocchino:
We will also decide whether we want to allow string expressions inside an interpolation, like this:
```
constant b = "b"
constant c = "c"
constant abcd = "a${b + c}d"
```
In this case we could just write
```
constant abcd = "a$b$cd"
```
or
```
constant bc = b + c
constant abcd = "a${bc}d"
```
so maybe it's not that useful. However, for consistency, it might be good to allow it.

All of this has to be worked out in the spec before we proceed to implementation.

PR #960, opened 2026-03-16
Topology ports users guide

### LINKED ISSUE #941: Topology Ports Users Guide
Create the users guide for topology ports https://github.com/nasa/fpp/issues/743

Kronos3:
Closes #941 

This also includes:
- Additional minor fpp-check test
- `implements` keyword for vim/emacs

### DIFF
--- a/docs/users-guide/Defining-Topologies.adoc
+++ b/docs/users-guide/Defining-Topologies.adoc
@@ -1134,7 +1134,7 @@ topology B {
 
 Here the FPP analyzer checks that `b.p` is a port of instance `b`;
 if not, an error will occur.
-It is also an error to provide specify multiple ports with
+It is also an error to specify multiple ports with
 the same name in the same topology.
 
 Now in topology `A`, we can write a connection to port `B.p` instead
@@ -1158,10 +1158,8 @@ topology A {
 As usual, the FPP analyzer checks that `a.p` is an output port,
 that `B.p` is a input port, and that the types of the ports
 are compatible.
-Here `B.p` has the type of `b.p`, the actual port to which the topology port 
-`B.p` is mapped,
-and `B.p` is an input port if and only if `b.p`
-is an input port.
+Here `B.p` has the direction and type of `b.p`, the actual port to which the 
+topology port `B.p` is mapped.
 If these requirements aren't met, then an error occurs.
 
 After import resolution, the result is the same as before:
--- a/docs/fpp-users-guide.html
+++ b/docs/fpp-users-guide.html
@@ -12258,7 +12258,7 @@ and a port of a component instance, like this:</p>
 <div class="paragraph">
 <p>Here the FPP analyzer checks that <code>b.p</code> is a port of instance <code>b</code>;
 if not, an error will occur.
-It is also an error to provide specify multiple ports with
+It is also an error to specify multiple ports with
 the same name in the same topology.</p>
 </div>
 <div class="paragraph">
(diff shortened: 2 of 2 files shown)


compiler/tools/fpp-check/test/top_ports/tests.sh
```diff
@@ -8,5 +8,6 @@ internal_port
 nested
 out_to_out
 top_to_top
+unincluded_port
```
bocchino: The term "unincluded" is awkward here. Can we say "missing_port" or "port_not_member"?

compiler/tools/fpp-check/test/top_ports/unincluded_port.ref.txt
```diff
@@ -0,0 +1,5 @@
+fpp-check
+[ local path prefix ]/compiler/tools/fpp-check/test/top_ports/unincluded_port.fpp:24.3
+  port b = c2.pIn
+  ^
+error: instance c2 is not a member of topology A
```
bocchino: Should the caret be under `c2` here?

docs/users-guide/Defining-Topologies.adoc
```diff
@@ -889,25 +888,30 @@ Further, to enable modular testing, it is useful for
...
+  instance A
 
   ...
 
 }
 --------
 
+Notice that the the topology `A` is not distinguished from a component instance
+`A`. This is because both are considered "interface instances". Typically
+topologies are named in `PascalCase` while component are named in `camelCase`.
```
bocchino: We need to explain that `import` is also accepted here, for backwards compatibility.

docs/users-guide/Defining-Topologies.adoc
```diff
@@ -1037,6 +1041,87 @@ For example, if `A` imports `B` and `B` imports `C`
 and `C` imports `A`, you will get an error.
 
 
+==== Topology Ports
+
+Earlier in this section we alluded that topology definitions can
```
bocchino:
```suggestion
Earlier in this section we explained that topology definitions can
```

docs/users-guide/Defining-Topologies.adoc
```diff
@@ -889,25 +888,30 @@ Further, to enable modular testing, it is useful for
...
-  import A
+  instance A
 
   ...
 
 }
 --------
 
+Notice that the the topology `A` is not distinguished from a component instance
+`A`. This is because both are considered "interface instances". Typically
```
bocchino: Let's remove the quotation marks from "interface instances." If we've explained that term somewhere else, we should refer to the explanation. If we haven't explained it, we should do it here.

docs/users-guide/Defining-Topologies.adoc
```diff
@@ -1037,6 +1041,87 @@ For example, if `A` imports `B` and `B` imports `C`
 and `C` imports `A`, you will get an error.
 
 
+==== Topology Ports
+
+Earlier in this section we alluded that topology definitions can
+be imported into other topologies using the same mechanism that
+a component instance would be imported. This is because to the
+perspective of the topology, they can largely be treated the same
```
bocchino: "largely treated the same way" is vague. Can we say more precisely how they are and are not the same?

docs/users-guide/Defining-Topologies.adoc
```diff
@@ -1037,6 +1041,87 @@ For example, if `A` imports `B` and `B` imports `C`
...
+  connections C {
+
+    b.somePortOut -> c.bIn
+
+  }
+
+}
+--------
+
+It can be useful to use topology ports in most cases when a sub-topology
```
bocchino:
```suggestion
It can be useful to use topology ports in most cases when a subtopology
```

docs/users-guide/Defining-Topologies.adoc
```diff
@@ -1037,6 +1041,87 @@ For example, if `A` imports `B` and `B` imports `C`
...
+is imported into a super topology. Defining topology ports allows
```
bocchino:
```suggestion
is imported into a parent topology. Defining topology ports allows
```

compiler/tools/fpp-check/test/top_ports/unincluded_port.ref.txt
Kronos3: Ah good call. Moving the caret required a refactor of `Topology.portMap` since I was only tracking the location of the topology port. I cleaned up the implementation to use another `case class` to model `TopologyPort` instead of `(PortInterfaceIdentfier, Location)` which was very non-descriptive.

docs/users-guide/Defining-Topologies.adoc
Kronos3: I removed this paragraph as I now explained what an interface instance is in a section above and therefore this is redundant.

docs/users-guide/Defining-Topologies.adoc
Kronos3: I added an explanation of interface instance to this section

PR #962, opened 2026-03-17
Fix sizeof struct

jwest115:
Consider member arrays when computing the size of a struct

### DIFF
--- a/compiler/lib/src/test/scala/semantics/Types.scala
+++ b/compiler/lib/src/test/scala/semantics/Types.scala
@@ -42,7 +42,12 @@ object Types {
     String(Some(AstNode.create(Ast.ExprLiteralInt(size), sizeNodeId)))
   }
 
-  def struct(name: Ast.Ident, anonStruct: AnonStruct = AnonStruct(Map()), id: AstNode.Id = 0, sizes: Struct.Sizes = Map()): Struct = {
+  def struct(
+    name: Ast.Ident,
+    anonStruct: AnonStruct = AnonStruct(Map()),
+    id: AstNode.Id = 0,
+    sizes: Struct.Sizes = Map()
+  ): Struct = {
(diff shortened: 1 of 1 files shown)


bocchino (commented):
Great catch! There was a similar bug in the spec. I fixed it.

It occurs to me that you can add your tests from the F Prime PR here: https://github.com/nasa/fpp/blob/main/compiler/lib/src/test/scala/semantics/TypeSpec.scala. Can you do that on this PR? That way the functional tests will be in the FPP testing too.

jwest115:
> Great catch! There was a similar bug in the spec. I fixed it.
> 
> It occurs to me that you can add your tests from the F Prime PR here: https://github.com/nasa/fpp/blob/main/compiler/lib/src/test/scala/semantics/TypeSpec.scala. Can you do that on this PR? That way the functional tests will be in the FPP testing too.

Will do!

jwest115:
> > Great catch! There was a similar bug in the spec. I fixed it.
> > It occurs to me that you can add your tests from the F Prime PR here: https://github.com/nasa/fpp/blob/main/compiler/lib/src/test/scala/semantics/TypeSpec.scala. Can you do that on this PR? That way the functional tests will be in the FPP testing too.
> 
> Will do!

@bocchino After looking at this a bit more, the size of computation requires the analysis which I don't think we have in the TypeSpec test context. Am I missing something?

bocchino:
> the size of computation requires the analysis which I don't think we have in the TypeSpec test context

That's true. But you should be able to construct an Analysis object that you can use for testing. I think it just needs three entries: one in `frameworkDefinitions.typeMap`, one in `frameworkDefinitions.constantMap`, and one in `valueMap`. For the type entry, you can use a type that's already there, such as `U16`. The rest of the object can be empty.

PR #963, opened 2026-03-18
Add:  string concatenation

abh80:
Based on the discussions in #954 
Added:
## **String concatenation**
- Adds Binop.Add(...) support to `String` type.
```fpp
constant c = "a" + "b" + "c" // gives "abc"
const a = "hello " + c  + "!" // gives "hello abc!"
```

- __Concatenation Tests__:
  - [x] lib/src/test/scala/semantics/ValueSpec.scala -> add string + string add cases
  - [x] **[New]:** tools/fpp-check/test/expr/string_concat_ok.fpp -> add string concatenation ok tests
  - [x] **[New]:** tools/fpp-check/test/expr/string_concat_ok.ref.txt -> corresponding empty file
  - [ ] **[???]**:  lib/src/test/input/syntax/lexer/ok/literals.fpp -> Add a string + string example; however, the lexer will accept it regardless. I think we can skip this.
  - [x] tools/fpp-check/test/expr/tests.sh -> add test files definition for both concatenation and interpolation.
  - [x] Cover edge error cases in `fpp-check`.

- __Spec Tasks__
  - [x] Add section 10.11: String Concatenation Expressions (syntax, semantics, and examples)
  - [x] Update section 10.13: Precedence and Associativity: remember that `+` means string concatenation
  - [x] Update Arithmetic Expressions: add a cross-reference to String Concatenation for the `+` sign on strings
  - [x] Update Type-Checking: Before doing a numeric common-type computation, add a string type rule for the `+` operator.
  - [x] Update Values: keep in mind that string concatenation expressions can make string values.
  - [x] Update `defs.sh` to include `String-Concatenation-Expression.adoc`
  - [x] User guide: Add a new section 3.3.9. String Concatenation.

- __Revised Tasks__:
  - [x] Add a new function [compiler/lib/src/main/scala/analysis/CheckSemantics/CheckExprTypes.scala](https://github.com/nasa/fpp/pull/963/files/fe22ae6d884885ae381f9a3f8862afaaa50dc943#diff-d6d4f5d7c7a75cd42d7664fd83ca25d3bc572a41068fe61064e7a13c8fccf510) `convertToNumericOrString`. This will fix misleading error messages when adding to strings.
  - [x] Implement and test new semantics: implemented in `lib/src/main/scala/analysis/Semantics/Value.scala` and tested in `lib/src/test/scala/semantics/ValueSpec.scala`


compiler/lib/src/main/scala/analysis/CheckSemantics/CheckExprTypes.scala
```diff
@@ -159,7 +159,13 @@ object CheckExprTypes extends UseAnalyzer {
     for {
       a <- super.exprBinopNode(a, node, e)
       t <- a.commonType(e.e1.id, e.e2.id, loc)
-      _ <- convertToNumeric(loc, t)
+      _ <- e.op match {
+        case Ast.Binop.Add => t match {
+          case _: Type.String => Right(t)
+          case _ => convertToNumeric(loc, t)
```
bocchino: We may need a new function `convertToNumericOrString` here. Otherwise the error message could be misleading. For example, if you try to add two things that are not numbers or strings, it will suggest that you can only add numbers.

docs/spec/Expressions/String-Concatenation-Expression.adoc
```diff
@@ -0,0 +1,23 @@
+=== String Concatenation Expressions
```
bocchino: In the spec, this material needs section needs to be merged into the section on Arithmetic Expressions. I think we should rename that section Binary Expressions (that's what they are called in the implementation). We need to have one syntax for binary expressions. We can't have two expressions sections with overlapping syntax.

bocchino:
> This would mean String.binop would work for any operator

For consistency with what is already there, I think we should do the following:
* Add a member `stringOp` to `Value.Binop`.
* Implement `stringOp` for `+`.
* For the other `stringOp` cases, throw an `InternalError` (those functions should never be called).
* Implement `binop` for strings analogously to the way it is currently implemented for numbers.

docs/spec/Expressions/String-Concatenation-Expression.adoc
abh80: Agreed, I will update it meanwhile

abh80:
I have removed the string concatenation expressions section and merged it with arithmetic sections under new binary expressions section

Before:
image

After:
image

abh80:
I have implemented the changes you asked for, can you give the review @bocchino

bocchino (approved):
This is looking great! I made a few changes.

abh80:
Thank you for letting me work on this feature, @bocchino. Would you like me to also work on the interpolation section? Of course, we must first discuss the overall implementation plan.

abh80:
@bocchino I have implemented the functional testing for this feature in FppTest, kindly check and let me know if that's okay.
https://github.com/abh80/fprime/commit/ff9cf21fecbbc0b796b0d9aab154e5a9ab6cfd71

For now I will open a draft pull request in nasa/fprime

PR #968, opened 2026-03-21
Improve operator<< for structs and arrays

Kronos3:
Before this PR, `operator<<` was implemented for structs and arrays by calling out to `toString()`.

There are two problems with this approach:
1. If you define `operator<<` for an abstract type inside a struct, it will not be called.
2. If you try to implement toString() and run out of space in the Fw::StringBase buffer you are writing to, the string gets truncated
  - This is not at all optimal for unit tests where `operator<<` is enabled.

I needed this change when working on https://github.com/nasa/fprime/pull/4890 (not a tandem just useful for debugging).


bocchino:
One thing we need to be careful about: With large arrays, if used carelessly, this approach can write a large amount of data to the screen and lock up the I/O. For standard unit testing, this seems unlikely to happen, as the auto-generated code uses the `<<` operator to report events, telemetry points, and port arguments, and those usually have a small size.

lntutor (commented):
The recursive ostream approach makes sense, but there is a byte-sized integer regression to address before the tandem work.

F Prime defines I8 and U8 as int8_t and uint8_t. Direct insertion such as os << obj.m_mI8, os << obj.m_mU8, or os << obj.elements[index] therefore selects the signed-char/unsigned-char ostream overloads and emits a character byte. The current toString path uses PRIi8/PRIu8 and emits the numeric value. This affects PrimitiveU8 arrays, primitive struct fields, and fixed-size I8/U8 member arrays.

Please cast byte-sized scalar values to a wider signed or unsigned integer type in the generated ostream path, including the fixed-array branch. A small runtime FPrime test that checks the rendered text for representative I8 and U8 values would also catch this; the regenerated golden C++ references compile but do not observe stream output semantics.

bocchino:
I'm going to close this PR for now. It hasn't been worked on in some time. We can reopen it when work resumes.

PR #974, opened 2026-03-27
Refactor port serialization

bocchino:
## What's Changed

* Refactor the Scala implementation of port code gen for modularity and clarity.
    * Break the single file into a top-level file and a sub-file for each class being generated.
    * Break out large functions into smaller functions.
    * Break out complicated lambda expressions into named functions.
* Refactor the generated code for port serialization. The new version uses a `Serializer` class to serialize and deserialize the port arguments. This avoids writing out inline code at each point where serialization or deserialization is required. This is required for handling serial ports, because it's not feasible to inline the code at each connection point.

## Reviewing Notes

* You should review the submitted versions of the files in `CppWriter/PortCppWriter`. The diff of these files is not going to be helpful -- there are too many changes.

## Testing Notes

* Passes `./check-cpp`
* Passes F Prime unit tests on this branch: https://github.com/bocchino/fprime/tree/direct-ports-serial.

## Work to Go

* Use the `Serializer` class to implement the direct port calls for serial-to-typed and typed-to-serial connections. 
* Right now I just factored the port code to use the `Serializer` class; eventually we should factor the component code to use this class as well. This will cut down on generated code.


compiler/tools/fpp-to-cpp/test/component/base/NoArgsPortAc.ref.cpp
```diff
@@ -83,20 +60,24 @@ namespace Ports {
...
-    (void) _buffer;
-
 #if FW_PORT_TRACING == 1
     this->trace();
 #endif
 
     FW_ASSERT(this->m_comp != nullptr);
     FW_ASSERT(this->m_func != nullptr);
 
+    (void) _buffer;
```
LeStarch, line 79: Typically "unused" annotations like this go at the top of the function.  Not sure this is the most desirable spot.

LeStarch (approved):
I had one small comment, not something I care a ton about.

compiler/tools/fpp-to-cpp/test/component/base/NoArgsPortAc.ref.cpp
bocchino: After discussion, we decided to keep the `(void)` expression where it is. The code generation is simpler this way, because in either case (`(void)` or actual use) the generated code goes in the same place.

Kronos3 (approved):
This looks great!

PR #988, opened 2026-04-13
Add string concatenation

bocchino:
Partially addresses #893.


Kronos3 (commented):
This looks great just one comment.

Seems like we are missing a test in `fpp-to-cpp/tests/constants/constants_strings.fpp` that we should add to validate this feature end-to-end.

abh80:
@Kronos3 
For the functional testing part mentioned in https://github.com/nasa/fpp/pull/963#issuecomment-4084529872 :
> I think that may be enough functional testing on the FPP side. We could also add a functional test to FppTest in nasa/fprime.

I have implemented this in https://github.com/abh80/fprime/commit/ff9cf21fecbbc0b796b0d9aab154e5a9ab6cfd71.
Please let me know if thats enough functional testing for this feature.

bocchino:
> Seems like we are missing a test in fpp-to-cpp/tests/constants/constants_strings.fpp that we should add to validate this feature end-to-end.

I added a test for this.

Kronos3 (approved):

bocchino:
Blocked on #989.

PR #989, opened 2026-04-14
Telemetry Write Serial Interface

### LINKED ISSUE #3713: Support for bulk telemetry
| | |
|:---|:---|
|**_F´ Version_**| latest |
|**_Affected Component_**| FPP |
---
## Feature Description

FPP currently only allows specifying telemetry channels one at a time. For example:

```
        telemetry DeviceXRegisterA: U16
        telemetry DeviceXRegisterB: U16
        telemetry DeviceYRegisterA: U16
        telemetry DeviceYRegisterC: U16
```

However, there are some software systems that need hundreds or thousands of telemetry items to be produced by a component, driven by a table. While sometimes there is a way to use an array or a struct to summarize many telemetry channels into fewer channels, sometimes there is not.

It would be very useful to support components that generate _bulk_ telemetry:

1. They have a single function to call to write out an arbitrary telemetry channel, rather than hundreds of functions (one for each telemetry channel). This would be parameterized by the telemetry channel ID.
2. They would not specify the complete list of telemetry in FPP, but rather via some other mechanism. (Such as a spreadsheet or another autocoder.) This would get incorporated into the dictionary but not drive the FPP process.

## Rationale

Allows F Prime to be used in a wider range of applications without the need to manually "work around" the affordances of FPP.

Kronos3:
This PR provides a serial interface `tlmWrite()` which enables two primary use cases:

1. Writing variable channels with the same type in a loop (see https://github.com/nasa/fprime/issues/3713 for Cel's use case)
2. Integrations with other frameworks. Other frameworks may implement their own serialization and need to call into F Prime to emit telemetry. This interface avoids needing to autocode serialization (from other framework) + deserialization into F Prime types + serialization (inside `tlmWrite_*` calls.

Closes https://github.com/nasa/fprime/issues/3713


compiler/tools/fpp-to-cpp/test/component/base/PassiveTelemetryComponentAc.ref.cpp
```diff
@@ -2111,6 +2111,36 @@ F32 PassiveTelemetryComponentBase ::
...
+    FwChanIdType _id;
+
+    _id = this->getIdBase() + id;
+
+    this->m_tlmOut_OutputPort[0].invoke(
+      _id,
+      _tlmTime,
+      value
+    );
+  }
```
Kronos3: Fix this to use the new direct port interface

compiler/tools/fpp-to-cpp/test/component/base/PassiveTelemetryComponentAc.ref.cpp
```diff
@@ -2111,6 +2111,36 @@ F32 PassiveTelemetryComponentBase ::
 
 #endif
 
+// ----------------------------------------------------------------------
+// Telemetry serialized write
+// ----------------------------------------------------------------------
+
+void PassiveTelemetryComponentBase ::
+  tlmWrite(
```
Kronos3, line 2119: Use this on the higher level tlm writes

Kronos3:
@celskeggs Can you take a look at this to see if it meets your use case in https://github.com/nasa/fprime/issues/3713? I recommend looking directly at the `.ref.cpp`/`.ref.hpp` files instead of the Scala code for review.

It doesn't address your request to disable telemetry channel codegen per-channel but we believe that is not necessary since you can use this lower level function and the standard functions will simply not be included in the final binary by the linker.

compiler/tools/fpp-to-cpp/test/component/base/ActiveSerialComponentAc.ref.cpp
```diff
@@ -6771,16 +6772,8 @@ void ActiveSerialComponentBase ::
...
-    Fw::SerializeStatus _stat = _tlmBuff.serializeFrom(arg);
-    FW_ASSERT(
-      _stat == Fw::FW_SERIALIZE_OK,
-      static_cast<FwAssertArgType>(_stat)
-    );
-
     FwChanIdType _id;
-
-    _id = this->getIdBase() + CHANNELID_CHANNELU32FORMAT;
+    _id = this->getIdBase() + id;
```
celskeggs, line 6776: Many of the use cases I have are for pushing telemetry channels by absolute ID, not relative ID. Would you consider removing the `this->getIdBase() + ` term and letting callers do that only if they need it?

Kronos3: That would imply that one instance could push telemetry for another instance which is sort of a no-no. I think if that's the use case you would configure some conditional in the component implementation's logic to push/not-push based on which instance you are in.

celskeggs: Sometimes telemetry IDs are table-driven; they're configured in nonvolatile memory and the configuration is loaded at runtime. You might have hundreds of telemetry items configured in this way. Not really addressable by conditional logic.

However, I'm willing to accept that the default should be to make them relative, and a component that truly needs absolute IDs can subtract its ID base before calling this function. In the event that this becomes problematic in terms of overhead, I will bring it back up.

Kronos3: Sounds good, thanks for the input!

bocchino (approved):
Looks good! I confirmed that `check-cpp` passes.

bocchino:
@Kronos3 Can we merge this and unblock the other FPP PRs? I don't see anything here that depends on merging the F Prime PR.

bocchino:
OK, I see the F Prime PR has been merged. I'll merge this PR now.

PR #996, opened 2026-04-24
Improve Topology Analysis Performance

### LINKED ISSUE #992: Topology analysis is slow as of v3.1.1a6
I have observed that on the Mac, the time for running `fpp-check` on the F Prime Ref topology went up from under a second in v3.1.1a5 to over 3 seconds in v3.1.1a6. This indicates there may be an algorithmic issue with the changes introduced in v3.1.1a6.

Steps to reproduce:

1. Check out v4.1.1 of F Prime
2. In `Ref/Top/topology.fpp`, rename the topology from Ref to RefTopology, to work around name resolution issues
3. Set the FPP version to v3.1.1a5 in requirements.txt
4. Install F Prime
5. In Ref/Top, run
    ```
    fprime-util generate -f
    ../../fprime-venv/bin/fpp-depend ../build-fprime-automatic-native/locs.fpp *.fpp > deps.txt
    time ../../fprime-venv/bin/fpp-check `cat deps.txt` *.fpp
    ```
6. Set the FPP version to v3.1.1a6 in requirements.txt
7. Repeat steps 4 and 5 above. Compare the output.

Kronos3:
My investigation revealed `Topology.copy()` was slow which turns out is because I added a variable which computes a filtered list. Once I made the value `lazy`, performance returned to normal.

Closes #992

### DIFF
--- a/compiler/lib/src/main/resources/META-INF/native-image/reflect-config.json
+++ b/compiler/lib/src/main/resources/META-INF/native-image/reflect-config.json
@@ -23,6 +23,10 @@
   "fields":[{"name":"0bitmap$1"}]
 },
 {
+  "name":"fpp.compiler.analysis.Topology",
+  "fields":[{"name":"0bitmap$1"}]
+},
+{
   "name":"fpp.compiler.codegen.AnalysisJsonEncoder$$anon$1",
   "fields":[{"name":"0bitmap$2"}]
 },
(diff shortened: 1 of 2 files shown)


bocchino (approved):
Looks good! I confirmed that with this change, the analysis time for F Prime Ref is within 0.1 s of the time when using FPP v3.1.0. I used the JVM version for both runs.

PR #1002, opened 2026-05-05
Revise code generation for parameters

### LINKED ISSUE #984: Default External Parameters Do Not Enforce Defaults
External parameters allow default value specifications:

```
    external param SECTION_CONFIGS: SectionConfigs default TELEMETRY_SECTION_DEFAULTS
```

However, the `loadParameters` function generated does not supply the default value when it fails to load from the database:

```c++
    // Get parameter SECTION_CONFIGS
    param_valid = this->m_paramGetOut_OutputPort[0].invoke(
      _id,
      _buff
    );

    // If there was a deserialization issue, mark it invalid
    if (param_valid == Fw::ParamValid::VALID) {
      // Pass the local ID to the delegate
      constexpr FwPrmIdType _localId = PARAMID_SECTION_CONFIGS;

      FW_ASSERT(this->paramDelegatePtr != nullptr);
      // Call the delegate deserialize function for m_SECTION_CONFIGS
      _stat = this->paramDelegatePtr->deserializeParam(_baseId, _localId, param_valid, _buff);
      if (_stat != Fw::FW_SERIALIZE_OK) {
        param_valid = Fw::ParamValid::INVALID;
      }
    }
    else {
      param_valid = Fw::ParamValid::INVALID;  // <------- Here the database failed to load. However, no call to `deserializeParam` with a default value was supplied
    }
```

Since `default` is allowed with `external` it seems that the default value should be supplied to ensure that the default is set.

bocchino:
* Make the behavior of external parameters consistent with the behavior of internal parameters. In particular, respect the default values provided for external parameters.
* ~~Move the scratch parameter buffer from the stack to a component instance member, to avoid overflowing the stack.~~
* Clean up the generated code.

Closes #984.

Tandem with https://github.com/bocchino/fprime/tree/fpp-issue-984-extern-param-defaults.

## Implementation Note

In the old code for loading external parameters, if there was no valid value to send to the external parameter interface, the interface was never called. In the new code, the interface is called with an INVALID flag argument. The new behavior provides more information to the external interface. However, the external interface will need to respect the rule that if it is called with an INVALID flag argument, then the data buffer does not contain valid data.

## Review Notes

The diff of the C++ output for the component base classes may be hard to review. Here is a simple model that illustrates the new code generation strategy:
```
module Fw {
  port Time
  port PrmGet
  port PrmSet
  port Cmd
  port CmdResponse
  port CmdReg
}

passive component C {

  time get port timeGetOut

  param get port paramGet

  param set port paramSet

  command recv port cmdIn

  command resp port cmdResponseOut

  command reg port cmdRegIn

  param InternalU32Default: U32 default 42

  external param ExternalU32Default: U32 default 42

  param InternalU32NoDefault: U32

  external param ExternalU32NoDefault: U32

}
```
You can run this model through `fpp-to-cpp` to see the four cases: internal parameter with default, external parameter with default, internal parameter with no default, and external parameter with no default.


Kronos3 (approved):
I think this codegen looks good though we may want to revisit the use of the scratch buffer instead of stack buffer in the future.

compiler/tools/fpp-to-cpp/test/component/base/QueuedTestComponentAc.ref.cpp
```diff
@@ -2204,323 +2204,340 @@ void QueuedTestComponentBase ::
...
-
   _id = _baseId + PARAMID_PARAMU32;
 
-  // Get parameter ParamU32
+  // Get serialized parameter ParamU32
   this->m_param_ParamU32_valid = this->prmGetOut_out(
     0,
     _id,
-    _buff
+    this->m___fprime_ac_paramBuffer
```
LeStarch: Why was this moved from a local to a member variable?

compiler/tools/fpp-to-cpp/test/component/base/ActiveExternalParamsComponentAc.ref.cpp
```diff
@@ -4770,195 +4860,197 @@ Fw::CmdResponse ActiveExternalParamsComponentBase ::
...
+      static_cast<FwPrmIdType>(idBase),
+      PARAMID_PARAMI32EXT,
+      this->m___fprime_ac_paramBuffer
     );
-
-    return Fw::CmdResponse::OK;
   }
-
-  return Fw::CmdResponse::EXECUTION_ERROR;
+  this->m_paramLock.unlock();
```
LeStarch, line 4894: This is inconsistent with the other unlock calls which use `.unLock()`.  While F Prime supports both, one standard should be chosen for a file.

compiler/tools/fpp-to-cpp/test/component/base/ActiveExternalParamsComponentAc.ref.cpp
```diff
@@ -3472,137 +3508,149 @@ I32 ActiveExternalParamsComponentBase ::
...
+  if ((valid == Fw::ParamValid::VALID) || (valid == Fw::ParamValid::DEFAULT)) {
+    m___fprime_ac_paramBuffer.resetSer();
+    FW_ASSERT(this->paramDelegatePtr != nullptr);
+    Fw::SerializeStatus _stat = this->paramDelegatePtr->serializeParam(
+      static_cast<FwPrmIdType>(this->getIdBase()),
+      PARAMID_PARAMI32EXT,
+      m___fprime_ac_paramBuffer
+    );
+    if(_stat == Fw::FW_SERIALIZE_OK) {
+      _stat = m___fprime_ac_paramBuffer.deserializeTo(_local);
```
LeStarch: Missing `this->` standard F Prime convention.

compiler/tools/fpp-to-cpp/test/component/base/ActiveExternalParamsComponentAc.ref.cpp
```diff
@@ -3472,137 +3508,149 @@ I32 ActiveExternalParamsComponentBase ::
...
-    valid = Fw::ParamValid::INVALID;
+  this->m_paramLock.lock();
+  valid = this->m_param_ParamI32Ext_valid;
+  if ((valid == Fw::ParamValid::VALID) || (valid == Fw::ParamValid::DEFAULT)) {
+    m___fprime_ac_paramBuffer.resetSer();
+    FW_ASSERT(this->paramDelegatePtr != nullptr);
+    Fw::SerializeStatus _stat = this->paramDelegatePtr->serializeParam(
+      static_cast<FwPrmIdType>(this->getIdBase()),
+      PARAMID_PARAMI32EXT,
+      m___fprime_ac_paramBuffer
```
LeStarch: Also missing this->

compiler/tools/fpp-to-cpp/test/component/base/ActiveExternalParamsComponentAc.ref.cpp
```diff
@@ -1666,165 +1666,202 @@ void ActiveExternalParamsComponentBase ::
...
   }
   else {
-    _paramValid = Fw::ParamValid::INVALID;
+    this->m_param_ParamStringExt_valid = Fw::ParamValid::DEFAULT;
   }
+  if (this->m_param_ParamStringExt_valid == Fw::ParamValid::DEFAULT) {
+    Fw::String _val = Fw::String("external default");
+    this->m___fprime_ac_paramBuffer.resetSer();
+    _stat = this->m___fprime_ac_paramBuffer.serializeFrom(_val);
+    FW_ASSERT(this->paramDelegatePtr != nullptr);
```
LeStarch, line 1758: _stat is not checked after line 1755 before it is overwritten in 1757.  We should check or explicitly void.

compiler/tools/fpp-to-cpp/test/component/base/ActiveExternalParamsComponentAc.ref.cpp
```diff
@@ -4770,195 +4860,197 @@ Fw::CmdResponse ActiveExternalParamsComponentBase ::
...
-    const FwPrmIdType _baseId = static_cast<FwPrmIdType>(this->getIdBase());
-
+  if (!this->isConnected_prmSetOut_OutputPort(0)) {
+    return Fw::CmdResponse::EXECUTION_ERROR;
+  }
+  const FwIdType idBase = this->getIdBase();
+  Fw::SerializeStatus _stat = Fw::FW_SERIALIZE_FORMAT_ERROR;
+  // Serialize the parameter
+  this->m_paramLock.lock();
+  this->m___fprime_ac_paramBuffer.resetSer();
```
LeStarch: Use of `___` is implicitly use of `__` in an identifier and is not allowed by C++:

```
"Each identifier that contains a double underscore __ [...] is reserved to the implementation for any use."
```

compiler/tools/fpp-to-cpp/test/component/base/ActiveExternalParamsComponentAc.ref.cpp
```diff
@@ -4770,195 +4860,197 @@ Fw::CmdResponse ActiveExternalParamsComponentBase ::
...
-  return Fw::CmdResponse::EXECUTION_ERROR;
+  this->m_paramLock.unlock();
+  if (_stat != Fw::FW_SERIALIZE_OK) {
+    return Fw::CmdResponse::VALIDATION_ERROR;
+  }
+  // Save the parameter
+  this->prmSetOut_out(
+    0,
+    static_cast<FwPrmIdType>(idBase + PARAMID_PARAMI32EXT),
+    this->m___fprime_ac_paramBuffer
```
LeStarch: The unlock on line 4879 has released the parameter hold, and then the value is sent out here. Could it not be the case that some other thread is editing this local member while it is being sent out?

Perhaps we need a local copy before unlocking?

compiler/tools/fpp-to-cpp/test/component/base/ActiveExternalParamsComponentAc.ref.cpp
```diff
@@ -4628,139 +4676,181 @@ void ActiveExternalParamsComponentBase ::
...
+    _response = Fw::CmdResponse::OK;
+  }
+  else {
+    this->m_param_ParamI32Ext_valid = Fw::ParamValid::INVALID;
+    _response = Fw::CmdResponse::VALIDATION_ERROR;
   }
+  this->m_paramLock.unLock();
 
   // Call notifier
   this->parameterUpdated(PARAMID_PARAMI32EXT);
```
LeStarch: Parameter updated is called even in the error case when the user may have not updated the parameter.

compiler/tools/fpp-to-cpp/test/component/base/ActiveParamsComponentAc.ref.cpp
LeStarch, line 4548: Notice here, the param update method is **not** called on deserialization failure.  We should be consistent regardless.

LeStarch (changes requested):
Some issues found

compiler/tools/fpp-to-cpp/test/component/base/QueuedTestComponentAc.ref.cpp
bocchino: This is a change that @timcanham asked for, to avoid overflowing the stack. However, you are right that it introduces a concurrency issue as described below. For now, I think we should put the variable back on the stack.

compiler/tools/fpp-to-cpp/test/component/base/ActiveExternalParamsComponentAc.ref.cpp
bocchino: Good catch! I think we should put the paramBuffer back on the stack. If we keep it as a member and make a local copy, that defeats the purpose of having the member, which was to avoid the stack allocation.

compiler/tools/fpp-to-cpp/test/component/base/QueuedTestComponentAc.ref.cpp
bocchino: Fixed -- I moved the parameter buffer back to the stack.

compiler/tools/fpp-to-cpp/test/component/base/ActiveExternalParamsComponentAc.ref.cpp
bocchino: Fixed -- I used unlock everywhere.

compiler/tools/fpp-to-cpp/test/component/base/ActiveExternalParamsComponentAc.ref.cpp
bocchino: Fixed -- it's a local variable now.

compiler/tools/fpp-to-cpp/test/component/base/ActiveExternalParamsComponentAc.ref.cpp
bocchino: It's a local variable now.

compiler/tools/fpp-to-cpp/test/component/base/ActiveExternalParamsComponentAc.ref.cpp
bocchino: Good catch! I added in an assertion that the serialization succeeded. I think this is correct, because we are serializing into a parameter buffer. If the serialization fails, then the parameter buffer is not big enough to hold one of the default parameter values. That's a coding error.

compiler/tools/fpp-to-cpp/test/component/base/ActiveExternalParamsComponentAc.ref.cpp
bocchino: I renamed the buffer to `_paramBuffer`.

compiler/tools/fpp-to-cpp/test/component/base/ActiveExternalParamsComponentAc.ref.cpp
bocchino: Fixed.

compiler/tools/fpp-to-cpp/test/component/base/ActiveExternalParamsComponentAc.ref.cpp
bocchino: Fixed -- I guarded the notification call.

compiler/tools/fpp-to-cpp/test/component/base/ActiveParamsComponentAc.ref.cpp
bocchino: Fixed.

bocchino:
@LeStarch I re-ran Devin and it found one more issue: in the auto-generated `paramGet` functions, we should not declare `_paramBuffer` in the case of an internal parameter. This variable is unused in that case. I will fix this issue, and then the PR should be ready to go!

bocchino:
@LeStarch The issue flagged by Devin is fixed. Can you approve?

thomas-bc (approved):

PR #1004, opened 2026-05-14
FPP To JSON Avoid Whitespace

### LINKED ISSUE #993: JSON output for topologies may contain duplicate information
With the addition of topology ports, topologies refer to instances that can be component instances or topologies. When generating JSON we need to be careful to use references to instances and not to restate all the model JSON for each instance referred to. We may not be doing this, so the JSON output may be unnecessarily large.

Kronos3:
This PR changes the default behavior of `fpp-to-json` to dump without any whitespace. You can add the `--format` or `-f` flag to get back the old behavior. All the fpp-to-json tests have been changed to use the `-f` flag by default. I've added a single test that does not add the format flag to verify the standard invocation.

Closes #993


bocchino:
This looks good. Can we reverse the sense of this, so that space compression is off by default and is requested by a no spaces flag? This is more consistent with what Circe does, and it preserves backwards compatibility. If we change the default behavior, for example, that may break tests in the FPP Python model that we will need to go fix. Clients that need space compression (e.g., F Prime) can run the tool with the flag enabled. It's also more consistent with the testing structure we have. We can add a new test to test the new flag, instead of retrofitting all the old tests to test the new flag and adding a new test to test the old behavior.

Kronos3:
I disagree since any tool that reads JSON will not have any problem reading without the whitespace. The only purpose in the whitespace is to make it more human readable which is a very non-standard use-case. This change is still backwards compatible since the JSON is still valid.

In the case of the tests it makes sense to keep it on since it is a rare case in which we _do_ want whitespace.

bocchino:
OK, that makes sense. I'll agree to this change in the default behavior. With this change we will need to

* Update the behavior of the F Prime Python Model test scripts to match. See https://github.com/fprime-community/fprime-python-model/blob/main/tests/generate_ref_files.py. I think we will need to add the new `-f` flag when it invokes `fpp-to-json` to generate its reference output.

* Update the description of `fpp-to-json` on the wiki: https://github.com/nasa/fpp/wiki/fpp-to-json.

* Update s. 16.7 of the User's Guide to record the new behavior.

Can you make these changes? Thanks!

Kronos3:
Yes, I'll handle it early next week!

Kronos3:
@bocchino I've made the updates.

bocchino (approved):
Looks good! Thanks for the updates.

PR #1006, opened 2026-05-27
Template semantics

Kronos3:
This PR includes the implementation of templates. I'm opening this PR so it's easier to track work and look at diffs, it is not ready for review/merge yet.


Kronos3:
@bocchino I've dug into the `convertValueToType` conversion issue that we discussed today and I'd like to encode my findings so we can find a path forward. It's quite a long block of text, sorry!

### Test case

```
array B = [2] U32

module template M(constant b: B) {
  constant c = [1, b]
}

expand M(constant [2, 3])
```

### Error Diagnosis

The error: `cannot convert value 1 to type array of Integer`. This error occurs at the point where FPP tries to promote the primitive integer `1`  to the `commonType(1, typeof(b)`

The answer was actually encoded exactly in this error message though the message itself was misleading. See this snippet below:

https://github.com/nasa/fpp/blob/cc7fec0f5efe76829e6a3cd75fb3a0d30e590021/compiler/lib/src/main/scala/analysis/Semantics/Type.scala#L347-L350

When an anonymous array size is known, the name of the array is `[size] EltType` and when it's unknown (None) it's `array of EltType`. When the integer promoted, the array length must exist which is the reason why this conversion fails:

https://github.com/nasa/fpp/blob/cc7fec0f5efe76829e6a3cd75fb3a0d30e590021/compiler/lib/src/main/scala/analysis/Semantics/Value.scala#L65-L70

> Note: This is not a simple case of `FinalizeType` needing to be run because there is no actual type to finalize. This is an anonymous array type that needs to be finalized which is not possible since the syntax should implicitly provide a size. See below for the cause.

### Cause

As an initial note, the `[unknown] Integer` type that FPP attempts to promote `1` to is coming from the commonType computed from the `CheckExprTypes` pass earlier in the execution.

Let's read the FPP spec on commonType:

image

I believe the issue is with step 6. We are stripping away the concrete array type for an anonymous array with unknown size. This was ok before because there was never a case where this case was hit. There was never a case in which a value of concrete array type could be constructed and that value be `commonType`-ed with another type. Templates constant parameters constructing values with any concrete type.

As a side note: I think anonymous arrays of None size should never be possible at any point in the analysis. Anonymous arrays stem from array literals which should _always_ be trivially determined since you can just count the number of elements in the syntax. It may be worth keeping this in the semantics as a mechanism to allow concrete Arrays to wrap AnonArray which would allow None sizes.

A side-note to my side note: Anonymous arrays of `None` size _will_ be a valid thing when we have `[0, ...]` ellipsis operators.

### Fix

We need to change the semantics of `commonType` to say:

// (Array, Array) case

6. Otherwise if T1 and T2 are array types, throw an error (I'm fairly sure there is there no valid commonType solution here, but maybe you can think of something?)

// (Array, AnonArray)
// (AnonArray, Array)

7. Otherwise if one of T1 or T2 is an array type A equivalent to [n1] T' and the other is an anonymous array type [n2] T'':

    a. If T'' is convertible to T' and n1 == n2, let T be A
    b. Otherwise the attempted resolution is invalid. Throw an error

// (Array, promotableToArray(T))
// (promotableToArray(T), Array)

8. Otherwise if one of T1 or T2 is an array type A equivalent to [n] T' and the other is a type convertible to an anonymous array type T'':

    a. If T'' is convertible to T', let T be A
    b. Otherwise the attempted resolution is invalid. Throw an error

... continue onward with the same rules


### Implementation note!

`commonType` is computed before concrete array sizes are known, I'm not sure if these rules can actually be implemented as stated above at the moment?

I'm mainly referring to Step 7a where we check `n1 == n2`. We could theoretically defer that check later when we attempt to convert the value to the concrete type.

bocchino:
> I believe the issue is with step 6.

I think it's trying to apply step 8 (common type of Integer and [_n_] U32) at a point where _n_ is unknown. I don't think we need to revise the spec. I think we need to revise the implementation. One way to do it is to accumulate constraints at the point of type checking where the sizes aren't known, and check the constraints later after we know the sizes.

bocchino:
> Otherwise if T1 and T2 are array types, throw an error (I'm fairly sure there is there no valid commonType solution here, but maybe you can think of something?)

I think the spec is correct here. For example, if A and B are both array types with shape [2] U32, then the common type should be the shape (anonymous array). The common type should be the least type to which both types are assignable.

bocchino:
> One way to do it is to accumulate constraints at the point of type checking where the sizes aren't known, and check the constraints later after we know the sizes.

Actually I suspect what may be happening is this:

1. Before adding template parameters, we could compute partial type info (missing array sizes) and that was enough to compute values; then we use the values to finalize the types and finish the type checking.
2. With template parameters, the partial type info is not sufficient to compute the values under the current rules -- that's what seems to be causing the crash. Specifically, we are trying to convert the value 1: Integer to the partial type "array of unknown size" and the system can't do that. We never needed to do that before, but now we do.

So a solution may be to relax the value conversion rules to allow the temporary formation of array values with unknown sizes, analogously to how we temporarily form array types with unknown sizes. That should provide enough scalar value info to finalize the array values and the array types.

compiler/lib/src/main/scala/analysis/CheckSemantics/EnterTemplateSymbols.scala
```diff
@@ -0,0 +1,249 @@
...
+    aNode: Ast.Annotated[AstNode[Ast.SpecTemplateExpand]]
+  ) = {
+    val (_, node, _) = aNode
+    val data = node.data
+
+    def exprToIdentList(expr: AstNode[Ast.Expr]): Result.Result[List[AstNode[Ast.Ident]]] = {
+        expr.data match {
+          case Ast.ExprIdent(e) => Right(List(AstNode.create(e, expr.id)))
+          case Ast.ExprDot(e, id) => for (left <- exprToIdentList(e)) yield left :+ id
+          // TODO(tumbar) Make error messages specific to the parameter type
```
Kronos3: TODO!

compiler/lib/src/main/scala/analysis/CheckSemantics/EnterTemplateSymbols.scala
```diff
@@ -0,0 +1,249 @@
...
+          }
+
+          // Enter the template parameters into a new scope
+          nestedScope <- {
+            val scope = Scope.empty
+            Result.foldLeft(params) (a.nestedScope.push(scope)) ((nestedScope, param) => {
+              for {
+                nestedScope <- nestedScope.put(NameGroup.Value)(param.getUnqualifiedName, param)
+                nestedScope <- nestedScope.put(NameGroup.Type)(param.getUnqualifiedName, param)
+                nestedScope <- nestedScope.put(NameGroup.PortInterfaceInstance)(param.getUnqualifiedName, param)
```
Kronos3: Don't add to all name-groups

compiler/lib/src/main/scala/analysis/CheckSemantics/EvalConstantExprs.scala
```diff
@@ -83,11 +88,35 @@ object EvalConstantExprs extends UseAnalyzer {
...
+    }
+  }
+
   override def exprArrayNode(a: Analysis, node: AstNode[Ast.Expr], e: Ast.ExprArray) =
     for (a <- super.exprArrayNode(a, node, e))
       yield {
         val eltType = a.typeMap(node.id) match {
           case Type.AnonArray(_, eltType) => eltType
+          case Type.Array(_, Type.AnonArray(_, eltType), _, _) => eltType
           case _ => throw InternalError("element type of array expression should be AnonArray")
```
Kronos3: "or Array"

compiler/lib/src/main/scala/analysis/Semantics/Type.scala
```diff
@@ -346,7 +346,7 @@ object Type {
     override def hasNumericMembers = eltType.hasNumericMembers
     override def toString = size match {
       case Some(n) => "[" ++ n.toString ++ "] " ++ eltType.toString
-      case None => "array of " ++ eltType.toString
+      case None => "[unknown] " ++ eltType.toString
```
Kronos3: revert this back

compiler/lib/src/main/scala/util/Location.scala
```diff
@@ -6,31 +6,53 @@ import scala.util.parsing.input.Position
 final case class Location(
   file: File, /* The file */
   pos: Position, /* The position */
-  includingLoc: Option[Location] = None /* Location where this location is included */
+  /* Location where this location is included */
+  includeLoc: Option[Location] = None,
+  /* Location where this location is included */
```
Kronos3: "expanded"

compiler/lib/src/main/scala/util/Location.scala
```diff
@@ -6,31 +6,53 @@ import scala.util.parsing.input.Position
 final case class Location(
   file: File, /* The file */
   pos: Position, /* The position */
-  includingLoc: Option[Location] = None /* Location where this location is included */
+  /* Location where this location is included */
+  includeLoc: Option[Location] = None,
```
Kronos3: Make into `includingLoc` and `expandingLoc`

compiler/lib/src/main/scala/analysis/CheckSemantics/CheckExprTypes.scala
```diff
@@ -11,12 +11,15 @@ object CheckExprTypes extends UseAnalyzer {
...
       a <- symbol match {
-        // Unqualified constant symbol: visit the constant definition
+        // Constant symbol: visit the constant definition
         // to ensure it has a type
         case Symbol.Constant(node) => defConstantAnnotatedNode(a, node)
-        // Unqualified enum symbol: if this is in scope, then we are in
+        // Enum symbol: if this is in scope, then we are in
         // the enum definition, so it already has a type
         case Symbol.EnumConstant(node) => Right(a)
+        // Template parameter symbol: we are already inside the template expansion
```
bocchino:
```suggestion
        // Template constant argument symbol: we are already inside the template expansion
```

compiler/lib/src/main/scala/analysis/CheckSemantics/CheckUses.scala
```diff
@@ -156,6 +155,55 @@ object CheckUses extends BasicUseAnalyzer {
...
+    a: Analysis,
+    aNode: Ast.Annotated[AstNode[Ast.SpecTemplateExpand]]
+  ) = {
+    val node = aNode._2
+    val expansion = a.templateExpansionMap(node.id)
+    val Right(tmpl) = a.getTemplateSymbol(node.data.template.id)
+    val scope = expansion.scope
+
+    // We do use-analysis on the scope of the definition + param scope
+    // This is a bit unorthadox but we need to build a new nested scope from scratch
```
bocchino:
```suggestion
    // This is a bit unorthodox but we need to build a new nested scope from scratch
```

compiler/lib/src/main/scala/analysis/CheckSemantics/FinalizeConstantExprs.scala
```diff
@@ -0,0 +1,65 @@
...
+        Analysis.convertValueToType(v, eltType)
+      }
+      val elts = e.elts.map(f)
+      val v = Value.AnonArray(elts)
+      a.assignType(node -> Type.AnonArray(Some(e.elts.size), eltType))
+        .assignValue(node -> v)
+    }
+  }
+
+  // FIXME(tumbar) This is probably not needed
```
bocchino: Let's fix this in this PR or open an issue to track it.

compiler/lib/src/main/scala/analysis/Semantics/TemplateExpansion.scala
```diff
@@ -0,0 +1,20 @@
...
+import fpp.compiler.ast._
+import fpp.compiler.util._
+
+/** An FPP template */
+case class TemplateExpansion(
+  /** The AST node defining the template */
+  defNode: Ast.Annotated[AstNode[Ast.DefModuleTemplate]],
+  /** The AST node expanding the template */
+  expansion: Ast.Annotated[AstNode[Ast.SpecTemplateExpand]],
+  /** Concrete parameters given to this template during expansion */
```
bocchino:
```suggestion
  /** Arguments bound to the parameters of this template during expansion */
```

compiler/lib/src/main/scala/analysis/Semantics/TemplateExpansion.scala
```diff
@@ -0,0 +1,20 @@
...
+  params: Map[String, TemplateArgSymbol],
+  /** Scope where parameter symbols are entered */
```
bocchino:
```suggestion
  /** Scope where template argument symbols are entered */
```

compiler/lib/src/main/scala/analysis/CheckSemantics/EnterTemplateSymbols.scala
bocchino: Let's fix this TODO here or open an issue to track it.

compiler/lib/src/main/scala/ast/Ast.scala
```diff
@@ -428,8 +430,16 @@ object Ast {
       QualIdent.NodeList.split(nodeList) match {
         case (Nil, name) => QualIdent.Unqualified(name.data)
         case (qualifier, name) => {
+          // TODO(tumbar) Don't duplicate node ids
```
bocchino: Let's fix this TODO here or open an issue to track it.

compiler/lib/src/main/scala/analysis/CheckSemantics/CheckTemplateInterfaceParams.scala
```diff
@@ -0,0 +1,24 @@
...
+  /** Check all interface parameters template expansions implement the proper interfaces */
+  def check(a: Analysis): Result.Result[Unit] =
+    Result.foldLeft(a.templateExpansionMap.toList) (()) ((_, expansion) => {
+      val (expansionNodeId, t) = expansion
+      Result.foldLeft(t.params.values.collect {
+        case arg @ Symbol.TemplateInterfaceArg(_, _) => arg
+      })(())((_, arg) => {
+        for {
+          // Make sure this concrete value meets the specified port interface constraints
+          // These parameters are already checked if they are used in a topology inside the
```
bocchino: This comment ends in mid-sentence

compiler/lib/src/main/scala/transform/ExpandTemplates.scala
```diff
@@ -0,0 +1,1125 @@
...
+{
+
+  type State = Analysis
+
+  type In = State
+
+  type Out = Unit
+
+  def default(a: Analysis) =
+    throw new InternalError("FppExpandTemplates: Transformer not implemented")
```
bocchino:
```suggestion
    throw new InternalError("ExpandTemplates: Transformer not implemented")
```

Kronos3:
@bocchino I've worked through the outstanding review items (except one that's a minor TODO). Pending your deeper review I believe this is feature complete with all the relevant tests.

One thing to note is that `fpp-locate-defs` now expands templates (more analysis than previously needed) and fpp-syntax as an option to expand templates (off by default).

bocchino:
Thanks! I think this PR is very close. I have looked over all the code, and it looks good, and we have reviewed the algorithms together.

Here are the remaining issues that I see:

1. What is the status of the TODO? Will we fix it here, or open an issue to fix it later?

2. I think we should move the "expand templates" capability into `fpp-format`, which is soon to become `fpp-to-fpp`. It already has an "expand includes" option, and I think "expand templates" is a natural addition.

bocchino:
>  think we should move the "expand templates" capability into fpp-format

Actually maybe we should have it in both places. That's how the include expansion is now. I think template expansion in `fpp-to-fpp` is important, so users can see the source code that results from the expansion.

Kronos3:
@bocchino I'm not sure how to include this in `fpp-to-fpp` since the expanded form of a template is not technically expressible in the FPP syntax. We would need some definition that tells `check-uses` to resolve in another symbol's scope.

One option would be to just expand it even though it's not _technically_ correct.

bocchino:
> I'm not sure how to include this in fpp-to-fpp since the expanded form of a template is not technically expressible in the FPP syntax.

OK, good point. I think we will need to make it expressible in the syntax. However, we can do this in a later PR.

One idea is to allow the syntax
```
expand T {
  [members]
}
```
Its meaning is to analyze [members] as if they were expanded from T. Then `expand T` (with no body) is internally expanded  to `expand T { [members] }`, where the members come from expanding T. I think this is almost what we do now, except that the expanded syntax exists in the AST but is not expressible in the source language.

We will also need a way to communicate the expanded syntax in the Python model.

Let's discuss in our next meeting.

Kronos3:
I added the requested feature to `fpp-format` (`fpp-to-fpp`) in https://github.com/nasa/fpp/pull/1006/commits/32b449136cb3c9fa4955fa352cd9bc3d69017799 and opened an issue tracking the outstanding required update to the spec/syntax in https://github.com/nasa/fpp/issues/1090.

I also resolved the outstanding TODO. I looked into it and the code was fine as is so I removed it.

bocchino (approved):
Looks great! All my comments have been addressed.

PR #1007, opened 2026-05-27
Templates

Kronos3:


bocchino:
This branch needs to be synchronized with main before it can be reviewed and merged.

Kronos3:
The failing CI in https://github.com/nasa/fpp/actions/runs/36032184579/job/107743394485 was due to the AST Node location map not being thread-safe. By default (maybe this is a change with JDK 25?), Scala tests are run in parallel. I've added a flag to run them serially in https://github.com/nasa/fpp/pull/1007/changes/58066d9bdfbfb4d1d2645a2a10af70d61d2440d3.

bocchino:
Good catch! I think that solution is good for now, since the Scala tests are so fast anyway. We could look into making the location map thread safe, but I don't think we have any other scenario in which we expect to see concurrent access to that map.

bocchino:
We should also add this change to v3.4.0. I'll do that in my next PR.

bocchino (approved):
Looks great!

bocchino:
It's merged!!

PR #1011, opened 2026-06-04
Revise enum code gen

### LINKED ISSUE #1001: Fpp Enum Autocode Uses Undefined Behavior
EnumAc files do not fix the type of the enum in the header.  For example `DpStateEnumAc.hpp`

```cpp
30       enum T {
...
```

This yields undefined behavior in the deserialization and `isValid` code:

```cpp
 48   bool DpState ::
 49     isValid() const
 50   {
 51     return ((e >= UNTRANSMITTED) && (e <= TRANSMITTED));
 52   }
...
 67   Fw::SerializeStatus DpState ::
 68     deserializeFrom(
 69         Fw::SerialBufferBase& buffer,
 70         Fw::Endianness mode
 71     )
 72   {
 73     SerialType es;
 74     Fw::SerializeStatus status = buffer.deserializeTo(es, mode);
 75     if (status == Fw::FW_SERIALIZE_OK) {
 76       this->e = static_cast<enum T>(es);   // <------ may commit undefined behavior
 77       if (!this->isValid()) {
 78         status = Fw::FW_DESERIALIZE_FORMAT_ERROR;
 79       }
 80     }
 81     return status;
 82   }
```

Even in error cases that subsequently produce a proper error, we should not rely on undefined behavior.  

The sanitizer does catch this:
```
/Users/runner/work/fprime/fprime/build-fprime-automatic-native-ut/F-Prime/Fw/Dp/DpStateEnumAc.cpp:51:14: runtime error: load of value 255, which is not a valid value for type 'enum T'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /Users/runner/work/fprime/fprime/build-fprime-automatic-native-ut/F-Prime/Fw/Dp/DpStateEnumAc.cpp:51:14 
```

There are possible solutions:

1. Fix the width of the enum in the header to StoreType.  Thus ensuring the range is valid.  This only works when a store type is selected.
2. Check the value of store type before saving it in the value of enumeration type.

bocchino:
This PR revises the `isValid` function for enums so the implementation does not rely on undefined behavior in C++. It also incidentally (1) modernizes `typedef` to `using` in the generated code and (2) cleans up the fpp-to-json tests.

Closes #1001.

Tandem with this PR: https://github.com/nasa/fprime/pull/5295.


compiler/tools/fpp-to-cpp/test/enum/AliasSerialTypeEnumAc.ref.cpp
```diff
@@ -46,7 +46,7 @@ std::ostream& operator<<(std::ostream& os, const AliasSerialType& obj) {
 bool AliasSerialType ::
   isValid() const
 {
-  return ((e >= A) && (e <= B));
```
Kronos3, line 49: Use `this->isValid()` for consistency.

compiler/tools/fpp-to-cpp/test/enum/AliasSerialTypeEnumAc.ref.cpp
```diff
@@ -70,11 +70,11 @@ Fw::SerializeStatus AliasSerialType ::
 {
   SerialType es;
   Fw::SerializeStatus status = buffer.deserializeTo(es, mode);
+  if ((status == Fw::FW_SERIALIZE_OK) && !isValid(es)) {
```
Kronos3: this->isValid()

Kronos3 (changes requested):
Looks good, some minor nits.

compiler/tools/fpp-to-cpp/test/array/E2EnumAc.ref.cpp
```diff
@@ -46,10 +46,7 @@ std::ostream& operator<<(std::ostream& os, const E2& obj) {
 bool E2 ::
   isValid() const
 {
-  return ((e >= A) && (e <= A))
-    || ((e >= B) && (e <= B))
-    || ((e >= C) && (e <= C))
-    || ((e >= D) && (e <= D));
+  return isValid(this->e);
```
Kronos3: `this->isValid()`

compiler/tools/fpp-to-cpp/test/enum/AliasSerialTypeEnumAc.ref.cpp
Kronos3: I'm not sure what happened with this comment, use the one below

compiler/tools/fpp-to-cpp/test/enum/AliasSerialTypeEnumAc.ref.cpp
bocchino: For an instance member, I would agree. However, `isValid` is a static member. Are we sure we want to use `this->` here? It seems misleading to me, because in fact `this` has no part in the dispatch of that function.

@LeStarch Do you have an opinion here? I guess the F Prime style guide does say to use `this->` for all class members. https://github.com/nasa/fprime/wiki/F´-Style-Guidelines#class-self-references-this-

Kronos3: Oh right I didnt realize this was a static function. Though that does bring up an interesting point, maybe `ClassName::isValid()` is best since it's more clear what the function is.

bocchino: After discussion, we will update the F Prime style guide to use this pattern.

bocchino: Fixed

compiler/tools/fpp-to-cpp/test/array/E2EnumAc.ref.cpp
bocchino: Fixed

Kronos3 (approved):

LeStarch (approved):

thomas-bc (approved):

PR #995, opened 2026-04-24
Add shift operations for numeric values

abh80:
Added:
## Bit shift operations to numeric values
```fpp
constant amount = 2
constant value = 8
constant shifted = value << amount  # shifted=22
```

Closes #922 

### Implementation
- [x] `Token.scala`: added `LSHIFT` and `RSHIFT` tokens.
- [x] `Lexer.scala`: added token parsing for new tokens.
- [x] `Ast.scala`:  
   - added new trait `ShiftType` to encapsulate shift types `LShift extends ShiftType ....`.
   - added `case class Shift(op: ShiftType) extends Binop` to Binop.
- [x] `Parser.scala`: 
   - added `expr binop` node creation for new shift tokens.
   - implemented the parsing for shift tokens.
   - implement shift operand parsing with least precedence.
- [x] `analysis/CheckSemantics/CheckExprTypes.scala`: 
   - added a new helper method `convertToInteger` for shift operations (ignoring float values).
   - add the new case `Ast.Binop.Shift(_)`.
- [x]  `Value.scala`:
   - added a new type `intShiftOp = (BigInt, Int) => BigInt` for shifting.
   - added a new `private[analysis] def intShiftOp(op: Value.Binop.intShiftOp)(v: Value)` (since generic binop does not support `(BigInt, Int) => Bigint)`)
   - added two new methods to `Value`: `isNegative` and `isValidShiftAmount` to help in `Analysis.scala`.
   - implement the `intShiftOp` method in `PrimitiveInt`, `Integer` and `EnumConstant`.
   - added `<<` and `>>` methods to `Value`  to perform shift operations.
- [x] `Analysis.scala`: added `lshift` and `rshift` method, using the new methods added to `Value`.
- [x] `CheckSemantics/EvalConstantExprs.scala`: add the new case `Ast.Binop.Shift(op)` using the new methods added to `Analysis`.
- [x] `util/Error.scala`: Added a new `InvalidShiftAmount` error to `SemanticError`.

## Tests
  - [x] ‎`compiler/tools/fpp-check/test/numeric_shift`: added robust test cases including edge cases and error checks.
  - [x] `compiler/lib/src/test/input/syntax/lexer/ok/symbols.fpp`: added the new `<<`/`>>` symbols.
  - [x] should reject float-operands.
  - [x] precedence check (1 + 2 << 3 parses as and shift operations must have lowest precedence).
  - [x] should reject negative-shift values.
  - [x] `fpp-syntax` tests.
[...]

PR #1019, opened 2026-06-09
Add left and right shift operations

bocchino:
Closes #995.


compiler/lib/src/main/scala/analysis/CheckSemantics/EvalConstantExprs.scala
```diff
@@ -147,6 +147,12 @@ object EvalConstantExprs extends UseAnalyzer {
             case Ast.Binop.Div => a.div(e.e1.id, e.e2.id)
             case Ast.Binop.Mul => Right(a.mul(e.e1.id, e.e2.id))
             case Ast.Binop.Sub => Right(a.sub(e.e1.id, e.e2.id))
+            case Ast.Binop.Shift(op) =>
+              val shiftFn = op match {
+                case Ast.ShiftType.LShift => a.lshift _
+                case Ast.ShiftType.RShift => a.rshift _
+              }
+              shiftFn(e.e1.id, e.e2.id)
```
Kronos3: Why is shift a nested operand here? For consistency and simplicity we should use `Ast.Binop.LShift` and `Ast.Binop.RShift`. Is there a situation where these two need to be related?

compiler/lib/src/main/scala/syntax/Lexer.scala
```diff
@@ -548,6 +550,16 @@ object Lexer {
...
+        case '<' =>
+          nextChar()
+          if ch == '<' then
+            token = LSHIFT
+            nextChar()
+        case '>' =>
+          nextChar()
+          if ch == '>' then
+            token = RSHIFT
+            nextChar()
```
Kronos3, line 564: If there is not a second `<` or `>` we need to error. Take a look at how hex integers are handled.

Also we should add a test case for `<` and `>` actually erroring. Currently this code will just eat those characters.

Kronos3 (changes requested):
In general this looks great! There is one code style thing (feel free to ignore if you like it the way it is) as well as a bug that we definitely need to fix in the lexer.

compiler/lib/src/main/scala/analysis/CheckSemantics/EvalConstantExprs.scala
abh80: I just thought to keep it that way since, some languages such as java, swift provides `>>>` shift operators too.

Kronos3: I think if we add that it would be another binary operator. Something like `LogicalRShift`.

Kronos3 (approved):
We should be good to go now

bocchino:
@Kronos3 Thanks for the updates! Looks good.

PR #1022, opened 2026-06-15
Improve the generation of invalid serialized enums in test code

### LINKED ISSUE #1012: Provide a way to generate an invalid serialized enum without constructing an invalid enum
In the unit test framework, the only way to send a bad enum value as a command argument is to construct an enum with an invalid value and send it. But this is undefined behavior. It seems the unit testing framework should provide a way to serialize an invalid enum value without constructing an invalid enum.

One way to do this might be to provide some UT-only code in the generated enum (compiled in for UT only) that would allow you to pass in an invalid value (stored as the serial type, not the enum type) and use that for serialization.

If we did this, then we could tighten up the non-UT code so that it fails an assertion on any attempt to construct an enum with an invalid value.

Related to #1001.

bocchino:
This PR does the following:

1. Fail an assertion when an enum object is constructed with an invalid value.

2. Add a unit-test only capability to generate an invalid enum in serial form without first constructing an invalid enum object.

These changes are important because (1) construction of an invalid enum object results in undefined behavior in C++ but (2) test code needs to construct invalid serial enums (e.g., to send bad bits to an enum command argument).

Closes #1012.

Tandem with this PR: https://github.com/nasa/fprime/pull/5324.


compiler/tools/fpp-to-cpp/test/array/E1EnumAc.ref.cpp
```diff
@@ -104,12 +117,15 @@ namespace M {
     sb.format("%s (%" PRIi32 ")", s.toChar(), e);
   }
 
-#elif FW_ENABLE_TEXT_LOGGING
+#endif
+
+#ifdef BUILD_UT
 
   void E1 ::
-    toString(Fw::StringBase& sb) const
```
LeStarch, line 110: It seems that the `toString` fallback for UT builds when SERIALIZABLE_TO_STRING is disabled has been removed.  Was this intended?

LeStarch: The belief is that this is not an allowed configuration.

LeStarch (approved):

PR #1029, opened 2026-07-07
Make types consistent in handler formal parameters

### LINKED ISSUE #894: Port `handler` and Command `cmdHandler` Pass Argument Using Deifferen types.
Here is an example of an autocode port handler:

```c++
      //! Handler for input port pythonInPort
      virtual void pythonInPort_handler(
          FwIndexType portNum, //!< The port number
          const FprimePythonReference::PythonEnumeration& arg1,
          const FprimePythonReference::PythonComplexStruct& arg2
      ) = 0;

```

Now look at the cmdHandler:

```c++
      //! Handler for command TYPES_COMMAND
      //!
      //! Enumeration and struct command test
      virtual void TYPES_COMMAND_cmdHandler(
          FwOpcodeType opCode, //!< The opcode
          U32 cmdSeq, //!< The command sequence number
          FprimePythonReference::PythonEnumeration enum_argument,
          FprimePythonReference::PythonComplexStruct struct_argument
      ) = 0;
```

The same types are passed in one case by `const &` an by `value` in the other.  This is a confusing and inefficient inconsistency.

bocchino:
This PR removes an inconsistency in the code generation that was introduced in order to match the behavior of the XML autocoder.

Closes #894.

Tandem with https://github.com/nasa/fprime/pull/5372.

~Blocked on https://github.com/nasa/fpp/pull/1022, which is holding the alpha release lock.~


Kronos3 (approved):
Implementation looks good. Does this break every component command implementation with non-primitive parameters? Seems like a hard-hammer to drop.

bocchino:
> Does this break every component command implementation with non-primitive parameters?

Yes, Michael and I discussed this. We concluded that we need to make this change on the F Prime side, and the required change on the user side is very straightforward and mechanical. It is an easy fix for AI to do.

thomas-bc (approved):

PR #1034, opened 2026-07-09
Revise Common Type Specification For Concrete Arrays

Kronos3:
See https://github.com/nasa/fpp/pull/1006#issuecomment-4921285292 for why this change is needed.

One outstanding item here is if we need to revise the common type step that says:

> Otherwise if T1 or T2 or both are struct types, then replace the struct type or types with structurally equivalent anonymous struct types and reapply these rules.


Kronos3:
Closing in favor of https://github.com/nasa/fpp/pull/1006#issuecomment-4921926435

PR #1038, opened 2026-07-10
Use TreeSet and TreeMap for better performance in topology analysis

### LINKED ISSUE #1037: Use TreeMap instead of Map for better performance in topology analysis
Scala Map/Set uses deep hashing for keys. The performance is very poor when the keys have large size. Now that we have topology ports, several of the map keys can become very large (on the order of an entire resolved topology). When we put these keys into a Map, the performance can be very bad -- minutes for an F Prime use case that we discovered today.

On a branch, I've replaced Map/Set with TreeMap/Set in various places and gotten the performance for the F Prime use case down to something more reasonable (seconds).

Related to https://github.com/nasa/fprime/pull/5389.

bocchino:
Closes #1037.

There are some changes in the test output, but they all seem to have to do with map ordering that is leaking into the output. Based on inspection, the new output seems to be the same as the old output up to reordering.

### DIFF
--- a/compiler/lib/src/main/resources/META-INF/native-image/reflect-config.json
+++ b/compiler/lib/src/main/resources/META-INF/native-image/reflect-config.json
@@ -9,9 +9,6 @@
   "name":"[Lio.circe.Encoder;"
 },
 {
-  "name":"[Ljava.io.File;"
-},
-{
   "name":"[Ljava.lang.String;"
 },
 {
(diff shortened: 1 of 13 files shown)


Kronos3 (approved):
~~Have you measured a performance improvement?~~ NVM I see that's mentioned in the issue (https://github.com/nasa/fpp/issues/1037#issuecomment-4932634815).

bocchino:
Yes, the run time for the F Prime test case went from minutes to 1.2 seconds.

Issue #1033, opened 2026-07-08
Initial modeling for FSW systems


bocchino:
This issue will be the first step in multi-deployment system modeling.

## Current Design for Closing This Issue

We will add two new kinds of definitions:
1. A definition `deployment topology T { ... }` will generate C++ code for topology `T`. A definition `topology T { ... }` may be used as a subtopology but will not generate code. This solves the linking issue raised in #1026.
2. We will add a new kind of definition `system S: T`, where `S` is a name and `T` refers to a deployment topology. When the code generator sees this definition, it will use `T` to generate a system dictionary with name `S`.

## Future Design

See #1060.

## Checklist

- [x] Revise spec (#1035)
- [x] Implement and test syntax (#1036)
- [x] Update wiki: Done on the system-model branch of the wiki.
- [x] Implement and test semantics (#1039)
- [x] Implement and test code gen (#1051)
- [x] Update User's Guide (#1059)
- [x] Integrate and test with F Prime

bocchino:
@LeStarch and I discussed some of the build system implications for multi-deployment systems. Our current thinking is as follows:
1. The FPP tools will analyze the system definition and provide a list of deployments to generate.
2. The build system will run the FPP tools and generate one deployment build for each item in the list.
3. The CMake configuration will specify which deployments to build for each platform. For example, Deployment 1 could be a Linux deployment, and Deployment 2 could be a VxWorks deployment. Then the Linux build would build Deployment 1 only, and the VxWorks build would build Deployment 2 only.

bocchino:
We also noted that eventually dictionaries and packet definitions should be moved to the system level. To close this issue, we'll add a system-level dictionary that will be the same as the dictionary for the single deployment topology of the system. To preserve near-term compatibility with the GDS, we'll keep generating the topology dictionary. Eventually we should switch the GDS to using the system dictionary instead of the topology dictionary. Also, eventually in FPP we should move the packet specification to the system level.

bocchino:
Preliminary integration and test with F Prime is done. The remaining integration and test will be done as part of closing #1072.

PR #1039, opened 2026-07-14
System model semantics

bocchino:
Completes the fourth item in https://github.com/nasa/fpp/issues/1033.


compiler/lib/src/main/scala/analysis/CheckSemantics/CheckSystemDefs.scala
```diff
@@ -0,0 +1,37 @@
...
+    aNode: Ast.Annotated[AstNode[Ast.DefSystem]]
+  ) = {
+    val node = aNode._2
+    val id = node.data.topology.id
+    val ss = Symbol.System(aNode)
+    for {
+      _ <- checkDuplicateDef(a, ss)
+      t <- a.getTopology(id)
+      d <- a.getDictionary(id)
+    } yield a.copy(systemMap = Map(ss -> FppSystem(aNode, t, d)))
```
Kronos3, line 34: Isn't only a single system is valid in the model?  Don't we need to check that here

bocchino: Yes, it's checked in line 31.

Kronos3: Ah I read that as checking for another system def with the same name. I think the error/diagnostic needs a note saying only a single system is valid per-fpp model.

bocchino: OK, good idea. I'll add that.

bocchino: Done!

Kronos3 (approved):
Looks good!

PR #1044, opened 2026-07-22
Fix choice cycle detection bug

### LINKED ISSUE #1043: False positive choice cycle detected
Consider the following FPP state machine model:

```
state machine M {
    guard g1
    guard g2
    guard g3

    initial enter INIT

    signal entered
    action signalEntered

    state INIT {
        entry do { signalEntered }
        on entered enter C1

        choice C1 {
            if g1 enter C2 \
            else enter C3
        }
    }

    choice C2 {
        if g2 enter OTHER \
        else enter C3
    }

    choice C3 {
        if g3 enter FINAL \
        else enter OTHER
    }

    state FINAL
    state OTHER
}
```

FPP finds a false choice cycle:

```
fpp-check
[ local path prefix ]/compiler/tools/fpp-check/test/state_machine/transition_graph/choice_cycle_ok.fpp:26.5
    choice C3 {
    ^
error: encountered a choice cycle:
  choice transition at [ local path prefix ]/compiler/tools/fpp-check/test/state_machine/transition_graph/choice_cycle_ok.fpp:16.19 to choice C2
  choice transition at [ local path prefix ]/compiler/tools/fpp-check/test/state_machine/transition_graph/choice_cycle_ok.fpp:23.14 to choice C3
  choice transition at [ local path prefix ]/compiler/tools/fpp-check/test/state_machine/transition_graph/choice_cycle_ok.fpp:17.18 to choice C3
```

Rob and I looked at the spec which seems to specify this properly. We think this is a problem in the choice "initial node" during the choice cycle detection.

Kronos3:
Closes #1043 

TODO:
- [x] Update the trace
- [x] Update algorithm wiki (pending discussion of this solution with Rob)
   - https://github.com/nasa/fpp/wiki/Check-Transition-Graph/_compare/3969ae7f257b12fc99fc1f76f301f887937c0814
   - I decided to keep it simple and imperative.

### DIFF
--- a/compiler/lib/src/main/resources/META-INF/native-image/reflect-config.json
+++ b/compiler/lib/src/main/resources/META-INF/native-image/reflect-config.json
@@ -9,6 +9,9 @@
   "name":"[Lio.circe.Encoder;"
 },
 {
+  "name":"[Ljava.io.File;"
+},
+{
   "name":"[Ljava.lang.String;"
 },
 {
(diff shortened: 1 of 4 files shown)


bocchino (approved):
Looks good!

PR #1045, opened 2026-07-26
Add serial value constructors for generated enums

lntutor:
## Summary

Adds generated enum construction and assignment from `SerialType`. These APIs preserve the provided serial value so callers can validate it with `isValid()` before use.

Addresses nasa/fpp#1052.

## Validation

- `./fpp-sbt test` — 586 Scala tests passed
- `compiler/tools/fpp-to-cpp/test/test` — all CLI generator tests passed
- `compiler/tools/fpp-to-cpp/test/check-cpp` — generated C++ compile sweep passed

The golden outputs were regenerated with the repository-provided `update-ref` workflow.

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/EnumCppWriter.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/EnumCppWriter.scala
@@ -199,11 +199,11 @@ case class EnumCppWriter(
                 |}
                 |
                 |//! Constructor (serial representation value)
-                |//! Call isValid() before using the constructed value.
                 |$name(
                 |    const SerialType e1 //!< The serial representation value
                 |)
                 |{
+                |  FW_ASSERT(isValid(e1), static_cast<FwAssertArgType>(e1));
                 |  this->e = static_cast<enum T>(e1);
                 |}
                 |
--- a/compiler/lib/src/main/scala/codegen/CppWriter/EnumCppWriter.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/EnumCppWriter.scala
@@ -265,7 +265,8 @@ case class EnumCppWriter(
         ),
         CppDoc.Type(s"$name&"),
         lines(
-          """|this->e = static_cast<enum T>(e1);
+          """|FW_ASSERT(isValid(e1), static_cast<FwAssertArgType>(e1));
+             |this->e = static_cast<enum T>(e1);
              |#ifdef BUILD_UT
              |this->m_serializeValueIsSet = false;
              |this->m_serializeValue = 0;
--- a/compiler/lib/src/main/scala/codegen/CppWriter/EnumCppWriter.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/EnumCppWriter.scala
@@ -198,6 +198,15 @@ case class EnumCppWriter(
                 |  this->e = e1;
                 |}
                 |
+                |//! Constructor (serial representation value)
+                |//! Call isValid() before using the constructed value.
+                |$name(
+                |    const SerialType e1 //!< The serial representation value
+                |)
+                |{
+                |  this->e = static_cast<enum T>(e1);
+                |}
+                |
                 |//! Copy constructor
                 |$name(
                 |    const $name& obj //!< The source object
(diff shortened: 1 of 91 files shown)


compiler/lib/src/main/scala/codegen/CppWriter/EnumCppWriter.scala
```diff
@@ -198,6 +198,15 @@ case class EnumCppWriter(
                 |  this->e = e1;
                 |}
                 |
+                |//! Constructor (serial representation value)
+                |//! Call isValid() before using the constructed value.
```
bocchino: Let's delete this advisory comment. Instead, we should assert validity inside the body of the constructor, as we already do when the argument is an enum value.

compiler/lib/src/main/scala/codegen/CppWriter/EnumCppWriter.scala
```diff
@@ -244,6 +253,26 @@ case class EnumCppWriter(
...
+        List(
+          CppDoc.Function.Param(
+            CppDoc.Type("SerialType"),
+            "e1",
+            Some("The serial representation value"),
+          ),
+        ),
+        CppDoc.Type(s"$name&"),
+        lines(
+          """|this->e = static_cast<enum T>(e1);
```
bocchino: Assert validity here before casting the value to the enum type.

bocchino (changes requested):
Looks good! Please see comments.

lntutor:
Addressed the inline review in 24705e2: serial-representation constructors now assert `isValid(e1)` before casting, matching the raw-enum constructor, and the advisory comment was removed. Rebuilt the compiler and ran `compiler/tools/fpp-to-cpp/test/enum/run`: 9 passed.

compiler/lib/src/main/scala/codegen/CppWriter/EnumCppWriter.scala
lntutor: Removed the advisory and kept the constructor validity assertion. I also regenerated every affected golden reference in 945f29c6.

compiler/lib/src/main/scala/codegen/CppWriter/EnumCppWriter.scala
lntutor: Added FW_ASSERT validity checking before the SerialType assignment cast and regenerated all affected references in 945f29c6. The Scala and full generator suites pass.

lntutor:
@bocchino The remaining serial-assignment validation is fixed in 945f29c6e: assignment now asserts isValid(e1) before casting, and all affected constructor and assignment golden references were regenerated. Fresh verification passed 586 Scala tests, the complete fpp-to-cpp test suite, and the F Prime enum C++ compile sweep. Please re-review when convenient.

bocchino:
The code changes look good! Can we also add unit test coverage for the new code to FppTest in nasa/fprime? We can use the same procedure as for the zero-sized strings.

bocchino (approved):
The code changes here look good. I'll approve this PR, and we can add unit test coverage in nasa/fprime separately.

lntutor:
Companion FPrime coverage is now open in nasa/fprime#5537. It extends the existing typed enum suite across all eight implementations for valid `SerialType` construction/assignment and invalid-value `FW_ASSERT` behavior. The tests produce 16 expected "failed to die" failures against the original unvalidated commit `ef86088`, pass 55/55 against current head `945f29c6e`, and the complete tandem FppTest run passes 135/135 tests after compiling all 1,994 targets.

lntutor:
Attribution-only note before squash merge: the sole source commit predates my corrected Git identity and uses the unlinked address `loinguyen@local`. Please preserve the PR authorship in the squash commit, for example with `Co-authored-by: Loi Nguyen <1948922+lntutor@users.noreply.github.com>`. No code or review-state update is needed; later branches already use a GitHub-linked identity.

bocchino:
> Please preserve the PR authorship in the squash commit

We are currently using merge commits for the FPP repo, so no squashing.

lntutor:
Current head `f290d6080ea52944267655cc8e694adf96ce8a32` is now in a clean merge-ready state: the PR is mergeable, the review decision is `APPROVED`, the visible CI on that exact SHA is green, and the resolved inline feedback is already incorporated on this head. The requested FPrime unit coverage was split out separately in nasa/fprime#5537 exactly as discussed, so there does not appear to be remaining code work on this FPP PR itself. If this still matches the intended direction, it looks ready for merge when convenient.

Kronos3 (approved):
Looks great!

Issue #1032, opened 2026-07-08
Run static analysis against the code generated by fpp-to-cpp


bocchino:
AI is flagging issues related to this. See #1030, #1031.

- [ ] Add the standard F Prime static analysis to the work flow for FPP development and testing
- [ ] Fix any issues discovered

lntutor:
I would like to take this on. I’ll first wire the repository’s existing generated-C++ compile check into the standard F´ static-analysis configuration, report the concrete findings, and keep any fixes narrowly scoped to those findings.

lntutor:
I confirmed the linked findings are reported by F´’s CodeQL JPL packs, while the standard F´ framework quality job also runs clang-tidy. For the requested FPP workflow, should the initial gate run the CodeQL JPL packs against `fpp-to-cpp` generated C++, the clang-tidy quality configuration, or both? I will keep the boolean `FW_ASSERT` remediation from #1030 independently covered by the generated-output tests.

lntutor:
The first confirmed remediation is ready for review: https://github.com/nasa/fpp/pull/1046. It resolves the linked non-boolean assertion issue (#1030) and leaves CI-gate selection plus the independently scoped format-status work for the next slices.

lntutor:
@bocchino I now have a locally validated report-only CodeQL prototype. It traces the complete generated-C++ compile suite against F Prime `devel`, runs all four JPL configurations, filters F Prime source findings, and uploads SARIF artifacts. Before opening the PR, should the initial workflow run all four configurations, only the known-autocode configuration, or include clang-tidy as well?

bocchino:
Thanks for looking into this.

> Before opening the PR, should the initial workflow run all four configurations, only the known-autocode configuration, or include clang-tidy as well?

Ideally we would reproduce, as part of FPP development, whatever static analysis F Prime CI does on the generated code. That way we can have some assurance that after passing through FPP development, there will not be any new static analysis issues flagged by F Prime CI. Does this answer the question?

PR #1047, opened 2026-07-26
Explicitly discard generated event format status

### LINKED ISSUE #1031: fpp-to-cpp: generated event handlers ignore Fw::StringBase::format() return value (trips JPL Rule 14, unchecked return value)
## Background

F´ runs the CodeQL JPL Coding Standard query pack in CI (`.github/workflows/codeql-jpl-standard.yml`). One of those queries, **JPL Rule 14 – Unchecked return value** (`cpp/jpl-c/checking-return-values`), flags any call to a non-`void` function whose return value is neither checked nor explicitly cast to `(void)`.

When this query is run over a stock F´ build, it flags a large, systematic set of findings inside the **autocoder output** (`*ComponentAc.cpp`) produced by `fpp-to-cpp`.

## What

The generated text-event log handlers call `Fw::StringBase::format(...)` and discard the returned `Fw::FormatStatus`. For example, in a generated `...ComponentAc.cpp`:

```cpp
Fw::TextLogString _logString;
_logString.format(
    _formatString,
    this->m_objName.toChar(),
    "DataReceiveError ",
    stat
);

this->LogText_out(0, _id, _logTime, Fw::LogSeverity::WARNING_LO, _logString);
```

`format()` returns a real, meaningful status (`Fw/Types/format.hpp`):

```cpp
enum class FormatStatus {
    SUCCESS,
    OVERFLOWED,             // formatted string did not fit
    INVALID_FORMAT_STRING,
    SIZE_OVERFLOW,
    OTHER_ERROR
};
```

Because the return is silently dropped, the generated code both (a) hides genuine formatting errors/truncation and (b) trips JPL Rule 14 on every generated event handler.

Scale in a current build (F´ `devel`, `fprime-fpp` `3.3.0a10`): **504 occurrences across 172 generated `*ComponentAc.cpp` files.**

## Requested change

Have `fpp-to-cpp` handle the `FormatStatus` returned by the generated `format()` calls — either by checking it (e.g. `FW_ASSERT`/graceful handling) or, at minimum, by emitting an explicit `(void)` cast, which is the disposition JPL Rule 14 asks for when a return value is intentionally ignored.

## Notes

- This report is specifically about the `format()` return value. The same CodeQL query also flags generated `operator=`, `operator+=`, and `operator++` calls (which return `*this`); those are idiomatic C++ and are **not** part of this request.
- On the F´ side, the `cpp/jpl-c/checking-return-values` query is being moved to the "known auto-code issues" CodeQL run (which excludes build/autocoder paths) so it continues to guard hand-written sources while this is addressed.

_Filed on behalf of the F´ team via lestarch-autobot._

lntutor:
## Summary

Adds an explicit `(void)` cast to the `Fw::TextLogString::format()` result in generated event handlers. This documents that the return value is intentionally ignored and resolves the JPL unchecked-return-value finding without altering event logging behavior.

Fixes #1031
Part of #1032

## Validation

- `./fpp-sbt test` — 586 tests passed
- `compiler/tools/fpp-to-cpp/test/component/base/run` — 8 generated component cases passed

Golden outputs were regenerated with the repository-provided workflow.

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentEvents.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentEvents.scala
@@ -311,7 +311,7 @@ case class ComponentEvents (
             List.concat(
               lines(
                 s"""|Fw::TextLogString _logString;
-                    |_logString.format(
+                    |(void) _logString.format(
                     |  _formatString,
                     |#if FW_OBJECT_NAMES == 1
                     |  this->m_objName.toChar(),
--- a/compiler/tools/fpp-to-cpp/test/component/base/ActiveEventsComponentAc.ref.cpp
+++ b/compiler/tools/fpp-to-cpp/test/component/base/ActiveEventsComponentAc.ref.cpp
@@ -3128,7 +3128,7 @@ void ActiveEventsComponentBase ::
 #endif
 
     Fw::TextLogString _logString;
-    _logString.format(
+    (void) _logString.format(
       _formatString,
 #if FW_OBJECT_NAMES == 1
       this->m_objName.toChar(),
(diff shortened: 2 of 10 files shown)


bocchino (approved):
Looks good!

Kronos3 (approved):

PR #1051, opened 2026-07-30
System model codegen

bocchino:
This PR is part of the work on issue #1033. It adds code generation for basic system modeling. It also revises the semantics to correspond to the latest design described in the comments to the issue. The main new concept is a deployment topology that specifies whether to generate code for a topology, or whether to use it for import only. Some salient features of the design:
* Import-only topologies can't specify telemetry packets.
* Deployment topologies can't be imported. To deploy an import-only topology, one can import it into a deployment topology.
* One can now define a system consisting of a single deployment topology. This is mainly a step to the future. It generates a system-level dictionary, but for near-term backwards compatibility the tools continue to generate dictionaries for all deployment topologies.

@LeStarch Could you review the tool output in the unit tests to make sure it agrees with expectations?
@Kronos3 Could you review the Scala implementation?


lntutor (commented):
I found two testable gaps while tracing the new system definition through parsing, semantic analysis, and dictionary generation:

1. System location specifiers are parsed and emitted, but they are never checked. `CheckSpecLocs` has definition handlers through `defTopologyAnnotatedNode`, but no `defSystemAnnotatedNode` calling `checkSpecLoc(a, Ast.SpecLoc.System, Symbol.System(aNode))`. As a result, `locate system S at "wrong.fpp"` remains in `locationSpecifierMap` and is silently accepted. Could you add that handler plus `system_ok` and `system_path_error` cases to `compiler/tools/fpp-check/test/spec_loc`?

2. In `compiler/tools/fpp-to-dict/test/top/run.sh`, `dictionaryDefs()` calls `diff_json DictionaryDefs` without `&& \\` before `diff_system_json DictionaryDefs`. The function therefore returns only the system-dictionary comparison status, so a topology-dictionary mismatch can be masked when the system dictionary still matches. Please chain those checks as in the other test cases.

The rest of the deployment-topology gating and system dictionary flow looks internally consistent in this pass.

bocchino:
@lntutor Thanks, those are both good catches! I will fix.

bocchino:
@lntutor Those issues should be fixed now.

lntutor (approved):
Rechecked exact head 9c903f1565b85896855a703efaa8557402fcf095. The system location check now uses Ast.SpecLoc.System, the new system_ok/system_path_error cases pass, and dictionaryDefs now chains the topology and system JSON comparisons. Local verification: fpp-check spec_loc 33/33 passed and fpp-to-dict top 12/12 passed (JSON-schema validation dependency was unavailable locally, so those schema-only steps were skipped). The two issues from my previous review are resolved.

compiler/lib/src/main/scala/codegen/DictionaryJsonWriter/DictionaryJsonWriter.scala
```diff
@@ -22,23 +22,59 @@ object DictionaryJsonWriter extends AstStateVisitor {
...
+        val data = node.data
+        val fileName = {
+            val baseName = s.writeSymbolName(Symbol.System(aNode))
+            DictionaryJsonEncoderState.getSystemFileName(baseName)
+        }
+        val json = {
+            val Right(topSymbol) = s.a.getTopologySymbol(data.topology.id): @unchecked
+            constructJson(s, topSymbol)
+        }
+        writeJson(s, fileName, json)
```
Kronos3, line 53: > One can now define a system consisting of a single deployment topology. This is mainly a step to the future. It generates a system-level dictionary, but for near-term backwards compatibility the tools continue to generate dictionaries for all deployment topologies.

Can we leave some sort of note here stating was the long-term plan is. I _believe_ the plan is to make `deployment topology` _not_ generate dictionaries in the future and just generate them at the system level?

Kronos3 (approved):
This looks good to me. I have one minor note.

compiler/lib/src/main/scala/codegen/DictionaryJsonWriter/DictionaryJsonWriter.scala
bocchino: Yes, that will be documented in the User's Guide as part of completing #1033. This PR contains semantics changes and some minimal changes to the User's Guide needed to keep the code examples consistent.

PR #1062, opened 2026-08-10
Rename fpp-format to fpp-to-fpp

Kronos3:
I am working on integrating the repurposed `fprime-fpp-format` PIP package with `fprime-util format` though this alias is conflicting with the other `fpp-format` executable in the new PIP package.

@bocchino You mentioned updating the user's guide to point to this tool as well, would you like to do that in this PR? I think I'll need an alpha to validate the full pipeline doesn't break before we merge this.

Update August 14, 2026: This PR will now address issue #1073.


bocchino:
> You mentioned updating the user's guide to point to this tool as well, would you like to do that in this PR? 

This needs some more thought. The existing `fpp-format` tool is used for internal testing of the FPP tools and the F Prime Python model. So we can't just abandon that tool. We should consider renaming it. I'm concerned that if we make `fpp-format` mean one thing in the user-facing tools and another thing in the developer-facing tools, it will be confusing.

bocchino:
> We should consider renaming it.

@Kronos3 I believe you suggested `fpp-to-fpp` as an alternate name. I think this is reasonable. I'm also seeing a use case for this in the F Prime Python model testing, where to generate reference output we need to introduce the auto-generated syntax for state machine enums. This tool is also useful for showing what the source files look like after expanding include specifiers. And, if this isn't already done, we can extend it to expanding templates.

bocchino:
I'd like to keep this PR blocked until we rename the existing `fpp-format` tool.

bocchino:
My current thinking is that we should (1) rename `fpp-format` in this repo and in the exported tools to `fpp-to-fpp` and (2) update the User's Guide to match. Then formatting FPP won't be handled by this repo at all. It can be handled by the `fpp-tools` repo. I don't think we need to document `fpp-tools` in the FPP User's Guide. I think `fpp-tools` can have its own docs. However, we should point users to `fpp-tools` and its docs in the README of this repo and in the User's Guide. The README already points to the VS Code plugin. We could also refer to the plugin in the User's Guide. Is there any more we need to do to point users to the tools?

Kronos3:
That sounds good to me, I've renamed the PR and I'll make that update when I get a chance. I think your doc update recommendation sounds good.

bocchino:
Sounds good, I also opened an issue on this: #1073.

bocchino:
Unblocking and converting to draft.

PR #1064, opened 2026-08-11
Clarify serialization in User's Guide

### LINKED ISSUE #1063: Clarify serialization rules for arrays and member arrays
| | |
|:---|:---|
|**_F´ Version_**| 4.2.2 |
|**_Affected Component_**| FPP User's Guide |
---
## Problem Description

This is following up a misunderstanding I ran into when consulting the [serialization section of the FPP User's Guide](https://nasa.github.io/fpp/fpp-users-guide.html#Writing-C-Plus-Plus-Implementations_Serialization-of-FPP-Values), which I brought up in discussion #5622. My reading of the serialization rules is that arrays and member arrays may or may not be serialized with padding between, before, or after array elements (behavior unspecified).

The discussion answered that the intention is for serialization of FPP types to add no padding.

The core of my misunderstanding is from Rules 5 and 6:

> 5. A value of [array type](https://nasa.github.io/fpp/fpp-users-guide.html#Defining-Types_Array-Type-Definitions) is serialized as a sequence of serialized values, one for each array element, in array order. Each value is serialized using these rules.
> 6. A value of [struct type](https://nasa.github.io/fpp/fpp-users-guide.html#Defining-Types_Struct-Type-Definitions) is serialized member-by-member, in the order that the members appear in the FPP struct definition, with no padding.
>    a. Except for [member arrays](https://nasa.github.io/fpp/fpp-users-guide.html#Defining-Types_Struct-Type-Definitions_Member-Arrays), each member is serialized using these rules.
>    b. Each member array is serialized as stated in rule 5.

The exception (Rule 6a) says that member arrays do not follow these rules (which would include not following Rule 6's "with no padding"), and Rule 5 specifies only that there is a sequence in the order of the array, not whether there is padding in or around the "sequence". Thus, member arrays are excluded from any specification about padding, and array types do not have a specification about padding either.

Since the only mention of "with no padding" is in Rule 6, I was reading the exception 6a to be excusing member arrays from the no padding rule, and array types are not included in the no padding rule at the base level.

## How to Reproduce

https://nasa.github.io/fpp/fpp-users-guide.html#Writing-C-Plus-Plus-Implementations_Serialization-of-FPP-Values

## Expected Behavior

Serialization rules clearly specify that serialization of FPP types adds no padding.

Also, confirm that implementation did not fall into allowing arrays or member arrays to add padding, due to the Rule 5 ambiguity and the exception 6a allowing member arrays to avoid following Rule 6's requirement of no padding.

bocchino:
Closes #1063.

### DIFF
--- a/docs/users-guide/Writing-C-Plus-Plus-Implementations.adoc
+++ b/docs/users-guide/Writing-C-Plus-Plus-Implementations.adoc
@@ -504,17 +504,21 @@ F Prime uses the following rules for serializing data:
 is serialized into big-endian order (most significant byte first) by default,
 using the number of bytes implied by the bit width.
 For example, the `U16` value 10 (decimal) is serialized as the
-two bytes `00` `0A` (hex). If little-endian order is desired, the optional mode parameter can be specified as `Fw::Serialization::LITTLE`. This stores the data least significant byte order. The `U16` value 10 (decimal) is serialized in little-endian as the two bytes `0A` `00` (hex).
+two bytes `00` `0A` (hex). If little-endian order is desired, the optional mode 
+parameter can be specified as `Fw::Serialization::LITTLE`. This stores the data 
+least significant byte order. The `U16` value 10 (decimal) is serialized in 
+little-endian as the two bytes `0A` `00` (hex).
 
 .. A value of signed integer type (`I8`, `I16`, `I32`, or `I64`)
-is serialized by first converting the value to an unsigned value of the same bit
-width and then serializing the unsigned value as stated in rule 1.a.
+is serialized by first converting the value to an unsigned value of the same 
+bit width and then serializing the unsigned value as stated in rule 1.a.
 If the value is nonnegative, then the unsigned value is
 the same as the signed value.
 Otherwise the unsigned value is the two's complement of the signed value.
 For example:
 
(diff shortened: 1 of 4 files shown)


docs/users-guide/Writing-C-Plus-Plus-Implementations.adoc
```diff
@@ -546,19 +550,48 @@ Each string character is serialized as an `I8` value according to rule 1.b.
...
+in these rules.
+Note that the type of the value may be an
+<<Defining-Types_Array-Type-Definitions,array type>>.
+
+.. Otherwise the member is a
+<<Defining-Types_Struct-Type-Definitions_Member-Arrays,member array>>,
+which is not a single value, but an array of typed values.
+In this case, the array is serialized as specified in rule 5,
+as if it were a single value with an 
+<<Defining-Types_Array-Type-Definitions,array type>>.
```
Kronos3: I think there was some duplication here? Looks like the struct section is repeated.

Kronos3 (changes requested):

docs/users-guide/Writing-C-Plus-Plus-Implementations.adoc
bocchino: Good catch! Fixed.

Kronos3 (approved):
Looks good!

PR #1069, opened 2026-08-13
Refine handling of dependencies and implied uses needed for dictionary generation

bocchino:
This PR refines the handling of the dependencies and implied uses needed for dictionary generation. Now these dependencies and implied uses are recorded only for deployment topologies. These are the only topologies that cause dictionaries to be generated.


Kronos3 (approved):
Looks correct.

PR #1076, opened 2026-08-19
Pin published wheel version to the release tag via SETUPTOOLS_SCM_PRETEND_VERSION

LeStarch:
This PR solves the issue where publishing the same release with two tags may pick the wrong PIP version for our python packages. This solves it by supplying an explicit version keyed to the release supplied by the CI system.


bocchino (approved):
Looks good by inspection. I have not attempted to verify that it is working.

thomas-bc (approved):

PR #1078, opened 2026-08-20
Compute commonTypeOption using underlying type

Kronos3:
While working with state machines I noticed a divergence between LSP/FPP Scala (LSP was wrongly reporting an error). I then ported the spec tests into fpp-tools and found two issues in Scala FPP:

1. Common Type Option was not being computing using the underlying type (written in the spec but not implemented)
2. [Not an issue but useful to have] We are missing `TypeOptionSpec` tests


bocchino (approved):
Nice catch! The fix and new tests look good.

PR #1085, opened 2026-08-24
Vector Type Specification

### LINKED ISSUE #1080: Vector Type Spec
Update the spec to define "vector" types (variable-sized arrays)

Kronos3:
Closes #1080 

I churned a bit on the spec here initially trying to fold the original "Variable sized array" proposal that I made as a sub-case of the array type. As I got further in the implementation I realized that they are distinct types and it would be much cleaner to treat them completely separately.

This PR specifies a new "Vector type" which is _similar_ to array type but has two distinct differences:

- Vector types include both a "length" and a "maximum size".
- Values can include [0, max size] elements

Vector types also have a notion of size-prefix type which must be a type whose underlying type is an unsigned integer. This prefix type is used only during serialization to serialize the length of the vector.

### DIFF
--- a/docs/fpp-spec.html
+++ b/docs/fpp-spec.html
@@ -1831,7 +1831,7 @@ it.</p>
 <code>dictionary</code>
 <em>]</em>
 <code>array</code> <a href="#Lexical-Elements_Identifiers"><em>identifier</em></a> <code>=</code>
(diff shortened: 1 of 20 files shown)


bocchino:
To avoid cluttering the board, I would like to have at most one PR per issue that is not under active review.

Issue #665, opened 2025-03-20
Support bit fields in structs


bocchino:
Struct bit fields are useful for expressing serialized data formats that use bit packing. We would need to devise a syntax for this. The C and C++ way is to specify a type that is larger than the bit field, and then qualify the type with the smaller size, like this:
```
struct S {
  I32 a : 4 // a is a 4-bit signed field
  U32 b : 4 // b is a 4-bit unsigned field
}
```
This seems awkward. The only purpose of writing `I32` seems to be to specify the `I` part; you could have written `I8` or `I16` or `I64` there, and the meaning would be the same. It seems to me it would be more natural to directly specify the signedness and the bit width, like this:
```
struct {
  a: signed 4
  b: unsigned 4
}
```
One question is whether the bit packing would control the in-memory representation, or just the serialized form. If it controls the in-memory representation, then how do we implement the bit packing? Using C++ bit packing? Or by code-generating the bit shifts? If it doesn't control the in-memory representation, then the implementation could choose a primitive integer type to hold each bit string. For example, the struct above could be implemented with a pair of `U8` fields. That could be more efficient for some operations, but it would waste memory.

mwette:
I would make the packing more explicit.  Something like:
```
packed U32 {
  a: signed 4 offset 0
  b: unsigned 4 offset 5
} 
```
I think it will be painful to support C type structs with bitfields which allow mixed integer sizes and unspecified layout of the container.  Maybe offset could be optional.

bocchino:
Can you explain your reasoning further here? In your example, the offsets seem like information that can and should be inferred by the analyzer.

mwette:
The offsets are probably not needed.   I was being more explicit but later realized that conversion to C++ would be more cumbersome.

bocchino:
Superseded by #1092.

PR #1091, opened 2026-08-27
Add struct bit-fields specification

thomas-bc:
## Summary

Spec update for https://github.com/nasa/fpp/issues/665 (support bit fields in structs).

New struct member form, following the syntax direction suggested in the issue:

```
struct E {
  a: unsigned 3   # 3-bit unsigned field - in memory is a U8
  b: signed 9     # 9-bit signed field - in memory is a U16
  x: U32          # typed member
}
```
Key semantics:

- signed/unsigned become reserved words; bit width is an expression with 1 ≤ w ≤ 64.
- bit packing affects only the serialized form - the in memory representation is the smallest type that can contain it between 
- Consecutive bit field members form a bit field group, packed MSB-first in declaration order; group totals may be any bit count and are zero-padded to the next byte boundary (pad bits ignored on deserialization).
- Values must fit in the declared width (checked for default values).

### DIFF
--- a/docs/code-prettify/run_prettify.js
+++ b/docs/code-prettify/run_prettify.js
@@ -489,6 +489,7 @@ var IN_GLOBAL_SCOPE = false;
           "set," +
           "severity," +
           "signal," +
+          "signed," +
           "size," +
           "sizeof," +
           "stack," +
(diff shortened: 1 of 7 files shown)


bocchino:
Unfortunately there was some confusion. We don't actually want struct bit field members, we want arbitrary-width integer types as described in #1092. This gives us arbitrary-width integers everywhere, including structs, arrays, enums, and everywhere else that serialized integers can appear.

I'm going to close this PR. We can open a new one against #1092.

Issue #990, opened 2026-04-20
Add support for absolute qualified identifiers


bocchino:
Allow qualified identifiers that begin with a dot. An initial dot indicates that the qualified identifier is to be resolved from the top level, regardless of where it appears. This feature provides a way to refer to symbols that are hidden by definitions in an inner scope. It is similar to an initial `::` in a C++ name qualifier.

For example:
```
constant a = 0
module M {
  constant a = 1 # M.a hides a here
  constant b = a # Sets b to 1
  constant c = .a # Sets c to 0
  constant d = .M.a # Same as constant d = a
}
```

In this scheme, there are two kinds of qualified identifiers:
* **Relative qualified identifiers:** Each of these is a list of simple names separated by dots. It specifies a qualified name relative to the innermost scope in which the first simple name is defined. For example, when `a` is defined in scope `M`, then inside scope `M`, `a` refers to `M.a`. This is true regardless of whether `a` is defined at the top level.
* **Absolute qualified identifiers:** Each of these is a dot followed by a relative qualified identifier. It specifies a name relative to the top scope. For example, when `a` is defined at the top level, then `.a` refers to `a`. This is true even in a scope `M` that also defines `a`.

bocchino:
We will also need to add absolute identifier expressions (i.e., identifier expressions that begin with a dot).

PR #1094, opened 2026-09-01
Specification for absolute qualified identifiers

bocchino:
This PR completes the first part of #990.


sylvesterkaczmarek (commented):
The new identifier-expression production is internally inconsistent: it says `qualified identifier . I`, which requires a left-hand qualified identifier, yet the example `.a` has no such `Q` and is not generated by that wording. Define the alternative explicitly as an **absolute qualified identifier** (or add `. I` as the base production) and cover both `.a` and `.M.a` so the grammar and examples agree.

bocchino:
> The new identifier-expression production is internally inconsistent

That language is not a production. It specifies a qualified identifier of the form `.`_I_, which is one of the forms for a qualified identifier specified in § 15.1. I added the phrase "of the form" to clarify the intended meaning.

LeStarch (approved):

sylvesterkaczmarek (approved):
Rechecked current `3727150`. The spec now defines absolute qualified identifiers explicitly, includes single- and multi-part examples, and the identifier-expression wording is consistent with that definition. My earlier wording concern is resolved.

PR #1095, opened 2026-09-01
Syntax for absolute qualified identifiers

bocchino:
This PR represents the second step of #990.


sylvesterkaczmarek (commented):
One integration concern with landing the syntax phase independently: the semantic visitors currently discard `isAbsolute` (`Unqualified(name, _)`, `ExprIdent(id, _)`), so `.a` is accepted but can silently resolve exactly like relative `a`. Could absolute nodes be rejected until the semantics step lands, or should syntax and resolution merge atomically to avoid wrong-symbol binding?

bocchino:
This is a PR into a PR into a feature branch. The feature branch won't be merged into main until everything is implemented and tested.

bocchino:
In working on the semantics, I realized that in the AST, we should store the `isAbsolute` information only in the first unqualified name of the qualified identifier, not at every position. Update coming soon.

LeStarch (approved):
In as much as I am capable of reviewing scala, this looks reasonable.  No obvious issues.

sylvesterkaczmarek (approved):
Rechecked current `12d65b3`. The absolute marker is retained in the AST and writers, and this syntax change targets the feature stack rather than `main`, so it cannot land independently of the later semantics work. My earlier integration concern is resolved.

PR #1097, opened 2026-09-03
Generate parameterUpdated calls in parametersLoaded

durasi:
Implements nasa/fpp#1098.

Per @Willmac16's guidance on the issue, the calls go in the parametersLoaded hook rather than the loadParameters body.

The generated parametersLoaded base implementation was empty, so a component that centralizes update handling in parameterUpdated did not get called for each parameter at load time and had to iterate manually. It now calls parameterUpdated once per parameter, giving full coverage on load with no manual iteration.

This is a behavior change: parameterUpdated now fires for every parameter when parameters are loaded. Because parametersLoaded stays a virtual hook, a component can override it to restore the previous no-op behavior, for example to avoid double-handling when updates are managed elsewhere.

Updated the affected component and tester base reference outputs.

Testing: Scala unit tests pass (633). Component command-line tests pass, reference outputs regenerated.

### DIFF
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriter.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriter.scala
@@ -144,12 +144,13 @@ case class ComponentCppWriter (
   }
 
   private def getCppIncludes: CppDoc.Member = {
-    val userHeaders = List(
+    val prmValidHeader = if hasParameters then List("Fw/Prm/ParamValid.hpp") else Nil
+    val userHeaders = (List(
       "Fw/Types/Assert.hpp",
       "Fw/Types/ExternalString.hpp",
       "Fw/Types/String.hpp",
       s.getIncludePath(componentSymbol, fileName)
-    ).sorted.map(CppWriter.headerString).flatMap({
+    ) ++ prmValidHeader).sorted.map(CppWriter.headerString).flatMap({
       case s: "#include \"Fw/Types/String.hpp\"" =>
         lines(
           s"""|#if FW_ENABLE_TEXT_LOGGING
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriter.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriter.scala
@@ -144,13 +144,15 @@ case class ComponentCppWriter (
   }
 
   private def getCppIncludes: CppDoc.Member = {
-    val prmValidHeader = if hasParameters then List("Fw/Prm/ParamValid.hpp") else Nil
-    val userHeaders = (List(
-      "Fw/Types/Assert.hpp",
-      "Fw/Types/ExternalString.hpp",
-      "Fw/Types/String.hpp",
-      s.getIncludePath(componentSymbol, fileName)
-    ) ++ prmValidHeader).sorted.map(CppWriter.headerString).flatMap({
+    val userHeaders = List.concat(
+      guardedList (hasParameters) (List("Fw/Prm/ParamValid.hpp")),
+      List(
+        "Fw/Types/Assert.hpp",
+        "Fw/Types/ExternalString.hpp",
+        "Fw/Types/String.hpp",
+        s.getIncludePath(componentSymbol, fileName)
+      )
+    ).sorted.map(CppWriter.headerString).flatMap({
       case s: "#include \"Fw/Types/String.hpp\"" =>
         lines(
           s"""|#if FW_ENABLE_TEXT_LOGGING
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentParameters.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentParameters.scala
@@ -225,13 +225,18 @@ case class ComponentParameters (
     Some(
       s"""|\\brief Called whenever parameters are loaded
           |
-          |This function does nothing by default. You may override it.
+          |By default this notifies the component of each parameter via
+          |parameterUpdated. You may override it, for example to suppress
+          |notification on load.
           |"""
     ),
     "parametersLoaded",
     Nil,
     CppDoc.Type("void"),
-    lines("// Do nothing by default"),
+    if sortedParams.isEmpty then lines("// No parameters")
+    else sortedParams.flatMap((_, param) =>
+      lines(s"this->parameterUpdated(${paramIdConstantName(param.getName)});")
+    ),
     CppDoc.Function.Virtual
   )
 
--- a/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentParameters.scala
+++ b/compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentParameters.scala
@@ -131,7 +131,7 @@ case class ComponentParameters (
       List(
         getParamUpdateHookFunction,
         getParamLoadHookFunction,
-        getParamLoadedHookFunction
+        getParamsLoadHookFunction
       )
     )
   }
@@ -238,7 +238,7 @@ case class ComponentParameters (
     CppDoc.Function.Virtual
   )
 
-  private def getParamLoadedHookFunction = functionClassMember(
+  private def getParamsLoadHookFunction = functionClassMember(
     Some(
       s"""|\\brief Called for each parameter when parameters are loaded
           |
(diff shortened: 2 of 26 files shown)


compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentParameters.scala
```diff
@@ -225,13 +225,18 @@ case class ComponentParameters (
...
+          |By default this notifies the component of each parameter via
+          |parameterUpdated. You may override it, for example to suppress
+          |notification on load.
           |"""
     ),
     "parametersLoaded",
     Nil,
     CppDoc.Type("void"),
-    lines("// Do nothing by default"),
+    if sortedParams.isEmpty then lines("// No parameters")
```
bocchino: Let's remove the if-else condition. The hook function won't be generated unless there are parameters, so the if case is dead code.

compiler/tools/fpp-to-cpp/test/component/base/ActiveExternalParamsComponentAc.ref.cpp
```diff
@@ -3492,7 +3492,12 @@ void ActiveExternalParamsComponentBase ::
 void ActiveExternalParamsComponentBase ::
   parametersLoaded()
 {
-  // Do nothing by default
+  this->parameterUpdated(PARAMID_PARAMI32EXT);
```
bocchino: I think we should guard each call to `parameterUpdated`  for parameter P with `if (this->m_param_P_valid == true) { ... }`. Otherwise the situation seems dangerous. The usual practice is to assert on parameter validity when getting a parameter. If the parameter fails to load and you call parameterUpdated, and parameterUpdated assumes the parameter is valid, then FSW will crash. Also, if `param_valid` is false, then the parameter has not been updated (though an update attempt has been made).

We should also consider adding add a `parameterLoaded` hook function and calling that here by default. By default, `parameterLoaded(P)` could call `parameterUpdated(P)` if `m_param_P_valid == true`. This would provide more flexibility and precision. I'm a concerned about treating the loaded case in exactly the same way as the updated case -- there is information loss. When `parameterUpdated` is called, you don't know if the parameter was loaded from the database or updated by command.

durasi:
I pushed in 1802ae8a7.

Each parameterUpdated call is now wrapped in a VALID check, reusing the existing checkValidityFlag helper, so notification only fires for parameters that loaded successfully:

if (this->m_param_ParamU32_valid == Fw::ParamValid::VALID) {
  this->parameterUpdated(PARAMID_PARAMU32);
}

I also removed the dead emptyparams branch you flagged, since the hook is only generated when the component has parameters. Reference outputs regenerated; component tests pass.

On the parameterLoaded hook idea: I think that's a good addition separating "loaded" from "updated" avoids the information loss you mentioned. I'd suggest keeping this PR to the validity guarded behavior and opening a separate issue for the new hook, so this one stays focused. Happy to take that on as a followup if you agree.

bocchino:
Thanks!

> I'd suggest keeping this PR to the validity guarded behavior and opening a separate issue for the new hook, so this one stays focused.

I discussed this with @LeStarch. We would like to add the `parameterLoaded` function now, since it is a small change and it is closely related to this PR. Can you go ahead and add that? It should take the parameter ID and the validity status as arguments. By default it should call `parameterUpdated` with the ID if `FW_PARAM_OK` is true for the validity status.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentParameters.scala
```diff
@@ -225,13 +225,20 @@ case class ComponentParameters (
...
+          |notification on load.
           |"""
     ),
     "parametersLoaded",
     Nil,
     CppDoc.Type("void"),
-    lines("// Do nothing by default"),
+    sortedParams.flatMap((_, param) =>
+      wrapInIf(
+        checkValidityFlag(param, "VALID"),
```
bocchino: We need to accept the default case as well. Let's replace this with the newly available macro `FW_PARAM_OK(param)` (see `Fw/ParamValid.hpp` in the F Prime repo). That handles both cases (VALID and DEFAULT).

bocchino:
Also, let's call the `parameterLoaded` function as each parameter is loaded, and not all in a block in `parametersLoaded`. Let's revert `parametersLoaded` to doing nothing by default.

durasi:
I pushed in ed7513c15

Replaced the manual VALID check with FW_PARAM_OK(valid), so both VALID and DEFAULT are accepted. Added the Fw/Prm/ParamValid.hpp include to the generated .cpp so the macro is defined.
Added the parameterLoaded(FwPrmIdType id, Fw::ParamValid valid) hook. By default it calls parameterUpdated(id) when FW_PARAM_OK(valid) holds.
It's now called from the load path as each parameter is loaded, rather than in a block. parametersLoaded() is back to doing nothing by default.

Reference outputs regenerated; fpp-to-cpp tests and unit tests pass.

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentCppWriter.scala
```diff
@@ -144,12 +144,13 @@ case class ComponentCppWriter (
...
   private def getCppIncludes: CppDoc.Member = {
-    val userHeaders = List(
+    val prmValidHeader = if hasParameters then List("Fw/Prm/ParamValid.hpp") else Nil
+    val userHeaders = (List(
       "Fw/Types/Assert.hpp",
       "Fw/Types/ExternalString.hpp",
       "Fw/Types/String.hpp",
       s.getIncludePath(componentSymbol, fileName)
-    ).sorted.map(CppWriter.headerString).flatMap({
+    ) ++ prmValidHeader).sorted.map(CppWriter.headerString).flatMap({
```
bocchino:
```suggestion
    val userHeaders = List.concat(
      guardedList (hasParameters) (List("Fw/Prm/ParamValid.hpp")),
      List(
        "Fw/Types/Assert.hpp",
        "Fw/Types/ExternalString.hpp",
        "Fw/Types/String.hpp",
        s.getIncludePath(componentSymbol, fileName)
      )
    ).sorted.map(CppWriter.headerString).flatMap({
```

compiler/lib/src/main/scala/codegen/CppWriter/ComponentCppWriter/ComponentParameters.scala
```diff
@@ -130,7 +130,8 @@ case class ComponentParameters (
       "Parameter hook functions",
       List(
         getParamUpdateHookFunction,
-        getParamLoadHookFunction
+        getParamLoadHookFunction,
+        getParamLoadedHookFunction
```
bocchino: For consistency, let's rename `getParamLoadedHookFunction` to `getParamsLoadHookFunction`.

bocchino:
@durasi This is looking good. I made a couple more coding suggestions. Can you confirm that the unit tests pass in F Prime when you use this build of FPP? You can use the procedure here: https://github.com/nasa/fpp/wiki/Adding-New-Features-to-FPP#integration-with-f-prime. If this change causes anything to break in F Prime, then we will need to make a tandem PR to F Prime.

durasi:
Ran the F Prime integration test with this build of FPP (769faa3c6), following the wiki procedure. Injected the local FPP build into a fresh nasa/fprime devel checkout and ran the FppTest unit tests.

Result: 18 of 19 pass, including the parameter component tests (passive/active/queued). One fails:

FppTest_topology_special_ports_ut_exe — Subprocess aborted
Assert: FppTest/topology/components/Comp/Comp.cpp:35

Root cause this is the expected consequence of the new behavior, and needs a tandem F Prime change:

Comp has param Param: U32 default 0. The special-ports topology calls loadParameters() without the parameter being set, so it loads as DEFAULT.
With this PR, parameterLoaded now calls parameterUpdated whenever FW_PARAM_OK(valid) holds i.e. for DEFAULT as well as VALID.
Comp::parameterUpdated calls getParameter(), which asserts valid == Fw::ParamValid::VALID (Comp.cpp:35). With a DEFAULT parameter this assert now fires.

I confirmed this is caused by the change: with released FPP v3.3.0 the same test passes (19/19); only with this build does special_ports fail.

The fix belongs in F Prime's Comp test component, and how to adjust it is a design choice I'd rather leave to you. A few options:

A: relax the assert in getParameter to FW_ASSERT(FW_PARAM_OK(valid)), accepting VALID and DEFAULT.
B: leave getParameter as is and have Comp::parameterUpdated skip getParameter when the parameter is DEFAULT.
Or whatever approach you prefer for how this test component should behave under the new hook.

Once you let me know which direction you'd like, I'll prepare the tandem F Prime PR

bocchino:
Thanks, I don't think this requires a tandem PR with an FPP alpha release. Can you make a PR to F Prime with option A (accept VALID and DEFAULT)? You can reference this PR in that PR. Then CI should still pass in F Prime, and once that is merged, we can merge the change here.

bocchino:
Blocked on the required change to F Prime.

durasi:
Opened the tandem F Prime PR: nasa/fprime#5877
it updates the Comp test component's getParameter to accept VALID and DEFAULT via FW_PARAM_OK (option A). 
I verified it passes FppTest (19/19) both with released FPP v3.3.0 and with this PR's build, 
so F Prime CI should stay green. 
Once it's merged, this PR can follow

bocchino (approved):
Looks good!

LeStarch (approved):

Issue #813, opened 2025-09-17
Templates


Kronos3:
FPP Templates will add the ability to create parameterizable pieces of FPP syntax that can be expanded.

PR #1102, opened 2026-09-09
Fix templates testing gaps, codegen, and spec compliance

Kronos3:
I was reviewing #813 using AI and Claude found some gaps in the templates testing (and therefore surfaced a couple of bugs) as well as non-compliance with the spec.

This PR includes the following changes:

1. Consistency/typo fixes in the spec
2. `interface [ident]` -> `instance [ident]` in the parser for template param/args.
3. Don't push all template parameters to the same name group (use the proper name group for each param)
4. Downstream code-gen needs an _expanded_ AST. This one is the main interesting change see my note below.
5. Explicitly dis-allow shadowing for template symbols in the expanded scope. This was something we described in the spec but I had wrongly assumed that the symbol entry would catch this. (See `EnterTemplateSymbols.scala` changes).
6. A bunch of tests to cover the full end-to-end surface of the templates feature

### (4) Changes to CheckSemantics/fpp tool entrypoints

This is the main note-worthy change to FPP made in this PR.

`CheckSemantics.tuList` now returns a `(Analysis, List[Ast.TransUnit])` instead of simply `Analysis`. This is because to expand templates we need to perform some analysis (i.e. enter symbols) and the template expansion will modify the translation units. I've moved `tul <- AddStateEnums.transUnitList(tul)` back into `CheckSemantics`.

Another option we could consider is to split these steps into their own separate function and then have CheckSemantics be `(Analysis, List[Ast.TransUnit]) -> Analysis:

```
...
      tul <- AddStateEnums.transUnitList(tul)
      a <- EnterSymbols.visitList(a, tul, EnterSymbols.transUnit)
      a_tul <- ResolveTemplates.transUnit(a, tul)
      a <- Right(a_tul._1)
      tul <- Right(a_tul._2)
...
```


compiler/lib/src/main/scala/util/Error.scala
```diff
@@ -757,20 +770,26 @@ object SemanticError {
     p2Number: Int,
     matchingLoc: Location
   ) extends Error
-  /** Mismatched template parameter */
+  /** Wrong number of template arguments */
   final case class MismatchedTemplateParameters(
```
bocchino: Let's rename the case class here to match the comment.

compiler/lib/src/main/scala/util/Error.scala
```diff
@@ -757,20 +770,26 @@ object SemanticError {
     p2Number: Int,
     matchingLoc: Location
   ) extends Error
-  /** Mismatched template parameter */
-  final case class MismatchedTemplateParameters(
+  /** Wrong number of template arguments */
+  final case class WrongNumberOfTemplateParameters(
```
bocchino:
```suggestion
  final case class WrongNumberOfTemplateArguments(
```

The name should match the comment. The comment looks right.

bocchino:
Looks good! I made a couple of comments, and we discussed a refactoring. Otherwise I think this is ready to merge!

Kronos3:
@bocchino I did the refactor we discussed last night. I'm going to do one more pass over these changes and then we can merge.

bocchino (approved):
Looks good!

PR #1106, opened 2026-09-10
Update to Java 25 and Scala 3.9.0

### LINKED ISSUE #1103: Update to Scala 3.9.0 and Java 25
This will update to the latest versions of Java and Scala that are available.

bocchino:
This PR brings Java and Scala up to the latest available releases. It continues to use GraalVM Community Edition for the native builds.

## Notes

1. Java has tightened some requirements for using its API, and Scala is still not fully compliant. So some warning-suppression flags are required in the scripts that run the tools.
2. GraalVM Community Edition has dropped support for Intel Mac, so on that platform this configuration will fall back to using native tools. If we want to support native tools on Intel Mac, we will need another solution.
3. The changes to the Scala code are standard and are required by the migration from Scala 3.1 to Scala >= 3.8.
4. There are some changes in the output of `fpp-to-json`. On inspection, they appear to be ordering changes, likely due to an unspecified ordering in the output and a change in behavior with the updated JVM. I manually tested this FPP version against the F Prime Ref integration tests, and I see the same behavior as with fpp/main (the DpCatalog tests pass in CI and fail locally on my Mac, but this behavior seems unrelated any changes here).
5. I had Claude add a test-wheels script here. Claude pointed out that the old script in the actions repo was out-of-date because FPP has since been factored into a single binary. This step removes all dependencies on the actions repo. Later we can pull the scripting here into the actions repo or somewhere else, if desired. For now I will focus on getting the scripting to work here -- not a trivial task in itself.

Closes #1103.


python/fprime_fpp/__main__.py
```diff
@@ -1,9 +1,39 @@
...
 
 from pathlib import Path
 
+# The JAR file is compiled for this version of Java
+MINIMUM_JAVA_VERSION = 25
+
+# Flags that suppress JVM warnings. These require JDK 23 or later.
+JAVA_WARNING_FLAGS = [
+    "--sun-misc-unsafe-memory-access=allow",
+    "--enable-native-access=ALL-UNNAMED",
```
Kronos3, line 14: Did you look into why these are needed? Is it Scala in general, one of our Scala dependencies, or something we are doing?

compiler/build.sbt
Kronos3, line 25: I find it very odd that none of these need a version bump. Bumping the versions might fix the warning flags required.

python/fprime_fpp/__main__.py
bocchino: From the warning output, it looks like the warning is caused by the Scala implementation itself, for example, in the way that Scala implements lazy variables.

Kronos3 (approved):
I think these updates are reasonable. My only nit is to look into upgrading the dependencies to their latest. I think it's a good opportunity to do it.

I think updating will also fix some of those warnings. For example, see https://github.com/scala/scala-parser-combinators/releases/tag/v2.5.0 which is the latest version of the parser combinator we are using specifically calls out fixing these warnings for JDK 24+.

bocchino:
> My only nit is to look into upgrading the dependencies to their latest. I think it's a good opportunity to do it.

Agreed. I'll upgrade all those dependencies to the latest except for circe; upgrading that one causes a compilation error that I don't want to deal with right now. I confirmed that updating everything (including circe) does **not** fix the warnings.

compiler/build.sbt
bocchino: See below.

LeStarch (approved):

PR #1110, opened 2026-09-14
Update Python wrapper to respect `JAVA_HOME`

### LINKED ISSUE #1107: Python wrapper should honour JAVA_HOME before falling back to java on PATH
## Summary

The Python wrapper (`python/fprime_fpp/__main__.py`) locates the JVM for the JAR fallback with `shutil.which("java")` only. It should honour `JAVA_HOME` first (`$JAVA_HOME/bin/java`, or `java.exe` on Windows) and only then fall back to `java` on `PATH`.

## Motivation

`JAVA_HOME` is the conventional way to select a specific JVM, and the other Java-based F´ tool (YAMCS, via its `yamcsd` script) already prefers `$JAVA_HOME/bin/java` when set. Today a user who has installed a Java 25 and pointed `JAVA_HOME` at it, but whose `PATH` still resolves an older `java`, gets the wrapper's "requires Java 25 or later" error even though they have configured a suitable JVM.

This also matters for a planned `fprime-java` pip package that provides a `java` console script (a jlink-trimmed Temurin 25 runtime) so that users without Java can run the FPP JAR wheels unmodified. That shim intentionally does *not* consult `JAVA_HOME` — precedence between "the user's JVM" and "the pip-provided JVM" should be decided by the caller, and `JAVA_HOME` → `PATH` is the standard order.

## Proposed change

In the JAR branch of `main()` (currently around the `shutil.which("java")` call on the `jvm-25-scala-3.9.0` branch):

```python
def find_java():
    java_home = os.environ.get("JAVA_HOME")
    if java_home:
        candidate = Path(java_home) / "bin" / ("java.exe" if os.name == "nt" else "java")
        if candidate.is_file():
            return str(candidate)
    return shutil.which("java")
```

and use `find_java()` in place of `shutil.which("java")`. The existing `java_major_version()` check and error messages apply unchanged to whichever JVM is selected, so a `JAVA_HOME` pointing at an old JDK still produces the clear "requires Java 25" message.

## Notes

- Verified locally that the current wrapper on `jvm-25-scala-3.9.0` (PR #1106) ignores `JAVA_HOME` and uses the first `java` on `PATH`.
- No change to the native-binary path; this affects only the JAR fallback (and `fpp-to-json`, which always uses the JAR).

bocchino:
Closes #1107.

### DIFF
--- a/python/fprime_fpp/__main__.py
+++ b/python/fprime_fpp/__main__.py
@@ -34,6 +34,14 @@ def java_major_version(java):
         return int(match.group(2))
     return major
 
+def find_java():
+    java_home = os.environ.get("JAVA_HOME")
+    if java_home:
+        candidate = Path(java_home) / "bin" / ("java.exe" if os.name == "nt" else "java")
+        if candidate.is_file():
+            return str(candidate)
+    return shutil.which("java")
+
 def main():
     """ Run fpp inferring a subcommand from the provided executable path
 
@@ -58,9 +66,9 @@ def main():
     # Then check for the JAR file
     elif jar_file.exists():
         # Check for java availability when running the JAR file
-        java = shutil.which("java")
+        java = find_java()
         if not java:
-            print(f"[ERROR] {sys.argv[0]} requires 'java'. Please install 'java' and ensure it is available on the PATH.")
+            print(f"[ERROR] {sys.argv[0]} requires 'java'. Please install 'java' and ensure it is available at JAVA_HOME or on the PATH.")
             sys.exit(-23)
         # Check the java version, so that a too-old JVM produces a clear message
         # instead of an UnsupportedClassVersionError


bocchino:
I'm blocking this PR until I understand the implications of this change. See the discussion in #1107.

bocchino:
After some discussion with Claude, I'm OK with the general approach, but I wonder if we should have more error reporting. Also of note:
* Checking for Java versions will add startup overhead to the tools, and we've already seen that the JVM tools are slow.
* Claude points out that at least when JAVA_HOME is used, one can read the version from a file inside JAVA_HOME, and that's faster than probing `java` for the version. We may be able to use that, but it may not work when JAVA_HOME is not set.

bocchino:
A minimal approach to reporting (with no extra overhead) might be to say, in the error case, where we got the java from (JAVA_HOME or the PATH). That would at least say precisely what is wrong.

bocchino:
After discussion, we will report the source of the java (JAVA_HOME or path).

thomas-bc (approved):

Kronos3 (approved):

PR #1117, opened 2026-09-19
Fix templates 2

Kronos3:
This PR represents fixes found by Claude Opus while analyzing templates as a whole against `main` and consistency with the spec. Most of this branch is testing gaps identified by Claude as well as fixes for edge cases and a some more interesting findings.

This PR adds tests that validate an expanded template has the same codegen as a "hand-exampled" template.

This PR includes the following fixes:

- Resolve names in a template body against the body's own scopes, not the merged expansion-site scope
  - This is the main interesting change see some fun examples of this below
- A template parameter may no longer satisfy an implied framework use
- Honor Value.AnonArray.scalar everywhere; re-finalize constant expressions after FinalizeTypeDefs
- Analyze a bound template argument on demand instead of assuming it was already analyzed
- Fill a missing struct member from the member type's default, not the struct definition's default
- Treat a topology bound to an interface parameter as an imported subtopology
- Let a connection pattern name a bound interface parameter
- Reject a deployment topology bound to an interface parameter
- Make three orderings distinguish nodes cloned by two different expansions
- Visit a template body for dependencies only when the template is expanded
- Register and path-check locate specifiers written inside a template body
- Record implied uses of a string passed as a template type argument
- Degrade gracefully in generated-file computation when expansion fails
- Stop fpp-locate-defs swallowing expansion errors; stop fpp-locate-uses crashing
- Stop dropping expanded members from AST dumps
- Compute includes for a type template argument with UsedSymbols

Most of these fixes are edge case checks the making sure to resolve all template parameters at usage points.

## Template body resolution scope

This PR keeps track of which symbols were introduced by a template rather than merging them into the parent scope. It's important because symbols outside of the template can leak into the expansion if the template creates inner scopes and shadows the template parameters:

```fpp
module template T(constant n: U32) {
  module Inner {
    array A = [n] U32
  }
}

module M {
  module Inner {
    constant n = 3
  }

  expand T(constant 0)
}
```

Before the analyzer  resolved `n` (in `[n] U32`) -> `M.Inner.n` rather than the constant parameter `n` which is exactly the case we wanted to avoid with templates.

Now we keep track of the `localScope` per-expansion which holds symbols introduced by the template and this get's resolved against the outer scope _outside_ the template and stays unmerged _inside_ the template.


Kronos3:
I was able to validate this end-to-end by converting the DataProducts sub-topology into a template. I found that #1118 is needed for templates to be useful with sub-topologies.

Also string interpolation (#893) would be very useful with templates. 

Screenshot 2026-09-20 at 9 49 20 AM

Because we cannot build strings in expressions as part of init specs (phase), we can only instantiate this topology in a single module:

Screenshot 2026-09-20 at 9 51 36 AM

bocchino:
This branch needs to be synchronized with main before it can be reviewed and merged.

bocchino:
Related changes needed:

- [ ] Need to update the wiki with the new resolution algorithm for parameters
- [ ] Need to update the spec to prevent deployment topologies from binding to template arguments
- [ ] Update the spec to handle the case where a template parameter would shadow a framework definition

docs/spec/Definitions/Module-Template-Definitions.adoc
```diff
@@ -39,33 +39,37 @@ This rule ensures that <<Expansion-of-Templates,expanding a template>>
 never causes another template to be defined or another
 template expansion to occur.
 
-Each module template definition _M_ that appears in a model
```
bocchino, line 42: After discussion, let's put back the checks that were deleted. There is a concern that the check is too restrictive because we don't allow type bounds on the type parameters. However, (1) the check is not implemented now anyway and (2) we can add type bounds later if needed, to support this check.

bocchino: Also:
* Add a check for consistency of connections with port interface instances
* Document what is implement now

docs/spec/Ports.adoc
```diff
@@ -90,9 +90,9 @@ with the port name in _n_.
 
 . Add _i'_ to _I_.
 
-==== Template Parameter Definitions
+==== Port Interface Instance Template Parameters
```
bocchino, line 96: After discussion, let's handle this similarly to components and component instances. Let's have one section for the unbound parameter in the template and another section for the bound parameter (the definition) in the expansion.

Kronos3:
I've confirmed that the F Prime CMake build depends on the following FPP tools:

```
fpp-depend fpp-to-cpp fpp-locate-defs fpp-to-dict
```

So we should be safe to revert the fpp-filenames change back to not expand templates

Kronos3:
@bocchino After more thought about:

> Need to update the spec to prevent deployment topologies from binding to template arguments

The way this is checked at the moment is by checking that a `instance <top>` (import) into another topology is not a deployment topology (or a template arg bound to a deployment topology). I think this check is sufficient. There could be a situation in the future where deployment topologies _can_ be referred to and therefore have a use-case for being bound to a template parameter.

Kronos3:
One question that _does_ arise is should we allow template parameters to be used in telemetry packets? We have no bound that specifies which telemetry channels need to be on the arg _and_ more importantly we don't have a bound that forces the argument to be component instance.

~~I vote for disallowing the use of template parameters in telemetry packets because telemetry packets should be specified at the top level (deployment topology) anyway and it doesn't make sense to generate a deployment topology in a template.~~ I'm not actually sure if this is a good argument. Let's discuss. (I can imagine someone could instantiate the same component with two different instances and wants to template different packets across multiple instances)

bocchino:
> The way this is checked at the moment is by checking that a instance <top> (import) into another topology is not a deployment topology (or a template arg bound to a deployment topology). I think this check is sufficient.

I agree it's sufficient for now. I think it's in the same category as type bounds. Eventually we may want more support for checking correctness of the template definition; then we'll need more info in the parameter.

bocchino:
> One question that does arise is should we allow template parameters to be used in telemetry packets? 

I think it's OK to allow this for now. Again, we can put it in the general category of checking template defs, which we are not currently doing.

Kronos3:
I'm allowing component instances through bound template parameters now (and doesn't require any spec changes). See 4a8aa9e8533860dfb07a121da1a60a089b7eaaec for tests/diagnostic output

Kronos3:
@bocchino All the relevant updates are made to the spec, wiki, and implementation. I believe we are ready for merge!

bocchino (approved):
Looks great! Nice work!

Issue #1116, opened 2026-09-18
fpp-to-cpp drops rational format fields on aliases of F32 and F64


Dev-next-gen:
When the type of a formatted element is a type alias whose underlying type is F32 or F64, fpp-to-cpp ignores the rational format field and emits the default `%f`.

Example:

```fpp
type ATF32 = F32

array AliasTypeFormat = [3] ATF32 format "{.3e}"
```

The generated `toString` code contains:

```cpp
tmp.format("%f", static_cast<F64>(this->elements[index]));
```

whereas the same array declared with `F32` directly produces `"%.3e"`. The precision and the `e` conversion are both lost, silently: fpp-check accepts the format string, because `Analysis.computeFormat` resolves the alias before testing `isFloat`, and there is already a test covering that (`fpp-check/test/array/format_alias_float_not_int.fpp`).

The cause is in `FormatCppWriter.writeField`: the rational case matches on the syntax of the type name node (`tn.data match { case Ast.TypeNameFloat(_) => ... }`), and an alias name parses as `Ast.TypeNameQualIdent`, so it never matches and falls through to the default format. The integer case immediately above already dispatches on `s.a.typeMap(tn.id).getUnderlyingType`, which handles aliases correctly. The gap between the two cases is what suggests this is unintended rather than a design choice.

This affects every entry point of the writer: telemetry channels, event parameters, struct members and array elements. It applies to `{e}`, `{g}` and `{.Nf}` as well as `{.3e}`.

I have a two-line fix that makes the rational case dispatch on the underlying type exactly like the integer case does, plus a test case added to the existing `alias_type` case in `fpp-to-cpp/test/array` with the reference output regenerated by `update-ref`. Before the fix that test fails with

```diff
-    tmp.format("%.3e", static_cast<F64>(this->elements[index]));
+    tmp.format("%f", static_cast<F64>(this->elements[index]));
```

and after it passes. When I regenerated the reference files with the patched compiler, the only files that changed were the four new ones for this case, so no existing reference output moves. `sbt test` reports 633 tests succeeded, 0 failed.

The branch is here if it helps you judge, and I have opened no pull request: https://github.com/nasa/fpp/compare/main...Dev-next-gen:fpp:fix-format-alias-float

Following CONTRIBUTING.md, I am asking before implementing further rather than sending a pull request. If you would rather handle this yourselves, or if the behaviour is intended for a reason I have not seen, say so and I will leave it there.

[...]

PR #1119, opened 2026-09-20
Fix fpp-to-cpp dropping rational format fields on aliases of F32 and F64

Dev-next-gen:
| | |
|:---|:---|
|**_Related Issue(s)_**| Fixes #1116 |
|**_Has Unit Tests (y/n)_**| y |
|**_Documentation Included (y/n)_**| n |
|**_Generative AI was used in this contribution (y/n)_**| y |

---
## Change Description

`FormatCppWriter.writeField` dispatched the rational (floating-point) format case on the syntax of the type-name node, so a type alias whose underlying type is `F32` or `F64` never matched `Ast.TypeNameFloat` and fell through to the default `%f`. The integer case already dispatched on the denoted type; this change does the same for the rational case, resolving the alias before selecting the conversion. No FPP or interface change.

Also:
- A new test case `alias_type.fpp` under `compiler/tools/fpp-to-cpp/test/array`, exercising an array element whose type is an `F32` alias with a rational format, plus the regenerated reference outputs (`ATF32AliasAc`, `AliasTypeFormatArrayAc`) and the `run.sh` / `update-ref.sh` updates that include it.

## Rationale

Fixes #1116. With `type ATF32 = F32` and a format of `{.3e}`, semantic checking resolves the alias and accepts the format, but `fpp-to-cpp` then emitted `tmp.format("%f", ...)` instead of `"%.3e"`: the precision and the `e` conversion were silently dropped, while the same array declared with `F32` directly produced `"%.3e"`.

## Testing/Review Recommendations

- The new `alias_type.fpp` reference outputs now show `%.3e` where the old code produced `%f`; with only `FormatCppWriter.scala` reverted and the test kept, the reference comparison fails on exactly that difference.
- The production change is two lines in `FormatCppWriter.scala`, reusing the dispatch the integer case already does.

## Future Work

None.

## AI Usage (see [policy](https://github.com/nasa/fprime/blob/devel/AI_POLICY.md))

This contribution comes out of a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)). It uses Claude Code with Anthropic's Claude Opus 5.

- **Type of assistance**: finding the defect, the code change, the unit test case, and the text of this PR and the issue.
- **Scope**: everything in this PR — the rational-format dispatch in `FormatCppWriter.scala`, the new `alias_type.fpp` case and its regenerated reference outputs under `compiler/tools/fpp-to-cpp/test/array`.
- **How it was found**: the pipeline searched `fpp-to-cpp` code generation for cases where a resolved type alias is treated differently from its underlying type, and found the rational format field dropped on aliases of `F32`/`F64` while the integer case handled them.
- **Level of modification**: the change intentionally reuses the existing integer-case pattern (dispatch on the denoted type) rather than a new one. Nothing leaves the pipeline without a test that fails before the change and passes after it, run with the project's own tooling.

IAMAI

"Nothing this pipeline produces is submitted without human approval. The entire pull request — the code change, the test case and its reference outputs, and this description — was reviewed by a human before the pipeline was allowed to submit it."

(@Dev-next-gen)


Kronos3 (approved):
Nice catch!

bocchino (approved):
Looks good!

PR #1121, opened 2026-09-21
Count data product record IDs in Component.getMaxId

### LINKED ISSUE #1120: Component.getMaxId does not count data product record IDs
`Component.getMaxId` in `compiler/lib/src/main/scala/analysis/Semantics/Component.scala` folds over the maps that contribute relative identifiers:

```scala
val maxMap = Vector(
  commandMap,
  containerMap,
  eventMap,
  paramMap,
  tlmChannelMap
).maxBy(maxInMap)
```

`recordMap` is missing. It is declared in the same case class, right next to `containerMap`, and `hasDataProducts` already counts both.

Record IDs are relative offsets added to the instance base ID, the same as every other kind of ID in that list. `DictionaryEntries` builds record entries with `val id = ci.baseId + localId`, and `ComponentDataProducts` generates `const FwDpIdType id = this->m_baseId + RecordId::$name;`. So a record at relative ID `0x100` in an instance based at `0x100` occupies global ID `0x200`.

`ComponentInstance.maxId` is `baseId + component.getMaxId`, and `CheckComponentInstanceDefs.checkIdRanges` uses that value to reject instances whose base ID falls inside another instance's range. Because records are not counted, the reported range is too short whenever the largest relative ID of a component belongs to a record, and the check lets overlapping instances through.

The consequence is silent. `ConstructDictionary` accumulates entries into a `Map[BigInt, Entry]` keyed on `baseId + localId`, so two records that land on the same global ID do not collide loudly — one replaces the other, and the ground dictionary ends up describing fewer records than the model defines.

Minimal model:

```fpp
passive component C {
  product request port productRequestOut
  sync product recv port productRecvIn
  product send port productSendOut
  time get port timeGetOut
  product container Container id 0
  product record R0: U32 id 0
  product record R1: U32 id 0x100
}

instance c1: C base id 0x100
instance c2: C base id 0x200
```

`fpp-check` accepts this today. `c1.R1` and `c2.R0` both have global ID `0x200`. Replacing `product record R1: U32 id 0x100` with, say, a telemetry channel at the same relative ID makes `fpp-check` reject the same model, which shows the range check itself is fine — only its input is.

I have a one-line fix (add `recordMap` to the `Vector`) plus an `fpp-check` test, and I would be happy to open a PR if this is accepted.

## AI Usage (see [policy](https://github.com/nasa/fprime/blob/devel/AI_POLICY.md))

This report comes out of a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)). It uses Claude Code with Anthropic's Claude Opus 5.

- **Type of assistance**: finding the defect, reducing it to the model above, and the text of this issue.
- **Scope**: the analysis of `Component.getMaxId` and its effect on `CheckComponentInstanceDefs.checkIdRanges`.
- **Level of modification**: the behaviour described here was reproduced with the repository's own `fpp-check` test tooling before the issue was filed.

IAMAI

Dev-next-gen:
| | |
|:---|:---|
|**_Related Issue(s)_**| Fixes #1120 |
|**_Has Unit Tests (y/n)_**| y |
|**_Documentation Included (y/n)_**| n |
|**_Generative AI was used in this contribution (y/n)_**| AI |

---
## Change Description

`Component.getMaxId` folds over the maps that contribute relative identifiers, but it leaves out `recordMap`:

```scala
val maxMap = Vector(
  commandMap,
  containerMap,
  eventMap,
  paramMap,
  tlmChannelMap
).maxBy(maxInMap)
```

`recordMap` is declared right next to `containerMap` in the same case class, and `hasDataProducts` already counts both, so the omission reads as an oversight from the time data product records were added.

This adds `recordMap` to the fold, keeping the list in alphabetical order.

## Rationale

Record IDs belong in that list. `DictionaryEntries` builds record entries with `val id = ci.baseId + localId`, and the generated C++ computes `const FwDpIdType id = this->m_baseId + RecordId::$name;` — a record ID is a relative offset added to the base ID, exactly like a command opcode or a channel ID.

`ComponentInstance.maxId` is `baseId + component.getMaxId`, and `CheckComponentInstanceDefs.checkIdRanges` uses it to reject an instance whose base ID lies inside another instance's range. With records uncounted, the range is too short whenever a component's largest relative ID is a record's, and the check lets overlapping instances through.

The failure is silent rather than loud: `ConstructDictionary` accumulates entries in a `Map[BigInt, Entry]` keyed on `baseId + localId`, so two records sharing a global ID do not raise anything — the second one replaces the first, and the ground dictionary describes fewer records than the model does.

## Testing/Review Recommendations

`compiler/tools/fpp-check/test/component_instance_def/conflicting_ids_record` defines a component whose largest relative ID is a record's, and two instances whose ranges overlap on that record:

```fpp
product container Container id 0
product record R0: U32 id 0
product record R1: U32 id 0x100
...
instance c1: C base id 0x100
instance c2: C base id 0x200
```

`c1.R1` and `c2.R0` both land on global ID `0x200`.

Without the fix the suite reports `19 passed, 1 failed` and `fpp-check` prints nothing at all for that model. With the fix it reports `20 passed` and `fpp-check` prints the expected `base ID 512 for instance c2 lies inside the ID range [256, 512] for instance c1`. The whole `fpp-check` suite (`./do-all run`) passes with the change, and no existing `.ref.txt` moves — the new error only appears for models that were already overlapping.

I built and ran this against JDK 25 and sbt 1.10.11, matching `.github/workflows/build-test.yml`.

## Future Work

`docs/spec/Definitions/Component-Instance-Definitions.adoc` still describes the base ID as being added "for each command, event, telemetry, or parameter identifier". That wording predates data products — the implementation has been counting `containerMap` for a while, and this change adds records. I left the `.adoc` alone because updating it means regenerating the committed HTML, but it looks worth a separate pass.

## AI Usage (see [policy](https://github.com/nasa/fprime/blob/devel/AI_POLICY.md))

This contribution comes out of a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)). It uses Claude Code with Anthropic's Claude Opus 5.

- **Type of assistance**: finding the defect, the one-line code change, the `fpp-check` test, and the text of this PR and of #1120.
- **Scope**: everything in this PR — the `maxMap` fold in `compiler/lib/src/main/scala/analysis/Semantics/Component.scala` and the `conflicting_ids_record` test under `compiler/tools/fpp-check/test/component_instance_def`.
- **How it was found**: the pipeline compared the maps `getMaxId` folds over against the maps the same case class declares, and `recordMap` was the one that never reached the fold.
- **Level of modification**: the fix follows the existing shape of the fold rather than introducing a new one. Nothing leaves the pipeline without a test that fails before the change and passes after it, run with the project's own tooling — here the `compiler/install` script and `./do-all run`, in a container with the network cut during the test.

IAMAI

"Nothing this pipeline produces is submitted without human approval. The entire pull request — the code change, the tests and this description — was reviewed by a human before the pipeline was allowed to submit it."

(@Dev-next-gen)


Kronos3 (approved):

bocchino (approved):
Looks good!

Issue #1122, opened 2026-09-21
Update procedure for native image trace


bocchino:
With the update to Java 25, the procedure for the native image trace has changed. `reflect-config.json` is no longer used.

- [x] Update the procedure in the README
- [x] Update the checked-in trace files

bocchino:
Done in #1123.

PR #1123, opened 2026-09-21
Revise native image procedure

bocchino:
Closes #1122.


compiler/README.adoc
```diff
@@ -177,10 +177,9 @@ To do that, use the following procedure:
 . Run `./install-trace` to install JVM versions of the FPP tools with
 tracing enabled in the `bin` directory.
 
-. Edit the file
-`compiler/lib/src/main/resources/META-INF/native-image/reflect-config.json`
-so that it contains a pair of square brackets `[]` and nothing else.
-This will clear out any old trace information.
+. Delete all files in the directory
+`compiler/lib/src/main/resources/META-INF/native-image`.
+This step will clear out any old trace information.
```
Kronos3, line 182: Should we update `install-trace` to do this automatically?

bocchino: I don't think so. In general, installing the trace binaries shouldn't wipe out the trace. We could add a separate command to wipe out the trace, but it's a single command anyway.

bocchino: We could add a single command to do the entire process: install, wipe out trace, run.

Kronos3: I'm ok either way. I don't mind manually deleting this directory

bocchino: OK, then let's keep it simple (as is).

Kronos3 (approved):

